PR #336, opened 2014-08-16
Rename San Antonio, TX to Bexar County, TX and switch to ESRI shapefile source

kreed:
San Antonio recently added an ESRI shapefile download for their address data. I switched to that since it seems like it would be preferred over the ArcGIS server. (I guess?)

I also switched the coverage to Bexar country since the city's data covers every city and unincorporated area in the county except for the military installations.

Unfortunately the download currently fails due to an SSL error because the server is failing to send an intermediate CA certificate. I sent them an email about the issue, so hopefully that will be resolved. If not I guess we will have to work around that somehow.


geobrando:
Was able to download this without a certificate error, so this can be merged.  But @ingalls you might want to open up separate issue if the curl flag is something that still needs to be implemented.

kreed:
@geobrando Still not working for me. I don't think anything has changed. Never got a response to the emails I sent.

Web browsers will fetch the missing intermediate cert automatically. nodejs/wget/etc aren't able to do that, which causes the problem.
```
# wget https://gis.sanantonio.gov/Download/Addresses.zip
--2015-02-13 19:46:33--  https://gis.sanantonio.gov/Download/Addresses.zip
Resolving gis.sanantonio.gov (gis.sanantonio.gov)... 161.226.203.126, 161.226.217.126
Connecting to gis.sanantonio.gov (gis.sanantonio.gov)|161.226.203.126|:443... failed: Connection refused.
Connecting to gis.sanantonio.gov (gis.sanantonio.gov)|161.226.217.126|:443... connected.
ERROR: cannot verify gis.sanantonio.gov's certificate, issued by ‘CN=Entrust Certification Authority - L1C,OU=(c) 2009 Entrust\\, Inc.,OU=www.entrust.net/rpa is incorporated by reference,O=Entrust\\, Inc.,C=US’:
  Unable to locally verify the issuer's authority.
To connect to gis.sanantonio.gov insecurely, use `--no-check-certificate'.
```

geobrando:
@kreed I see. Misunderstood

iandees:
Sorry this sat around for so long. Could you re-open it against a more recent version of the source file?

kreed:
I can do that. We still need to work around the SSL issue though. Is there a way to disable certificate verification with the new downloader?

Issue #132, opened 2014-04-23
Australia


ingalls:
http://data.gov.au/dataset/launceston-city-council-addresses

CCBY3.0: Australia
http://mapping.launceston.tas.gov.au/arcgis/rest/services/Public/Addresses/MapServer/0

wboykinm:
@sbma44 This is hella-superseded by #1470, right?

sbma44:
yeah, but we generally keep duplicate sources. I see a `tas-launceston.json` in sources, though, so I think this can be closed.

PR #700, opened 2014-11-22
au-queensland

sbma44:
Originally discussed https://github.com/openaddresses/openaddresses/issues/132 by @CloCkWeRX 

Depends on https://github.com/openaddresses/openaddresses-conform/pull/30 -- this dataset is a CSV that is pipe-separated and lacks headers, which necessitated additional conform options.

See JSON for license notes. It was clearly listed as a CC-BY equivalent, but the license URL now redirects to the front page of their open data portal. I've emailed the relevant staff for clarification. As with `au-victoria`, delivery is via a web/email temporary link workflow. The unchanged file is cached in OA's S3 bucket.

image

### DIFF
--- a/sources/au-queensland.json
+++ b/sources/au-queensland.json
@@ -1,24 +1,27 @@
 {
-    "coverage": {
-        "country": "au",
-        "state": "queensland"
-    },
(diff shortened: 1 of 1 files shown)


andrewharvey:
The upstream dataset has moved too http://qldspatial.information.qld.gov.au/catalogue/custom/detail.page?fid={F878C43D-3087-4102-8F28-1CFEA49B34F1} as per the metadata it is still CC-BY 3.0 AU http://qldspatial.information.qld.gov.au/catalogue/rest/document?id={F878C43D-3087-4102-8F28-1CFEA49B34F1}&f=xml

As before, download is via email delivery. The latest release is 20150216, which is newer than the current version on OpenAddresses.io

Technically the coordinates are in GDA94 (EPSG:4283), however I think you can treat it as WGS84 and the result will be the same.

PR #720, opened 2014-11-25
MO, MA, RI Statewide layers!

ingalls:
- Consolidates MA Sources into single layer
- Add MO Statewide Layer
- Add conform object to RI state layer

Big thanks to the Michael Fashoway, the Structures & Addresses Data Coordinator who came in via email about the state-wide address layer they have been working on. 

cc/ @iandees @sbma44 @migurski

### DIFF
--- a/sources/us-ma.json
+++ b/sources/us-ma.json
@@ -1,0 +1,20 @@
+{
+    "coverage": {
+        "country": "us",
+        "state": "ma"
+    },
+    "data": "https://www.dropbox.com/s/4dm6v0yeb96ulip/SHPStatewideParcelsAug2014.zip?dl=1",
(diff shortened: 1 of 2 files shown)


MiniCodeMonkey:
The MA dropbox link is dead -- https://www.dropbox.com/s/4dm6v0yeb96ulip/SHPStatewideParcelsAug2014.zip?dl=1

Does anyone have a copy of this? Also, is there a better location we can put these special files in?

iandees:
Yes, I'd like to see us store the hand-built stuff in our S3 bucket so that we don't have to rely on Dropbox or external Github links.

If we find this data again let me know and we can hang on to it in a more permanent way.

ingalls:
Did the caching script back this up? If so we can simply grab the cached file and move it to the self-maintained directory.

Issue #771, opened 2014-12-15
Spain


sbma44:
[Spain's cadastral data seems to be open by decree](http://www.catastro.minhap.es/ayuda/legislacion/ovc/resoluciondgc20110323_tfs.pdf). Unfortunately, in [its easiest-to-use forms it is gated](https://www.sedecatastro.gob.es/OVCFrames.aspx?TIPO=TIT&a=masiv) by client-side certs that are only available to Spanish citizens who are using eServices.

Luckily, as part of the INSPIRE process, Spain has released an ungated version [here](http://blog-idee.blogspot.com.es/2014/03/actualizacion-de-los-datos-de-descarga.html). Less luckily, it consists of 7,000+ GML files, where the street name is made available through an XLINK record that `ogr2ogr` does not seem to resolve correctly. For now I am working under the assumption that this data is licensed under the same terms as the non-INSPIRE dataset. Some custom scripting will be necessary to resolve the street names and make this data processable by OA.

**NOTE:** OSM `talk-es` users have pointed me to some concerns about the quality of the cadastral dataset -- specifically, house numbers getting paired with the wrong street names. I'm not yet sure how big of a deal this is. Any Spanish speakers who can give the following discussion a closer read, by all means do so (I'm working off of Google Translate). 

http://gis.19327.n5.nabble.com/Catastro-duda-nombres-de-calles-td5777261.html

sbma44:
reopening: looks like Spain has an open data portal at http://www.cartociudad.es/portal/ which might allow us to get away from the cadastral source.

EamonKeane:
@sbma44 yes can confirm that works. I used that page to put together a full dataset without authentication required. This is done by downloading the 50 provinces from the 'CartoCity' dataset in the search by listing section.

The data returned is like this:
https://www.dropbox.com/s/54s9xvwqcd4g1s9/CARTOCIUDAD_CALLEJERO_ALAVA.zip?dl=0

<img width="423" alt="screen shot 2017-07-13 at 15 36 36" src="https://user-images.githubusercontent.com/1946655/28171825-9980d772-67e1-11e7-9fea-8388dc06e243.png">

By combining the PORTAL_PK shapefiles, you can reproduce the Portales y PK layer of this WMS:
http://www.cartociudad.es/wms/CARTOCIUDAD/CARTOCIUDAD?Request=GetCapabilities

I manually downloaded the 50 files, but I believe you could automate this process by using an AJAX scraper. I might look into this if I have time.

[...]

PR #805, opened 2014-12-31
Spain

sbma44:
14.5m points. Lacks Basque Country, as discussed at https://github.com/openaddresses/openaddresses/issues/771#issuecomment-68455596

Split into three files because of varying projections in the source data.


ingalls:
:+1: Looks great!

PR #808, opened 2015-01-01
add ISO 3166 for au-vic

jmkelly:


migurski:
First ISO 3166-2 I’ve seen. I’ll need to make it possible for this to render.

PR #812, opened 2015-01-02
Update us-ny-nyc.json

MiniCodeMonkey:
Corrected conform column names

### DIFF
--- a/sources/us-ny-nyc.json
+++ b/sources/us-ny-nyc.json
@@ -11,9 +11,9 @@
     "conform": {
         "lon": "x",
         "lat": "y",
-        "number": "HOUSE_NUMB",
-        "street": "STREET_NAM",
+        "number": "HS_NUM",
+        "street": "STREET_NM",
         "type": "shapefile",
-        "postcode": "zipcode"
+        "postcode": "ZIPCODE"
     }
 }


MiniCodeMonkey:
I'm going through the data output and found another issue. Some addresses have a suffix for the house number, e.g. `43 Bushwick Ave` and `43 1/2 Bushwick Ave` are both valid addresses.

Another example is `176A Beach 111th Street` and `176C Beach 111th Street` which are two separate addresses.

What is the preferred way to handle this with the current conform structure? The following *works*, but leaves an extra space after the street number if `HS_NUM_SUF` is NULL (which it of course is in most cases).

``json
"conform": {
    "advanced_merge": {
        "custom_number": {
            "separator": " ",
            "fields": ["HS_NUM", "HS_NUM_SUF"]
        }
    },
    "lon": "x",
    "lat": "y",
    "number": "custom_number",
    "street": "STREET_NM",
    "type": "shapefile",
    "postcode": "ZIPCODE"
}
```

MiniCodeMonkey:
Not sure if anyone notices this in a closed pull request. This should maybe be a separate issue?

@migurski @tmcw @ingalls @iandees

iandees:
I think leaving the space is fine for now. Ideally I'd prefer that the conform output keep this stuff as separate columns, but I don't think anyone has done that yet.

MiniCodeMonkey:
I would love to keep it in separate columns too. My only concern is the wild inconsistency between data sources (especially when it comes to differences between sources from separate countries).

Btw. completely off topic, but is there any other places for OA discussions such as an IRC channel, mailing list or something like that?

iandees:
I've been squatting in #openaddresses on FreeNode. It looks like https://github.com/openaddressesuk sends some Travis build notifications there, but otherwise it's completely dead. Feel free to join me there.

PR #813, opened 2015-01-02
Added us-ca-lassen_county.json

palewire:
Parcel polygons that include situs addresses in the metadata. Provided in response to a public records request in December 2014 and uploaded to GitHub for public hosting. A situs address dbf was provided separately from the parcels shapefile, which were then joined together in QGIS using an 'APN' according to instructions provided a county official.


wboykinm:
@palewire __Can all of California be added this way?__ :pizza:

palewire:
Maybe? There are 58 counties and some of them are very small, so I suspect we'll end up with a few holes.

migurski:
`:fistbump:`

PR #814, opened 2015-01-03
conform object added for VT e911 addresses - should now be parse-able

wboykinm:
These should be GTG. @iandees - is it okay for `accuracy` to be numeric? I couldn't find a good example.

Rooftops of VT, inbound:

### DIFF
--- a/sources/us-vt.json
+++ b/sources/us-vt.json
@@ -1,12 +1,29 @@
 {
     "coverage": {
+        "US Census": {
+            "geoid": "50",
+            "name": "Vermont"
+        },
         "country": "us",
         "state": "vt"
     },
     "website": "http://vcgi.vermont.gov/warehouse/theme_index",
-    "data": "http://maps.vcgi.org/gisdata/vcgi/packaged_zips/EmergencyE911_GDBE911.zip",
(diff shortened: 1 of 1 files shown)


iandees:
Looks like `accuracy` is not supported. Can you just remove that line from the conform object?

wboykinm:
@iandees Should I remove `accuracy` from the docs then? Seems like a pretty important item.

iandees:
It [looks like](https://github.com/openaddresses/openaddresses/blob/master/test/sources.js#L69-L70) `accuracy` should be a number, but the `openaddresses-conform` project doesn't use it at all and it's not listed in the [list of available `conform` options](https://github.com/openaddresses/openaddresses/blob/master/test/sources.js#L53).

I would say leave it out of this source for now.

wboykinm:
All good then; who merges?

iandees:
Meee.

iandees:
Next time @migurski runs it we should have data, but the map won't update until @sbma44 re-renders it.

sbma44:
@iandees I reran the map right after the Spain import, but it looks like Flanders data had a busted SRS. Rerunning it now, should be up by tomorrow.

PR #815, opened 2015-01-03
Contributing to: za-nl-ethekwini.json

luiscape:
Adding/Updating za-nl-ethekwini.json

### DIFF
--- a/sources/za-nl-ethekwini.json
+++ b/sources/za-nl-ethekwini.json
@@ -9,7 +9,7 @@
     "license": "http://citymaps.durban.gov.za/GISMetaData/Files/Metadata/Cadastral/Approved_Cadastre/addresspoint.htm",
     "skip": false,
     "year": "2014",
-    "type": "http",
+    "type": "zip",
     "compression": "zip",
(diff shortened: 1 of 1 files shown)


sbma44:
I'm afraid that tests are failing because `type` should be `http`. This is an ambiguously named field, but if you have a look at the [docs](https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md) you'll see that this field has to be `http`, `ftp` or `ESRI`.

Looking at the diff, I'm unsure of what else the PR is doing besides flipping that field. Did the data need a refresh?

ingalls:
This is coming from the online editor so my guess is that it was a test. I'm going to close this. @luiscape feel free to reopen if this is not the case.

luiscape:
@ingalls sorry for that folks. Indeed, it was a test -- or less than that. It wasn't totally clear at the time that I was making a commit. Sorry once again.

iandees:
@ingalls the bug I was pointing out was that it set the "type" to "zip". The editor shouldn't let that happen.

PR #817, opened 2015-01-04
Added ISO-3166-2 codes for Japanese prefectures

migurski:
I'm expecting to add support for admin-1 subdivisions to rendering on http://data.openaddresses.io.


migurski:
Travis passed, merging.

PR #823, opened 2015-01-06
us-mi-washtenaw: add conform object

maxerickson:
The addrtype field could probably be set from PROPCLASS with a simple mapping.


iandees:
Looks like you're missing commas at the end of each line, @maxerickson.

geobrando:
@maxerickson When you added the conform object you needed to start be adding a comma to the end of the previous line.  If you add that I think it will pass and can be merged.

I would recommend running your JSON through [JSONlint](http://jsonlint.com/) to make sure it's valid json.

PR #829, opened 2015-01-07
correct type of us-va to http (despite ftp in subdomain)

sbma44:


iandees:
Woopsadoodle thanks for catching that :).

sbma44:
my pleasure :smile:

PR #842, opened 2015-01-09
second Iceland source

kelvinabrokwa:
From same dataset using other street name column - "skip": "true"


kelvinabrokwa:
New Zealand screenshot
screenshot 2015-01-09 17 14 31

Issue #843, opened 2015-01-09
us-oh-cincinnati.json should be us-oh-hamilton.json


1ec5:
The rest of the Ohio files are named after the counties they cover. CAGIS covers all of Hamilton County, not just Cincinnati.

sbma44:
Thanks! I've renamed the file -- looks like its contents already referred to Hamilton County.

PR #844, opened 2015-01-09
Added Ohio Hamilton county FIPS

migurski:
Follow up to https://github.com/openaddresses/openaddresses/issues/843


geobrando:
@migurski Couldn't see why this didn't pass Travis.  Per your specification I think you meant `"name": "Hamilton County"` in your US Census object but that shouldn't cause [sources.js](https://github.com/openaddresses/openaddresses/blob/master/test/sources.js) to fail, right?

Maybe update against master?

PR #846, opened 2015-01-09
Added us-ca-orange_county.json

palewire:
Parcel polygons that include situs addresses in the metadata. Provided in response to a public records request in January 2015 and uploaded to GitHub for public hosting. A situs address dbf was provided separately from the parcels shapefile, which were then joined together in QGIS using an intermediary file's 'APN' value according to instructions provided by a county official.


palewire:
Did I cause this build failure over at Travis?

iandees:
Nope, looks like the Korean sources causing problems.

PR #860, opened 2015-01-12
fr-montpelliere

rowanwins:


iandees:
Hi @rowanwins! Thanks for your data here. Could you rename the file and include the coverage information? See our [contributor guidelines](https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md) for more information about those fields.

rowanwins:
Hi @iandees , have renamed the file and added the coverage. Apologies for missing this, first time contributor!

geobrando:
Hey @rowanwins , you have an extraneous comma in there.  Easy enough to make that mistake.  I would recommend running your json in something like [JSONlint](http://jsonlint.com/) to make sure it's valid JSON.

Also, per the contributor guidelines referenced previously, you should have a `"type"` field, which in this case would be set to `"http"`

PR #872, opened 2015-01-22
splitting up poland and fixing data links

kelvinabrokwa:


NelsonMinar:
Thanks for doing this!

Is `pl.json` redundant with `pl-dolnoslaskie.json`?

sbma44:
Hmm, I still see `pl.json` in the repo -- @kelvinabrokwa should that be deleted?

kelvinabrokwa:
@sbma44 that was my mistake. `pl.json` should indeed be deleted.

PR #881, opened 2015-01-30

redaready:

### DIFF
--- a/sources/NewSource.json
+++ b/sources/NewSource.json
@@ -1,0 +1,22 @@
+{
+    "conform": {
+        "type": "csv",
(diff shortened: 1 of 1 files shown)


sbma44:
Thanks for the submission! I see two different datasources here, however.
- One points to [the French national address database](http://adresse.data.gouv.fr/). We do want to import French data, but this project is under ODbL and so not usable for us.
- The other is for Buenos Aires, which has already been collected https://github.com/openaddresses/openaddresses/blob/master/sources/ar-ba-buenos_aires.json (if this is an update, please let me know!)

PR #883, opened 2015-02-04
Updated us-ca-monterey url & added conform

davidchiles:
Looks like the URL has changed


davidchiles:
Added a conform object.

Issue #12, opened 2014-02-27
Czech addresses


lxbarth:
If osmlab/addresses is to expand beyond CA and US, here're Czech addresses:

http://wiki.openstreetmap.org/wiki/Import_adres_z_RUIAN

mormegil-cz:
Not sure why the “Source Down” label, RÚIAN is one of official “primary registers” of the Czech Republic, and is the official primary source for such data. The source is definitely not down, only the specific example link used on the above-mentioned page is outdated (it referred to a year-old dump).

For an up-to-date data, see e.g. http://vdp.cuzk.cz/vdp/ruian/vymennyformat/vyhledej?vf.pu=S&_vf.pu=on&_vf.pu=on&vf.cr=U&vf.up=OB&vf.ds=Z&vf.vu=Z&_vf.vu=on&_vf.vu=on&_vf.vu=on&_vf.vu=on&vf.uo=O&ob.kod=539317&search=Vyhledat with links to data of a small town (the small floppy-disk icons in the list). In the gzipped XML, see `/vf:VymennyFormat/vf:Data/vf:AdresniMista/vf:AdresniMisto` for every address point. However, I don’t know whether Conform would be able to use the file without preprocessing, as this is not just GML but a specific XML file with a few quirks added. For instance, the position is in EPSG:5514 instead of WGS84. Also, street names are not included in every address point, but are separately in `/vf:VymennyFormat/vf:Data/vf:Ulice/vf:Ulice` and are referred to using an ID. Etc.

sbma44:
This is great, @mormegil-cz! Custom GML variants are sadly [nothing new](https://github.com/openaddresses/openaddresses/tree/master/scripts/es) around here, so I imagine we could handle this. I'm afraid that I don't have an ETA for being able to tackle this, though.

sbma44:
@mormegil-cz I'm digging into this now -- can you tell me which field from the layers in the GML file is the house number? I have clicked around a bit in AdresniMista but it's not completely clear to me.

mormegil-cz:
[...]

PR #884, opened 2015-02-06

dusekvojtech:


sbma44:
Thanks for the submission. As far as I can tell, this source doesn't include actual coordinates. We would need those to work with it. 

The good news is that work on Czech addresses *is* proceeding over at https://github.com/openaddresses/openaddresses/issues/12 -- keep an eye on that (and share any notes you might have!) if you'd like.

Closing this ticket for now.

PR #888, opened 2015-02-11
Add city of norfolk

geobrando:
Adding City of Norfolk, VA.

Added an ```accuracy``` field to the conform object despite the fact that it was previously stated in #814 that it is currently not used by [openaddresses-conform] (https://github.com/openaddresses/openaddresses-conform).  If not desired, I can remove but we should also then change the contributing docs as previously recommended.


geobrando:
OK, if  ```accuracy``` causes Travis to fail, then contributing docs should definitely be changed, right?

jalbertbowden:
thanks to this thread i stumbled onto va beach's open data portal. they have around 200 datasets. awesome.

geobrando:
@jalbertbowden Nice!

Issue #889, opened 2015-02-11
Esri Open data portal


geobrando:
I just created a pull request in #888 for a shapefile source I found using the [Esri Open Data Portal](http://opendata.arcgis.com/).  The site is now fully operational and allows for download of data in csv, KML, and shapefile formats, as well as provides urls for a GeoJSON or GeoService API.  Most of these appear to be converted on the fly from Esri servers hosted by the different municipalities.

My questions are:
1. What would be the preferred format when submitting a source from the portal?
2. I know that there have been issues previously with pulling addresses from Esri servers.  It was stated in #336 that shapefiles are preferred over Esri servers.  Is there any benefit to changing our existing sources from Esri to shapefiles or another format?
3. The shapefile from the source I submitted was named `e946260f3b23471fa0e3657034bac382_0.zip`.  I don't how their process for generating these names.  Do you think there's any reason to believe that the `data` urls might not be stable over time?

migurski:
There may be a benefit to switching; Esri Feature JSON sources require a time-consuming crawl of address points, while shapefile downloads are a single download. It's better to have one URL 404 when debugging, than to chase an error through hundreds of requests.

Unsure about the data URL stability. Give it a shot, we'll see what happens. We do cache data over time, so an unstable URL won't eject data from http://data.openaddresses.io/

iandees:
We should definitely take advantage of Shapefile downloads from opendata.arcgis.com, but I'm pretty sure that they're only going to include data sources that are already on arcgisonline.com. Most of our existing ESRI sources are not using arcgisonline.com, they run their own ESRI servers on their own hardware. Unless I'm mistaken, I don't think opendata.arcgis.com will help us in those cases.

geobrando:
@iandees There's some overlap. The portal exposes all data from orgs that have the ```open data``` button turned on, whether hosted on ArcGIS Online or on their own Esri servers.

geobrando:
Found a fatal flaw in my plan to add Esri Open Data sources to the current processing pipeline. It's clear that this initiative is not as "open" as I'd previously thought, as the site was designed for a multi-step process using a download console in the browser.

As mentioned, the portal exposes feature service data hosted on either private Esri servers or directly on http://opendata.argcis.com.  Shapefiles, kmz, spreadsheets, etc are generated by some middleware and cached.

[...]

PR #890, opened 2015-02-13
update Surrey, BC and Pitkin, CO

geobrando:
Updating Pitkin, CO to a shapefile source using the new Esri Open Data Portal
Updating Surrey, BC from Esri to geojson source, which should allow #880 to be closed without merging.


geobrando:
Per #889, closing this until we decide how to deal with these sources.

PR #892, opened 2015-02-17
us-nj-cape_may added conform

davidchiles:
My first time adding a conform object. Let me know if there are any errors or issues.

### DIFF
--- a/sources/us-nj-cape_may.json
+++ b/sources/us-nj-cape_may.json
@@ -12,5 +12,20 @@
     "data": "http://www.capemaycountyims.net/Downloads/CMC_Addresses.zip",
     "website": "http://www.capemaycountyims.net/",
     "compression": "zip",
-    "type": "http"
+    "type": "http",
+    "conform" : {
+        "type": "shapefile",
+        "lon": "x",
+        "lat": "y",
+        "merge": [
+            "PREFIX_1",
+            "STNAME_1",
+            "STTYPE_1",
+            "SUFFIX_1"
+            ],
+        "number": "STNUMB_1",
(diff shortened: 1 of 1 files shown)


davidchiles:
Is [conform.accuracy](https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#attribute-tags) supposed to be there. That seems to be [where the test is failing](https://github.com/openaddresses/openaddresses/blob/master/test/sources.js#L68).

geobrando:
@davidchiles I attempted to add accuracy data in #888, which also caused Travis to fail.  Although it's requested in ```Contributing.md```, it doesn't appear to be currently utilized or even supported.

iandees:
Yes, I don't know where the accuracy attribute came from but it shouldn't
be there. If someone could write a PR that removes it from contributing.md
that'd be helpful.

geobrando:
@iandees This was @ingalls in Sha: 97890eae99af90070322e9f92e0ca2a33cd82b19 but the proper tests were not added to ```sources.json```.  I'll let him weigh in.

migurski:
We are not using `accuracy` in machine either.

ingalls:
This is intended to make future conflation easier. We are pulling from more and more sources that have overlap and building tools to dedup and clean up the data into a single dataset are going to be needed. By having a known accuracy level it makes the dedup process easier as duplicate addresses in the same geographic proximity can be deduped based on the accuracy level.

iandees:
@ingalls then we need to make the unit tests recognize it so the Travis build doesn't fail when people try to add it.

PR #893, opened 2015-02-18
Added quadrant i.e. NE or SE to "street"

MiniCodeMonkey:


geobrando:
:+1:  Haven't been able to tell what the consensus is from the owners on adding this data to street versus capturing in optional fields for future use, but agree that this is imperative for DC addresses.

migurski:
Sounds plausible to me.

PR #901, opened 2015-02-24
Accuracy test

geobrando:
@ingalls @iandees Added proper testing for an accuracy property in conform objects that was causing #892 to fail


iandees:
Thanks for doing this!

ingalls:
Hey @geobrando thanks for getting this pushed out! Was on my plate to do this today!

PR #902, opened 2015-02-24
Montgomery County, MD

geobrando:
Was able to get the County Planning office to put up a shapefile.  That's ~350,000 addresses!!


iandees:
Very nice! Thanks for contacting them.

Issue #726, opened 2014-11-26
us-md


tritania:
[Link](http://data.imap.maryland.gov/datasets/20878dd2c20949a8a14a5a282ab98807_0?geometry=-77.734%2C39.003%2C-76.416%2C39.323) contains data for the entire state of Maryland.

slibby:
I believe this can be closed due to https://github.com/openaddresses/openaddresses/blob/master/sources/us/md/statewide.json

talllguy:
The statewide data referenced by @slibby and @geobrando in https://github.com/openaddresses/openaddresses/issues/726#issuecomment-150349696 represents parcel points, not address points.

For this project is is it worth making a distinction?

slibby:
Yes, definitely want a distinction, but I am not aware of a statewide point address dataset (yet) -- is there one you know of? If so, it would supersede the parcel-based dataset (as long as the total number of valid addresses is similar).

The original link is broken in this issue, so I'm not sure what it was pointing to, my interest in closing it was just general issue tidiness since the dataset wasn't available anymore anyways. I know you're local, so feel free to guide on it.

talllguy:
There is a statewide address point layer, but it is not public. Maryland's state government is very much an open data proponent, but there are certain counties that are not ready to give away this data, thus the state cannot publish it. I will update this issue if there are any updates on the matter of statewide address points.

In the meantime, there are certain counties that publish their data outside of the state umbrella, which tend to be of better quality. Would those datasets be worth referencing?

iandees:
Yup! There's no problem with including both the local county data and the statewide layer.

PR #903, opened 2015-02-24
Adding State of Maryland

geobrando:
Just added Montgomery County, MD in #902 but then realized there's a state layer already referenced in #726 with coverage of the same area.  However, this shapefile source is a different data url as #726 was one the troublesome Esri portal sources.  Does not cover entire State so some other individual county sources may be added.

If we're not including duplicates introduced by overlapping state and county sources then we'll need to figure out a way to rectify that.

### DIFF
--- a/sources/us-md.json
+++ b/sources/us-md.json
@@ -2,7 +2,7 @@
     "coverage": {
         "US Census": {
             "geoid": "24",
-            "name": "State of Maryland"
+            "state": "Maryland"
         },
         "country": "us",
         "state": "md"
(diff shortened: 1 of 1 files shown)


iandees:
Overlapping datasets is OK. We'll end up with duplicate rows in the dataset but we'll worry about that later.

geobrando:
@iandees Great.  This can be merged then

iandees:
Cool, I was waiting for the tests to pass and will merge away now.

PR #905, opened 2015-02-24
Update us-ar.json

### LINKED ISSUE #904: Arkansas's data URL 404s
[Arkansas' source file](https://github.com/openaddresses/openaddresses/blob/master/sources/us-ar.json) specifies `"data": "http://www.geostor.arkansas.gov/ArcGIS/rest/services/FEATURE_SERVICES/GEOSTOR_STRUCT/MapServer/0",` but [that URL doesn't work anymore](http://www.geostor.arkansas.gov/ArcGIS/rest/services/FEATURE_SERVICES/GEOSTOR_STRUCT/MapServer/0). I gather that they moved to an S3 URL. We'll need to track down that URL and update `us-ar.json`.

iandees:
Fixes #904.

### DIFF
--- a/sources/us-ar.json
+++ b/sources/us-ar.json
@@ -5,7 +5,7 @@
         "state": "ar"
     },
     "website": "http://www.geostor.arkansas.gov/G6/Home.html",
-    "data": "http://www.geostor.arkansas.gov/ArcGIS/rest/services/FEATURE_SERVICES/GEOSTOR_STRUCT/MapServer/0",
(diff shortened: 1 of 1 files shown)


geobrando:
Was just about to submit a PR to fix this.

I'd also change ```website``` as it now redirects to http://gis.arkansas.gov/

iandees:
The `website` attribute is pointing to the address points product page:

http://gis.arkansas.gov/?product=situs-address-points

geobrando:
Ah.. Nevermind. That's just as good. Didn't notice that you updated it at all.

PR #908, opened 2015-02-24
Update us-in.json

### LINKED ISSUE #907: Add Indiana
Indiana aggregates all but one of its counties' address points files in a single file, [available via an ArcGIS service](http://maps.indiana.edu/arcgis/rest/services/Infrastructure/Streets_Address_Points_IDHS/MapServer), and [documented fully](http://maps.indiana.edu/metadata/Infrastructure/County_Address_Points_IDHS_IN.txt). Somebody who knows what they're doing can add `us-in.json` using this information. :)

iandees:
Fixes #907.

### DIFF
--- a/sources/us-in.json
+++ b/sources/us-in.json
@@ -1,0 +1,26 @@
+{
+    "coverage": {
+        "US Census": {
+        	"geoid": "18",
+        	"state": "Indiana"
+        	},
+        "country": "us",
+        "state": "in"
+    },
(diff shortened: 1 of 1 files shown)


jarz:
Poked around at us-in.json -- looks like it's actually GDB, not shapefile.

iandees:
In theory GDAL should be able to open and use it, but I don't think we've ever actually tested that.

PR #914, opened 2015-02-25
Update us-ma

### LINKED ISSUE #912: us-ma source broken
Zipped shapefile no longer found at [```data``` url] (https://www.dropbox.com/s/4dm6v0yeb96ulip/SHPStatewideParcelsAug2014.zip?dl=1) from [```us-ma.json```](/openaddresses/openaddresses/blob/master/sources/us-ma.json)

@ingalls Was this something you created from county shapefiles or was it being hosted by someone from the state?  Couldn't find a statewide layer at the [portal](http://www.mass.gov/anf/research-and-tech/it-serv-and-support/application-serv/office-of-geographic-information-massgis/datalayers/ftpl3parcels.html)

geobrando:
Worked with MassGIS to get a working shapefile back up on dropbox, so this should persist for a while. But agree that these should be moved to S3.

This fixes #912

### DIFF
--- a/sources/us-ma.json
+++ b/sources/us-ma.json
@@ -1,20 +1,26 @@
 {
     "coverage": {
+        "US Census": {
+            "geoid": "25",
+            "state": "Massachusetts"
+        },
         "country": "us",
         "state": "ma"
     },
-    "data": "https://www.dropbox.com/s/4dm6v0yeb96ulip/SHPStatewideParcelsAug2014.zip?dl=1",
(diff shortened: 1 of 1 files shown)


migurski:
Feel like we should create an upload process for getting stuff into S3 without needing AWS keys.

geobrando:
Would someone like to merge this or should the file be added to the bucket?  It's permanently hosted in a dropbox account owned by MassGIS so I think it should be fine to keep the source data there. Either way.

PR #920, opened 2015-03-01
Update to 2014 North Carolina addresses

iandees:
Update the conform object, too.


geobrando:
Great.  We need to remove the old NC sources that are no longer valid that were added in SHA: 7d6ac4d4c107a4513f0dd5761da2920feec832d3

PR #927, opened 2015-03-10

schnuerle:

### DIFF
--- a/sources/NewSource.json
+++ b/sources/NewSource.json
@@ -1,0 +1,30 @@
+{
+    "coverage": {
+        "US Census": {
+            "geoid": "21185",
+            "name": "Oldham County",
+            "state": "Kentucky"
+        },
+        "country": "us",
+        "state": "ky",
+        "county": "Oldham"
+    },
+    "data": "ftp://ftp.lojic.org/Pub/federal/Address%20Sites.zip",
+    "type": "ftp",
+    "conform": {
+        "merge": [
+            "str_name",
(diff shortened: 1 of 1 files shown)


iandees:
Can you rename this file `us-ky-oldham.json`. Also, the `filename` key in here shouldn't be there.

schnuerle:
Hey  it's actually for Jefferson county, not Oldham.  Maybe I should have mentioned that.  So the json file would be named us-ky-jefferson.json.  But which file do you need renamed, the one here? ftp://ftp.lojic.org/Pub/federal/Address%20Sites.zip  

If so I can't edit that file name (it's from a city organization's system).

iandees:
Nope, the file in your pull request. It's currently named "NewSource.json".

iandees:
Also, if this is for Jefferson county the "coverage" object needs to be updated to reflect the correct county.

schnuerle:
Ok I'll work on all that.  I just used an online form so there were no options on it for geoid, etc.

geobrando:
@iandees As mentioned, these were submitted via the web map for those not comfortable working directly in the repo.  I think it's up to us to rename as we're verifying the source.

iandees:
Ah, righto.

@ingalls: can you fix the website so it 1) creates the correct file name and 2) populates the new file with correct values?

In the mean time, I'm going to close this and copy and paste the source. @schnuerle, thanks very much for the contribution!

PR #932, opened 2015-03-11
Contributing to: andor-manetheren-two_rivers.json

ingalls:
Adding/Updating andor-manetheren-two_rivers.json

### DIFF
--- a/sources/andor-manetheren-two_rivers.json
+++ b/sources/andor-manetheren-two_rivers.json
@@ -1,0 +1,12 @@
+{
+    "coverage": {
+        "country": "Andor",
+        "state": "Manetheren",
+        "city": "Two Rivers"
+    },
+    "data": "tworivers.com/addresses.zip",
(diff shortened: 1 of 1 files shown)


ingalls:
See openaddresses/openaddresses.io#47

ingalls:
cc/ @iandees @geobrando

geobrando:
Looks good.  I guess my thinking it would require more thought if we were trying to form filenames that didn't require any manual adjustment from other contributors (i.e. Country ISO code+ region name).  If that's the goal, we could either use some dropdown menu or a lookup table.  Also, the current naming scheme asks for the largest extent (county OR city but not both).   But your changes should at least make it easier to rename.

ingalls:
Yes, I'm hoping to flesh this out later today with something that will verify the fields to ensure better compliance with our coverage standards. Just wanted to get something out the door to fix this as I know how annoying its been.

PR #935, opened 2015-03-14
Update "conform" documentation for lat/lon.

NelsonMinar:
For every type we understand but CSV, the lat/lon is implicitly named
by being in the geometry field, so the contributer doesn't need to
specify it. The Python code will ignore these settings. It's possible
the Node code relies on them being set; if so, the Node code should
assume the names "X" and "Y" to match the rest of how it works.

For details on this change, see https://github.com/openaddresses/machine/issues/73


NelsonMinar:
Thanks for the review! You're right we want these omitted, not left as empty tags, I updated the text. I left it as "should" because I think we really mean that over "may be" (in the normative sense.)

I'm not sure about etiquette committing to this project yet, so I'm hoping someone else presses the "merge" button as a sort of rough consensus on the change.

migurski:
Looks okay to me, and backwards-compatible with existing documents.

PR #936, opened 2015-03-19
update us-pa-allegheny.json

psylum:
updated the ftp reference to the latest zip file and ensured zip code attribute is uppercase.

the actual .shp is in a subdirectory, so this may need a file tag. I was also tempted to remove the license tag since the contributor info says to elide empty values, but decided to leave that be for now.


iandees:
You only need the `file` tag if there are more than one .shp file.

Issue #925, opened 2015-03-07
Recife, Brazil


astoff:
http://dados.recife.pe.gov.br/dataset/area-urbana/resource/c82a414a-066f-4acc-922c-e3a63ed047f8

The address of each parcel is contained in the field endereco_completo, in the format "(street name), (house number)".

Data is under the ODbL.

slibby:
This issue can be closed. The source still needs a conform object with a regex to extract the Street Name, but that "issue" is already captured in each /results run.

PR #939, opened 2015-03-20
Contributing to: br-pe-recife.json

astoff:
Adding/Updating br-pe-recife.json


astoff:
The format of the address field is explained in https://github.com/openaddresses/openaddresses/issues/925

astoff:
@geobrando There is no way to split the street and number in this dataset, is there?  (The address field in the shapefile is in the format street name, comma, blank, house number.)

PR #941, opened 2015-03-24
Contributing to: br-es-vitória.json

astoff:
Adding/Updating br-es-vitória.json

### DIFF
--- a/sources/br-es-vitoria.json
+++ b/sources/br-es-vitoria.json
@@ -10,8 +10,6 @@
     "type": "http",
     "compression": "zip",
     "conform": {
-        "lon": "x",
-        "lat": "y",
         "street": "LOGRADOURO",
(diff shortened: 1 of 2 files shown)


geobrando:
Since lat/lon are no longer required conform fields we need to update the tests and this should pass.

astoff:
Should I add lat.lon fields so this can be merged?

Also, there is no formal license statement on the website, but I was informed by email that the data can be used freely provided attribution in given.  What is the best way to keep track of this information?

ingalls:
For attribution see: https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#optional-tags

astoff:
The diacritical in the file name was causing an error, so I renamed the file.  The website wizard should strip accented characters from city names.

PR #943, opened 2015-03-27
Contributing to: usa-sw-albuquerque.json

PaulCrickard:
Adding/Updating usa-sw-albuquerque.json

### DIFF
--- a/sources/usa-sw-albuquerque.json
+++ b/sources/usa-sw-albuquerque.json
@@ -1,0 +1,11 @@
+{
+    "coverage": {
+        "country": "USA",
+        "state": "SW",
+        "city": "Albuquerque"
+    },
+    "data": "http://coagisweb.cabq.gov/arcgis/rest/services/locators/CABQ_Composite/GeocodeServer",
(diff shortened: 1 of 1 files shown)


migurski:
Mind renaming to `us-nm-albuquerque.json` to fit the existing naming structure?

geobrando:
@migurski This was submitted via the web app and we likely need to make further improvements.  Currently it's constructed so that it's applicable to any nation and includes text boxes that ask users for `country` `region` and `city`.  It's unclear that in the context of the US `region` is asking for the state.

@PaulCrickard Thanks so much for the contribution.  Unfortunately, we're not able to pull proper addresses from Esri geocode servers. I'll have a look and see and if there's a mapserver or featureserver with addresses.

geobrando:
Couldn't find an address layer on that server.  But it looks like we already have Albuquerque addresses at [us-nm-albuquerque.json](openaddresses/sources/us-nm-albuquerque.json).

Thanks again @PaulCrickard and please be on the lookout for additional sources to contribute.

PR #946, opened 2015-03-28
Address data for Ferrara municipality (Italy)

sabas:
Adding/Updating it-er-ferrara.json


sabas:
It's a shapefile, the actual number is splitted in two fields, NUMERO and ESPONENTE (ESPONENTE can be null)

sabas:
I think this json is now complete, let me know if I missed something.

PR #263, opened 2014-07-09
Add Riverside County, CA, USA

daguar:
I had trouble opening the data file (got some strange zip errors) but this is what's posted on their site.

PR #947, opened 2015-03-28
Address data for Biella (Italy)

sabas:
Adding/Updating it-lo-biella.json


sabas:
In this case I need assistance, on the shapefile there's the address number (CIVICO), and the way is described by its ID. The way name is in the CSV file, so it would need a join between the DBF and the CSV.

geobrando:
Hey @sabas. By "way" I assume you're talking about the street name.  The current processing pipeline is not able to perform table joins.  But if you're able to join the address names to the shapefile we could host the data.  If not, you could provide a link to the csv here in case someone else wanted to tackle it.

sabas:
Yes, I can do it, the csv is inside the zip file.. You'd need to point at a fixed host or I can upload it somewhere?

geobrando:
@sabas If you host it somewhere like a Dropbox account and provide the link the maintainers can pull it into the Amazon S3 bucket.

sabas:
@geobrando here it is
 http://toolserver.openstreetmap.it/openaddresses/biella_joined.zip

I made a field NOME_VIA with the street name (CIVICO has the housenumber)

geobrando:
@sabas Great. Could you update the `street` field with the new name?

Would someone with credentials to the S3 like to take this and update the `data` field with the new path? You could also take care of #263 while you're at it.  @kelvinabrokwa @ingalls @sbma44

sabas:
Done!

iandees:
Here's the URL:

http://data.openaddresses.io.s3.amazonaws.com/cache/it-lo-biella.zip

sabas:
I made it look like the polish file

migurski:
:+1:!

geobrando:
Still not sure that's correct but we should just accept this.   Thanks @sabas

PR #954, opened 2015-03-28
Update US conform info

migurski:
Updated sources in CA, CT, IA, IL, IN, KS, KY, MD, MS, MN, and NC.

### DIFF
--- a/sources/us-nc-craven.json
+++ b/sources/us-nc-craven.json
@@ -11,7 +11,7 @@
     },
     "conform": {
         "type": "shapefile-polygon",
-        "split": "PROP_ADD"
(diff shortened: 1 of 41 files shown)


sources/us-nc-craven.json
```diff
@@ -9,6 +9,14 @@
         "state": "nc",
         "county": "Craven"
     },
+    "conform": {
+        "type": "shapefile-polygon",
+        "split": "PROP_ADD"
```
iandees: Missing a comma here.

migurski: Thanks, just pushed a fix.

PR #955, opened 2015-03-29
Contributing to: .json

evaryont:
Adding/Updating .json

### DIFF
--- a/sources/.json
+++ b/sources/.json
@@ -1,0 +1,12 @@
+{
+    "coverage": {
+        "country": "",
+        "state": "",
+        "city": ""
+    },
+    "data": "http://eservices.scottsdaleaz.gov/dmc/shape/sampledata/Streets.zip",
+    "website": "http://www.scottsdaleaz.gov/maps/sampledata",
+    "attribution": "",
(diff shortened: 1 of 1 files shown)


evaryont:
Woah, that was unintentional. First time filling out the contribution form, didn't realize what the fields all mapped to.

This is for the city of Scottsdale, AZ in the USA.

geobrando:
@evaryont No worries.  Unfortunately, this data appears to be street centerlines and not address data.  Looks like their actual [address data](https://eservices.scottsdaleaz.gov/eservices/dmc/ProductIntro.aspx) is only available for $500+processing fee.  But thanks for contributing.

evaryont:
What is the policy for data like this? I don't mind collecting the money to purchase it and rehosting it (assuming that Scottsdale GIS would allow that), but it would defintely be only a one-time situtation.

iandees:
We haven't started paying for data yet, but I imagine we'll have to soon if we want to increase our coverage.

In this particular case we'd have to work out special arrangements because their disclaimer says the data cannot be redistributed after purchase.

PR #959, opened 2015-03-31
ca-on-ottawa.json - Updated source to shapefile, added conform.

MAL315:


MAL315:
This is where I live so I'm really hoping this one will work :).. if there's anything missing please let me know.

PR #965, opened 2015-03-31
ca-bc-kelowna.json - Updated file path/conform

MAL315:
The zip file has a folder called "LegalParcelsSHP.shp" and inside that folder is another file called LegalParcels.shp.
Conform in .dbf file says StNo and RdName. Not sure if the type shapefile-polygon needs to change.


iandees:
Looks like you're missing a comma there. And the "file" goes inside the conform object, I think.

MAL315:
It should be ok now :)..

PR #988, opened 2015-04-11
Update us-mn-hennepin.json

geospatialem:
Updated Hennepin County, Minnesota by adding notes/metadata resources and changing the associated fields (lon, lat, street, postcode).

### DIFF
--- a/sources/us-mn-hennepin.json
+++ b/sources/us-mn-hennepin.json
@@ -13,10 +13,7 @@
     "type": "http",
     "compression": "zip",
     "website": "http://gis.hennepin.opendata.arcgis.com",
-    "note": {
-        "metadata": "https://www.arcgis.com/home/item.html?id=aefaf3090e8c49b1b08eb6fb77ba2d9e",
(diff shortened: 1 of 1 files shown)


iandees:
Hi Kitty! Thanks for the update to the source for my favorite county in the US :smile:. The build failed because there's no comma at the end of the "metadata" line. Try fixing that and see if it makes the build go green.

geospatialem:
I'm not sure why the build is still failing.. any ideas, @iandees?

iandees:
I don't think it likes the `notes` field. Can you put all that stuff in a single string instead of an object?

geospatialem:
Sorry about the long commit trail, @iandees. 

I'm wondering if we can create another readme file that can help describe the field names that are supported. It took me a while to get the process down *(sorry about that)* but it may be beneficial for others as well. What do you think?

iandees:
The fields are described in [CONTRIBUTING.md](https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#optional-tags), which isn't particularly obvious. It would make sense to point to it from the readme.

PR #992, opened 2015-04-17
added sources for Trento and Rovereto - Italy

napo:


geobrando:
Hi @napo . The `srs:` field should be placed in the conform object.  But since your shapefile appears to have a `.prj` file, you should omit it.  Or did you include it because the `.prj` is not valid?

PR #997, opened 2015-05-01
Add conform object for Chisago County MN

mmdolbow:
They also have an x and y field, but with the weird SRS I thought using the lat/longs would work better. Looks like they have several hundred addresses with no lat/long, but that's out of over 21K addresses. And without a clear EPSG for Esri's 103721 srs, not sure what the lesser evil is.

### DIFF
--- a/sources/us-mn-chisago.json
+++ b/sources/us-mn-chisago.json
@@ -9,8 +9,19 @@
         "state": "mn",
         "county": "Chisago"
     },
+    "conform": {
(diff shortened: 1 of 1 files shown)


iandees:
When we scrape ESRI servers we ask them to reproject the geometry to 4326 for us, so there's no problem with SRS here.

mmdolbow:
(Slaps forehead)...I knew that was possible. Sweet!

PR #999, opened 2015-05-04
Update Lexington with newer address point source

hampelm:


iandees:
The CSV file is not compressed. Could you remove the `compression` attribute?

iandees:
Thanks!

PR #1010, opened 2015-05-22
Update us-tx-san_antonio.json

kreed:
Depends on openaddresses/machine#107


kreed:
@iandees Should be good to merge now

PR #1015, opened 2015-05-28
Create us-tx-odessa.json

kreed:
I'm not really sure how to name sources like these where city-provided data covers the whole county. Is this approach OK or should I be doing it differently?


iandees:
This works fine. The name of the source file isn't too important.

PR #1017, opened 2015-05-28
Update us-tx-dallas.json

kreed:


iandees:
Looks like the conform object is missing the lat and lon attributes.

We probably don't need to require them on CSV (we should fall back to `OA:y` and `OA:x` if they aren't specified), but can you add them to be explicit for now?

kreed:
Yeah, I guess the machine is OK with omitted lat/lon for ESRI sources but the tests here don't like it.

PR #1026, opened 2015-06-06
Contributing to: canada.json

eruci:
Adding/Updating canada.json


sbma44:
Alas, this is ODbL. For the time being, at least, we are striving to keep OpenAddresses content limited to data that is licensed as CC-BY or better.

eruci:
That’s fine. I’ve upgraded the license.

Cheers,
e.

PR #1028, opened 2015-06-07
Fixed conform tags in Poland

migurski:
Removed incorrect field merge and added postcode, region, and city tags.

### DIFF
--- a/sources/pl-dolnoslaskie.json
+++ b/sources/pl-dolnoslaskie.json
@@ -13,11 +13,10 @@
     "conform": {
         "srs": "EPSG:2180",
         "lon": "X",
-        "merge": [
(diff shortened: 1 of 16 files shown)


sbma44:
@migurski my recollection is that the merge was there because a substantial portion of the database does't have a street address. merging the village name into that field seemed to be the best way to produce useful geocoding behavior.

migurski:
Huh, okay. If we’re producing output with city fields, would that be enough?

migurski:
We’re pretty close to final output with city names, so I’m going to go ahead and let this through.

sbma44:
sounds good. I'm facing a similar situation with the Czech address set right now. Plenty of municipalities without street names.

PR #1031, opened 2015-06-17
KS Counties

ingalls:


migurski:
Are these coming from the counties themselves?

ingalls:
@migurski indirectly. I've added the website field in a pull request.

They're part of the statewide ORKA dataset (http://jade.kgs.ku.edu/orka2/CoSelect.aspx)

PR #1040, opened 2015-06-24
Add KS Sources

ingalls:
cc/ @iandees @migurski

### DIFF
--- a/sources/us-ks-Wavaunsee_county.json
+++ b/sources/us-ks-Wavaunsee_county.json
@@ -1,0 +1,18 @@
+{
+    "coverage": {
+        "country": "us",
+        "state": "ks",
+        "county": "Wabaunsee"
+    },
+    "conform": {
+        "type": "csv",
+        "number": "add_num",
+        "merge": ["road_pre", "road_name", "road_type"],
+        "street": "auto_street",
+        "postcode": "zipcode",
+        "lon": "x",
+        "lat": "y"
+    },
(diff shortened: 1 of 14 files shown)


migurski:
Seeing some failures on these, like:

* http://data-test.openaddresses.io.s3.amazonaws.com/runs/205/output.txt
* http://data-test.openaddresses.io.s3.amazonaws.com/runs/201/output.txt

etc.

iandees:
I think you need to specify `OA:x` and `OA:y` for `lon` and `lat`, respectively. I'm not sure if the processing steps will guess that automatically.

ingalls:
@migurski how does one know what the run # from the PR?

ingalls:
@iandees seems to be working as expected: http://data-test.openaddresses.io.s3.amazonaws.com/runs/225/output.txt

migurski:
@ingalls for the moment, you’d go to https://github.com/openaddresses/openaddresses/settings/hooks/5060155 and look at responses from recent deliveries, which would send you here: http://ci.openaddresses.io/jobs/24958d54-78ea-4811-be46-0d157c2702bf

That runaround is because I’m just wrapping up testing on the CI service. Pretty soon, results would show up alongside the ones from Travis on this page right here.

migurski:
Turned it on. Thomas County looks to be failing because of [corrupt data coming back from ArcGIS](http://data-test.openaddresses.io.s3.amazonaws.com/runs/282/output.txt). It worked okay previously, so I’m going to merge.

PR #1042, opened 2015-07-01
Contributing to: brasil.json

PauloLuan:
Adding/Updating brasil.json


migurski:
Can you remove `"compression": "text"`?

PR #1043, opened 2015-07-02
added brazilian addresses

PauloLuan:


iandees:
Thanks for your contribution! It doesn't look like the data you linked to includes any location information for the address. This is a very important part of our database and at this point we can't use new data sources without it.

sergiyprotsiv:
There is a list of all 70+ million addresses in Brazil at http://www.ibge.gov.br/english/estatistica/populacao/censo2010/cnefe/default_cnefe.shtm

From what I understand, they add coordinates to rural addresses, while the urban ones are linked to census enumeration areas that usually cover a few blocks within a city. The enumeration blocks are here:
http://www.ibge.gov.br/english/geociencias/default_prod.shtm (look for Mesh of the Enumeration Area - 2010)

Not sure if such geographic precision is sufficient, but at least it is one way to check the validity of other sources.

PauloLuan:
I removed the data from my repository because I don't know the license of the data. I will try to find another official data in order to put it on another Pull Request. Thanks in advance, best regards.

PR #1046, opened 2015-07-04
Update us-in.json

iandees:


migurski:
The [supposed zip file](http://opendata.arcgis.com/datasets/29f37aa89f3a4bdabe5e5eb803768795_0.zip) is actually a JSON thing telling when the zip file might become available:

    {"status":"processing","processing_time":376.881,"count":6550}

[Machine code doesn’t like it](http://data.openaddresses.io.s3.amazonaws.com/runs/885/output.txt).

Not really sure how to account for this, unless we’re okay with looking at the `opendata.arcgis.com` domain name and extrapolating from there. Maybe @ajturner has an idea.

iandees:
Yea, I thought I had triggered the generation of the Shapefile cache last night, so maybe the caching didn't last very long? Maybe it didn't finish.

I'll change it over to using our ESRI scraper.

ajturner:
@migurski I gave a [long explanation of the JSON response here](https://github.com/openaddresses/openaddresses/issues/889#issuecomment-115421374) - and am very open to what is a more 'typical' way to handle async file processing consistently across request types (particularly for oddities like Shapefile zip)

Simply put for this issue - HTTP status codes trump, check the 202 until it is 200. These are currently generated whenever the data are updated (if recent server version) or nightly (which is this case) to ensure recent data.

That will be improving in the future and all feedback is welcome.

migurski:
Ah it’s a 202, that’s helpful. I’ll see what I can do with that.

iandees:
This particular one has been going for at least a couple days now (it's 3M rows). @ajturner, what happens when it takes that long? Will the cache be invalidated by the time it's done?

migurski:
Hm—it needs to be ready within a couple hours for OA, or else it will fall afoul of the three hour time limit we have for sources.

dmfenton:
If it's going for three days, something has gone wrong. In this case, underlying server has failed. I'm digging in to pinpoint where it failed and we didn't send a 504 response. I'll also get in touch with the provider and see if we can help them get this working better.

chelm:
FYI I've been watching this little (not little at all) Indiana address service churn on our production servers for days...The service is having a hard few weeks to say the least. Let me see if I can dig up any other info. 

Also it being a "MapServer" is not favorable. The types of requests that are made to scrape Mapservers are not optimal.

chelm:
The service has ~3 million features that are being requested in chunks of 50 features at a time. So it takes around 61k requests to get this particular service :)

The reason its requesting 50 features at a time is due to the service's "maxRecordCount" being set to 50. Joy. 
http://maps.indiana.edu/ArcGIS/rest/services/Infrastructure/Streets_Address_Points_IDHS/MapServer/0?f=json

iandees:
@chelm when we scrape this endpoint we don't get any geometry out of the data. Is your scraper pulling it down too or is it just bad in their database?

dmfenton:
@iandees what do your requests look like?

Also @chelm is totally right, they just have this server configured poorly to support downloads. We're getting in touch with the people in charge to see if we can help.

iandees:
It would be great to get these folks talking, but I do want to point out that they offer a FileGDB download. Our stack just doesn't support it yet (see https://github.com/openaddresses/machine/issues/150) and the REST endpoints offer more metadata than the GDB.

iandees:
Oh, also the REST endpoint seems to have newer data, but I haven't verified that.

chelm:
@iandees I'd be curious to see the data discrepancy between the service and the fgdb.  

@dmfenton: @iandees beings up a good point: where are the geometries in requests like this: http://maps.indiana.edu/ArcGIS/rest/services/Infrastructure/Streets_Address_Points_IDHS/MapServer/0/query?objectIds=2514763%2C2514768&outFields=*&returnGeometry=true&orderByFields=&f=json

Are there just no geometries?

iandees:
@chelm @dmfenton The GDB download is here: http://maps.indiana.edu/download/Infrastructure/Streets_Address_Points_IDHS.zip

It has entries like this:

```
OGRFeature(County_Address_Points_IDHS_IN):3582
  FID (Integer) = 3588
  SOURCEID (String) = AddressPointsFeature
  ORIGIN (String) = Adams County
  LOADDATE (DateTime) = 2014/07/26 00:00:00
  ADDRESS (String) = 2280 N 100 E
  NUMBER (String) = 2280
  SUFFIX (String) = E
  ADD_TYPE (String) =
  PREFIX (String) = N
  ADD_NAME (String) = 100
  ADD_CITY (String) = DECATUR
  COUNTY (String) = ADM
  STATE (String) = Indiana
  ZIPCODE (String) = 46733
  CNTY_ID (Integer) = 1
  FGEO_ADD (String) = 2280 N 100 E
  UNITTYPE (String) = (null)
  UNITID (String) = (null)
  POINT (675657.2313232421875 4516145.8604736328125)
```

...so there are geometries *somewhere*, but maybe the REST endpoint is incorrectly set up?

chelm:
Thanks @iandees We'll poke the server admin and see if we can get the service tweaked.

iandees:
Closing this for now.

PR #1049, opened 2015-07-05
Update us-wi-sauk.json

iandees:


ingalls:
Implemented: https://github.com/openaddresses/openaddresses/blob/5b19dfe53b2069259badd5e9706111cb02dcf015/sources/us/wi/sauk.json

PR #1054, opened 2015-07-06
Update us-ms-desoto.json

iandees:


migurski:
Looks like the fields have changed and we no longer need to split one.

PR #1057, opened 2015-07-08
Contributing to: united_sates-dc - district of columbia-washington.json

DavidYJackson:
Adding/Updating united_sates-dc - district of columbia-washington.json


ingalls:
Hey @DavidYJackson we actually already have data for this source. See: https://github.com/openaddresses/openaddresses/blob/master/sources/us-dc.json

ingalls:
A list of sources and their data can be found: http://data.openaddresses.io/runs/1436335974.402/index.html

PR #1058, opened 2015-07-08
Contributing to: united_sates-dc - district of columbia-washington.json

DavidYJackson:
Adding/Updating united_sates-dc - district of columbia-washington.json


ingalls:
Duplicate of #1057

PR #1061, opened 2015-07-09
Create us-mn-scott.json

mmdolbow:
Adding, need to test merge capability


mmdolbow:
Darn, let me try that again.

PR #1063, opened 2015-07-09
Update us-mn-scott.json

mmdolbow:
Trying this again, think I've got it now. This service really breaks up the address, so we have a lot of options.


migurski:
Looks okay to me!

scott

PR #1073, opened 2015-07-12
Add a couple Minnesota sources

iandees:


migurski:
Some weird coordinates with that St. Louis county data.

us-mn-st_louis

PR #1074, opened 2015-07-13
skip us-mn-scott, add kandiyohi

mmdolbow:
Looks like we might need to wait a bit for this one to settle. There is a new "OpenData" folder on the root of the endpoint, which is encouraging.

### DIFF
--- a/sources/us-mn-kandiyohi.json
+++ b/sources/us-mn-kandiyohi.json
@@ -1,0 +1,24 @@
+{
+    "coverage": {
+        "US Census": {
+            "geoid": "27067",
+            "name": "Kandiyohi County",
+            "state": "Minnesota"
+        },
+        "country": "us",
+        "state": "mn",
+        "county": "Kandiyohi"
+    },
+    "website":"http://gis.co.kandiyohi.mn.us/maps/kandigis/",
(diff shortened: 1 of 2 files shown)


mmdolbow:
The scott file has failed, which is why I want to skip it. What do I need to do to make this pull request happier?

iandees:
Get one of us to merge it :smile:. Thanks! Hopefully we can get Scott County back.

iandees:
Looks like they've moved their stuff behind a proxy:

http://gis.co.scott.mn.us/Proxy/proxy.ashx?http://gis.co.scott.mn.us/scgis/rest/services/Property_Info/PUBLIC_PARCEL_APP_RW/MapServer/25

The proxy uses some ESRI API token to gain access to the backing ESRI endpoint. I don't think our code will handle that as-is.

mmdolbow:
There is a lot of peer pressure in the metro to open up the data, and I hopefully will hear about it if it happens. Thanks.

mmdolbow:
Yes, the proxy establishes a login, and once you have a session open you can view the REST endpoint. But that session expires and then you're toast. However, still holding out hope that "OpenData" folder might have a service in it someday.

PR #1076, opened 2015-07-16
Contributing to: estonia.json

tkardi:
Adding/Updating estonia.json


migurski:
`http://xgis.maaamet.ee/adsavalik/xads?FoF6Eqxx` doesn’t look to be a zip file. Is it a temporary link that’s expired? Does the license allow us to cache a copy?

tkardi:
woops, yes, it seems to be a temporary link which is generated upon opening up http://xgis.maaamet.ee/adsavalik/ads. 
There is no concrete link to a licence from the Land Board but the idea behind this site is that you can download the whole address dataset in csv format. So caching should be OK, but the timestamp for the dataset should be noted down aswell. The right link to download ("laadi alla") should be the one where Nr = 1 and the KOV (named "LG" in eng version) field is empty.

migurski:
Okay, I have a copy. Can you change the URL in the source to http://data.openaddresses.io/cache/es.zip, and add a note about ISO-8859-1 text encoding like in [fr/la-réunion.json](https://github.com/openaddresses/openaddresses/blob/master/sources/fr/la-réunion.json)? I will upload this copy to our cache shortly.

tkardi:
I think the right ISO two letter code is ee for Estonia. Es should be Spain. Yes i can add the note.

migurski:
:+1:, I sloppy-copied and fixed the comment.

tkardi:
Looking at the file, it seems you got the addresses only for the city of Tallinn (KOV="Tallinna linn"). On http://xgis.maaamet.ee/adsavalik/ads there's two entries with Nr == 1 - the whole country (incl Tallinn) is covered in the file where the value for KOV (LG in the English version) is blank.

migurski:
Okay, since I don’t read Estonian can you get the correct file and make it available someplace like Dropbox?

tkardi:
Sure thing no prob. But can you first have a look at issues [csv data filtering #1078](https://github.com/openaddresses/openaddresses/issues/1078) and [conditional merge of csv fields #1077](https://github.com/openaddresses/openaddresses/issues/1077) aswell, what do you think? i can do some processing then aswell if needed.

tkardi:
here you go: https://drive.google.com/file/d/0BwPIVpJhp8rHZGYzZHVGdkdQVGM downloaded from http://xgis.maaamet.ee/adsavalik/ads

migurski:
Got it!

PR #1079, opened 2015-07-17
Add barton

ingalls:
Extracts addresses from accessors PDF using build scripts found here: https://github.com/ingalls/barton_ks_addresses

### DIFF
--- a/sources/us-ks-atchison_county.json
+++ b/sources/us-ks-atchison_county.json
@@ -1,5 +1,6 @@
 {
     "coverage": {
+        "US Census": {"geoid": "20005", "name": "Atchison County", "state": "Kansas"},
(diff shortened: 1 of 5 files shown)


ingalls:
hmm @migurski do you know what the error is being caused by off the top of your head or should I go digging?

migurski:
Looks like [a stray “:” in one of the column names](http://data.openaddresses.io.s3.amazonaws.com/runs/2371/output.txt).

ingalls:
Sigh, thanks @migurski I was thinking it was a problem with the source data, my sleepy brain didn't even consider it being the json file. Fixed.

migurski:
:+1:

PR #1082, opened 2015-07-18
Updating a bunch of Texas sources

iandees:


geobrando:
Failures in this PR:

* Austin source no longer valid, could be replaced by Socrata portal data: https://data.austintexas.gov/api/geospatial/bpa2-q2tj?method=export&format=Shapefile

* Pflugerville url changed, now http://services.arcgis.com/qDOnMQ2aoGtLgVxL/ArcGIS/rest/services/Addresses_6_2015/FeatureServer/0

* Two additional failures were valid Esri sources failing for unknown reasons and could be attempted again.

Issue #378, opened 2014-09-07
Conform columns


MiniCodeMonkey:
Hey,

It looks like processed files that have been through openaddresses-conform have the following fields: `lat`, `lon`, `number`, `street`. (Correct me if I'm wrong, this might only be the case for the sources that I'm looking at).

In order to improve the usefulness of the data, I would like to open up the discussion for considering adding additional fields to the data source json files.

US Addresses generally conform to the following structure:
* Street number
* Directional prefix (e.g. N or S)
* Street name (e.g. Quincy)
* Street suffix (e.g. Street, Court, etc.) - USPS has a [full list of street suffixes](https://www.usps.com/send/official-abbreviations.htm)
* Directional suffix (e.g. N or SE)

Right now all of these seems to be stored in only two fields -- street number and street name even though many data sources have addresses stored in multiple fields.

Here's an example from the `us-va-arlington` data source.

```json
{
      "type": "Feature",
      "properties": {
        "OBJECTID": 15163,
        "STNUM": 2765,
        "STCODE": "NQCYST",
        "STRTDIR": "N",
        "STRTDIRPREFIX": "N",
        "STRTNAME": "QUINCY",
        "STRTTYPE": "ST",
        "STRTDIRSUFFIX": "",
        "FULL_ADDRESS": "2765 N QUINCY ST",
        "STATUS": "VERIFIED",
        "UNITCOUNT": 1,
        "STNAME": "N QUINCY ST",
        "UNIT": null,
        "UPDATE_DATE": 1156291200000,
        "ZIP5": 22207
      },
      "geometry": {
        "type": "Point",
        "coordinates": [
          -77.1007728,
          38.908486
        ]
      }
    }
```

Would it make sense to add the above standardized street components in the json source files when available in addition to the full street name?

Another component is the zip code which could be added as well.

Mathias

geobrando:
Has there been any more discussion about additional fields in the conform object to capture other address elements?  Even if there's no plan in place to process these, shouldn't we at least be capturing these in new sources as we go along as previously mentioned?

ingalls:
@geobrando yes, they are part of the spec here: https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md they just aren't being actively processed as of yet.

geobrando:
[...]

PR #1084, opened 2015-07-18
add coshocton county, ohio

skorasaurus:
Before merging, I have a few questions and notes: 

* There are several cases where a particular address is in both a Township and a City. See [this wikipedia entry for more information](https://en.wikipedia.org/wiki/Civil_township#Midwestern.2C_central.2C_and_western_states). 
Some addresses are in a township (column COMMUNITY) and a city (CITY) ; some are only in a township; In one instance, they are in the same CITY and COMMUNITY (West Lafayette)

* Street names and suffixes are in separate columns. In my 2nd commit, I added the street suffixes, based on the syntax in a couple existing entries - [us-tx-brazos](https://github.com/openaddresses/openaddresses/blob/4ecffa3872b925a3a35951ee318caa334402b9aa/sources/us-tx-brazos.json) and us-oh-fairfield). However, this syntax is missing a subproperty(?) named 'separator' described (https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#advanced-merge) (Related #378). Which syntax should I use?

     Related to that, there's two additional columns to further describe the address number (one column listing the Apt/Bldg/Lot and another column listing Lot number). I didn't add them yet, waiting to hear your advice. 
 
* There's a fair amount of rows that have null addresses; The only columns that have values in these rows are AVSHPID (a unique identifier for each row) and COMMUNITY.

### DIFF
--- a/sources/us-oh-coshocton.json
+++ b/sources/us-oh-coshocton.json
@@ -1,0 +1,19 @@
+{
+    "coverage": {
+        "US Census": {"geoid": "39031", "name": "Coshocton County", "state": "Ohio"},
+        "country": "us",
+        "state": "oh",
+        "county": "Coshocton"
+    },
+    "conform": {
+        "number": "ADRSNUM",
+        "street": "ROADNME",
+        "postcode": "ZIP",
+        "type": "shapefile",
+        "accuracy": 5
+    },
+    "data": "http://www.coshoctoncounty.net/files/1413456603Address.zip",
+    "website": "http://www.coshoctoncounty.net/agency/taxmap/GISfiles.php",
+    "type": "http",
+    "compression": "zip"
+}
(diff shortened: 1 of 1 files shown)


migurski:
Looks legit.

coshocton

How would the USPS view addressing from the Township vs. the City?

The merge syntax you chose is probably fine.

migurski:
(If you want to spot-check [the resulting data](http://data.openaddresses.io.s3.amazonaws.com/runs/2766/us-oh-coshocton.zip) I can wait to merge)

skorasaurus:
checking now. will edit.

skorasaurus:
The [parser](https://github.com/openaddresses/machine/blob/f55a46c7128c9924d8d2841cbbde90f9b3ee9717/openaddr/expand.py#L29) is expanding out CR (which is 'County Route') to 'Crescent' :/ 

selection_004

migurski:
Yuck, sorry. There’s an [ongoing discussion](https://github.com/openaddresses/machine/issues/32) about not messing with the names, but I’ll merge this for now.

PR #1091, opened 2015-07-22
Contributing to: brazil-sp-são_paulo.json

tiagofassoni:
Adding brazil-sp-são_paulo.json

The dataset is by no means perfect, and probably some custom code will
have to be made in order to make it work correctly. But let's see how it
goes.


geobrando:
Duplicate of #1092

PR #1092, opened 2015-07-22
Adding brazil-sp-são_paulo.json

tiagofassoni:
The dataset is by no means perfect, and probably some custom code will
have to be made in order to make it work correctly. But let's see how it
goes.

### DIFF
--- a/sources/br-sp-são_paulo.json
+++ b/sources/br-sp-são_paulo.json
@@ -1,0 +1,18 @@
+{
+    "coverage": {
+        "country": "br",
+        "state": "SP",
+        "city": "São Paulo"
+    },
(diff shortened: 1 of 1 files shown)


migurski:
The full file page looks to be "GEOLOG/LAYERS_GEOLOG/DEINFO_SEGTOS_212.shp".

tiagofassoni:
Yes, it is. But I've read more about it, and I found a couple of issues more:

1) the dataset uses address interpolation. Openaddresses doesn't seem to support them.
2) there are a lot of fields in brazilian portuguese. I don't know the json options to convert them.

slibby:
Given the lack of point features, I'm going to close this until there's better data available or until OA supports street-range data.

PR #1099, opened 2015-07-24
Contributing to: us-md-frederick.json

eczajk1:
Adding/Updating us-md-frederick.json


migurski:
Looks good—I’m not seeing any numbers in the output, maybe it should be `ST_NUM` instead of `ST_NUMBER` ?

eczajk1:
Yup, you're right

migurski:
Looks good!

frederick

PR #1100, opened 2015-07-26
Czech addresses

sbma44:
8.4M addresses from 18,000+ GML files. impossible without lots of help from @mormegil-cz in https://github.com/openaddresses/openaddresses/issues/12

image
image
image


migurski:
Wowza.

PR #1101, opened 2015-07-26
Fix Maricopa County, AZ (US) geoid

daguar:
Looks like it should be 04013 per http://www2.census.gov/geo/docs/reference/codes/files/national_county.txt


migurski:
:+1:

PR #1103, opened 2015-07-27
fix to the coverage object

tkardi:
- removed "state" and "city" tags
- added the "ISO 3166" tag


migurski:
Looks great!

ee

tkardi:
Looks familiar :) I'll see what i can do about city/district/region aswell.

Issue #735, opened 2014-11-27
Merging of datasets


colemanm:
The data available for [Escambia County, FL](http://www.escpa.org/MapMain.aspx) is available, but in two parts:
- [Parcels](http://www.escpa.org/snapshots/downloads/County/shapefile/Parcel.zip)
- [Owner mailing data](http://www.escpa.org/files/misc/ecpa_mail.zip)

The parcel data doesn't contain the address info, but has geometry (parcel polygons). The mailing data is an address list with no geo information, but could be joined to the parcel data with the folio number. This seems like a challenging thing to build a processing pipeline around. Any thoughts?

ingalls:
Closing as stale

Issue #177, opened 2014-05-02
Chilliwack BC


ingalls:
Unfortunately after processing it doesn't look like the parcel file has any address data.

ingalls:
Closing as stale - Feel free to open a new issue if someone is actively going after these.

PR #1104, opened 2015-07-27
Contributing to: it-taa-rovereto.json

napo:
Adding/Updating it-taa-rovereto.json

### DIFF
--- a/sources/it-taa-rovereto.json
+++ b/sources/it-taa-rovereto.json
@@ -1,0 +1,12 @@
+{
+    "coverage": {
+        "country": "it",
+        "state": "taa",
+        "city": "rovereto"
+    },
+    "data": "http://www2.comune.rovereto.tn.it/servizionline/extra/opendata/Civici.zip",
+    "website": "http://www2.comune.rovereto.tn.it/servizionline/extra/opendata/",
+    "attribution": "",
+    "type": "http",
+    "compression": "zip"
+}


migurski:
Thank you @napo! The data [doesn’t look to contain street names](http://data.openaddresses.io.s3.amazonaws.com/runs/5544/sample.json) —is it valid address data?

migurski:
Duplicate of #1106.

napo:
@migurski you are right.
The dataset of Rovereto contains only the id of the street names, if you want have the complete dataset, you have to join the other table.
http://www2.comune.rovereto.tn.it/servizionline/extra/opendata/Strade.zip

The dataset of Trento is nice but, there is a problem:
you need to filter the data because the dataset contains each place where is a label with the address number. This means that, in some buildings, there are more as one label. Eg. a label on the main entrance, another on the gate or the secondary entrance.

geobrando:
Hello @napo. Thanks for the contribution.  Since the data from both sources need processing before they can be included, I'm closing #1106.  OA would certainly appreciate any effort to process these.  OA could host the files that you produce, or you could host them somewhere and script updates as offered. 

[Machine](https://github.com/openaddresses/machine) is currently able to perform tasks such as joining/splitting fields, generating polygon centroids, and srs transformations.  In my personal opinion, it would be useful to add functionality to machine and the source schema expanded to allow joins between multiple data files, filtering, etc.  But I don't see that becoming a reality any time soon.  You'll see others raising a need for this [here](https://github.com/openaddresses/openaddresses/issues/735) and [here](https://github.com/openaddresses/openaddresses/issues/177).

If you plan to submit new sources, please check out the new [CONTRIBUTING.md](https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md) as there have been some changes to the source JSON schema, filename convention, and the folder structure.

PR #1106, opened 2015-07-27
new version for Trento & Rovereto Italy

napo:


migurski:
The Trento data looks great! The Rovereto data [doesn’t seem to have streets](https://github.com/openaddresses/openaddresses/pull/1104#issuecomment-125282418), through.

PR #1110, opened 2015-07-29
Contributing to: ca-on-haldimand.json

MDhald:
Adding/Updating ca-on-haldimand.json

### DIFF
--- a/sources/ca-on-haldimand.json
+++ b/sources/ca-on-haldimand.json
@@ -2,10 +2,16 @@
     "coverage": {
         "country": "ca",
         "state": "on",
-        "city": "haldimand"
+        "city": "Haldimand"
     },
-    "data": "https://drive.google.com/uc?export=download&id=0B41bzFk-ZijJZk44bjFtcFNtWUE",
-    "website": "http://opendata.haldimandcounty.on.ca/datasets/558a001d161c43458c9bfe763e2a8518_92",
-    "attribution": "Haldimand County",
-    "type": "shapefile"
+    "type": "ESRI",
+    "data": "http://maps.haldimandcounty.on.ca/arcgis/rest/services/StandardBasemapp/MapServer/92",
+    "website": "http://www.haldimandcounty.on.ca",
+    "conform":{
+        "type": "geojson",
+        "number": "STREET_NO",
+        "street": "STREET_NAM",
+        "city": "Haldimand",
+        "region": "Ontario"
+    }
 }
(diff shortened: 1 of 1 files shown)


migurski:
Thank you! FYI, The first error [in the CI log](http://data.openaddresses.io.s3.amazonaws.com/runs/6528/output.txt) is [fixed](https://github.com/openaddresses/machine/issues/171) but I still need to deploy it, possibly later today.

geobrando:
Type field is wrong, but this is one of those Esri portal sources I mentioned in #889.  Looks like the contributor downloaded and saved in their google drive folder to get around it.  If we want to keep, someone with credentials could upload this file to the cache.  There are many Esri sources with the same situation.

slibby:
This could be updated to reference the direct shapefile dowload link, and I think that would pass OK: http://opendata.haldimandcounty.on.ca/datasets/558a001d161c43458c9bfe763e2a8518_92.zip

Alternately, the underlying service could be used as "type": "esri" : http://maps.haldimandcounty.on.ca/arcgis/rest/services/StandardBasemapp/MapServer/92

geobrando:
Hey @slibby.  See the issue that I referenced above for further discussion about the problem of including Esri open data portal sources.  This might pass OK if a PR is submitted now as the file has been recently cached but presents a problem for future runs, depending on the update cycle.  But yes, an `esri` type is not the ideal solution but would work.

slibby:
@geobrando right, for some reason I thought .shp zips played a little better but you're right, the UI handles it. 

There is a new feature that allows open data site administrators to set a dataset as static, which means it caches and holds, rather than clearing it after 24h. This setting would allow it to be crawled properly for OA, if an administrator is willing to enable it and if the dataset is "static enough"

MDhald:
is there any process I can use or do to get our data involved in the Open Addresses?

slibby:
@MDHald Yes! If you'd like to update your contribution .json file to match something like this (updating relevant attributes): 

```
{
    "coverage": {
        "country": "ca",
        "state": "on",
        "city": "Kingston"
    },
    "type": "ESRI",
    "data": "http://maps.cityofkingston.ca/ArcGIS/rest/services/DataCatalog/DataCatalog_Addresses/MapServer/0",
    "website": "https://www.cityofkingston.ca/explore/data-catalogue",
    "conform":{
        "type": "geojson",
        "number": "ADDRESS_NUMBER",
        "street": "STREET",
        "city": "Kingston",
        "region": "Ontario"
    }
}
```

But using the link behind that Open Data layer (a Map Service) instead: http://maps.haldimandcounty.on.ca/arcgis/rest/services/StandardBasemapp/MapServer/92 

Then it should pass the continuous integration checks.

slibby:
@MDhald I have misled you, sorry! Can you remove the "city" and "region" attributes from your "conform" object? These should specify existing "fields" to apply to the output, not static entries, and I was mistaken on that fact.

ingalls:
Fantastic, thanks!

PR #1112, opened 2015-07-31
update us-wa-pacific

jgravois:
added one conform and tidied up another.

thanks for the san bernardino addresses.  used them to help a non-profit with a hyperlocal mailer.

### DIFF
--- a/sources/us-ca-san_bernardino_county.json
+++ b/sources/us-ca-san_bernardino_county.json
@@ -11,9 +11,7 @@
         "city": "CITY",
         "number": "NUMBER",
         "merge": ["STREETNAME", "STREETTYPE"],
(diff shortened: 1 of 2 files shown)


migurski:
Hm, "mail address" might not be the right field to split in `us-wa-pacific`, I think it’s the postal address of the owner: e.g.:

pacific

jgravois:
> I think it’s the postal address of the owner

good catch.  the appropriate field appears to be `SITUS`, but just a heads up, the number and streetname are delimited with a backslash.

screenshot 2015-07-31 13 20 53

migurski:
You’re right—it’s not great but will probably work!

pacific

Some zeroes in SB, but should work.

san-bernardino

PR #1113, opened 2015-08-01
add us-ks-shawnee county

ingalls:


ingalls:
@NelsonMinar @migurski could one of you check this one out when you get a chance? It works with both the node & python versions of `esri-dump`. Otherwise I can load the file into the cache.

iandees:
The source is just timing out.

ingalls:
Yeah I saw that in the log. What makes me suspicious is that
curling/esridumping the source works perfectly when done locally

NelsonMinar:
Huh, weird, the log shows it downloaded stuff from an ESRI endpoint for 25 minutes and then the server went offline. I suspect a transient error with the source server. Now that it's merged we should know more after a few attempts.

PR #1118, opened 2015-08-04

dylanhthomas:


migurski:
Looks like the source is 404?

geobrando:
Confirmed it's a 404 and also already exists as a source https://github.com/openaddresses/openaddresses/blob/master/sources/us/tn/franklin.json

Could not find a replacement.

iandees:
Bummer, we'll have to keep our eyes out.

PR #1119, opened 2015-08-07
add lane county ks

ingalls:
## Lane County KS

- Data is in a non-standard format and must be individually looked up via pin code.
- Tools to replicate can be found here: https://github.com/ingalls/scraper/tree/master/LaneKS

## apply-us-data.py
Updates to print an error to console when it cannot lookup a county name instead of hard fail.

### DIFF
--- a/apply-us-data.py
+++ b/apply-us-data.py
@@ -68,10 +68,14 @@ for path in glob('sources/us-*.json'):
         print info['coverage'], state_fips, state_name, county_names
         continue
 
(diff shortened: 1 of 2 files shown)


iandees:
Can you put the scripts you used in the scripts dir in our repo?

ingalls:
@iandees yeah I can move these over, At the moment I'm still iterating on the scripts. Once I finish up this style of kansas sources I'll move them over.

PR #1122, opened 2015-08-10
Add osage ks

ingalls:


migurski:
:+1: 

osage

ingalls:
@migurski When we start adding parcels I also have the parcel data for a lot of these.

migurski:
Oh that’s great. Are you from Kansas? It seems to be your state of choice.

PR #1124, opened 2015-08-10
Geary ks

ingalls:


ingalls:
I'm going to merge this for now as it doesn't change the current failing source other than adding a correct conform tag.

PR #1135, opened 2015-08-11
Contributing to: south_africa-western cape-cape town.json

sindile:
Adding/Updating south_africa-western cape-cape town.json


migurski:
Can you remove the "compression" tag? I understand that our website update tool introduces it, but it’s not valid.

geobrando:
Another version submitted in #1144, but also has already been added to sources. Can be closed.

https://github.com/openaddresses/openaddresses/blob/master/sources/za/wc/cape_town.json

PR #1140, opened 2015-08-11
Fix broken sources

ingalls:


ingalls:
Well 2/3 working is better than 0/3. Will come back and examine the last fail later.

PR #1144, opened 2015-08-13
Contributing to: south_africa-western cape-cape town.json

sindile:
Adding/Updating south_africa-western cape-cape town.json

### DIFF
--- a/sources/south_africa-western cape-cape town.json
+++ b/sources/south_africa-western cape-cape town.json
@@ -1,0 +1,12 @@
+{
+    "coverage": {
+        "country": "South Africa",
+        "state": "Western Cape",
+        "city": "Cape Town"
+    },
+    "data": "http://www.capetown.gov.za/en/OpenDataCatalogue/_layouts/OpenDataPortalHandler/DownloadHandler.ashx?DocumentName=Street_addresses.zip&DatasetDocument=http%3a%2f%2fcityapps.capetown.gov.za%2fsites%2fopendatacatalog%2fDocuments%2fStreet%2520addresses%2fStreet_addresses.zip",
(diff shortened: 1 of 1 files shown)


slibby:
@sindile I think you'll need a couple changes before this will pass the CI:
1. Set "compression": "zip"
2. Add a "conform": {} object to your JSON file as is laid out here: https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md

"conform": {
  "type": "shapefile",
  "street": "STREET_NM_FIELD",
  "number": "NUMBER_FIELD"
  }

geobrando:
@slibby In case you're not aware, issues beginning with "Adding/Updating" have been created using http://openaddresses.io/contribute/, which was created to lower the barrier for participation to those not comfortable adding fully formed sources via a github PR.

I believe the thought has been to allow others that are more comfortable to correct these when necessary.  But I'm not sure how to do that without first merging the PR.  Also, I don't think a conform object is required to pass CI.

slibby:
@geobrando Cool, did not know but now I do! I agree I don't really know how to edit/help this contribution without just starting my own PR in this case.

sindile:
I used http://openaddresses.io/contribute/

to bring to the attention of Openaddress Community the availability of 
this address data sort. Maybe I can be advised on what I need to do in 
order for the import to succeed.

Regards,
<http://openaddresses.io/contribute/>

Sindile Bidla

geobrando:
@sindile Your work is done. Thank you very much for the contribution.  Someone else associated with the project will take care of it.

geobrando:
Looks like this source has previously been added

https://github.com/openaddresses/openaddresses/blob/master/sources/za/wc/cape_town.json

PR #1149, opened 2015-08-13
Updated notes about new merge syntax

### LINKED ISSUE #1148: Reflect new merge syntax in contributing documentation
Reference https://github.com/openaddresses/machine/pull/180.

migurski:
We replaced the `auto_street` usage in the merge case, but what about the split case? Would it still be `auto_street` in the `street` tag, or should it be `OA:street` or something else?

* [x] Determine correct documentation for `auto_street` with split as well as merge.

Closes #1148.


migurski:
Got the nod from @ingalls.

PR #1157, opened 2015-08-17
updated url

corynissen:


iandees:
Looks good, thanks for updating the URL!

corynissen:
Does "skip": true need to be changed in order for these addresses to be processed?

migurski:
Yes, "skip" will need to be removed.

PR #1170, opened 2015-08-18
Move source files into subdirectories to make navigating easier

iandees:


migurski:
CI did not run on this pull request because the server timed out, I guess because it’s so enormous?

migurski:
Also I see you’re preserving the country code prefixes on the file names, i.e. `xx/xx-region.json`. The France files should be renamed to look like this, too.

ingalls:
Would you both be open for discussion for following the france example and dropping the county code? This seems redundant as it is in the file structure and the `coverage` object of the source.

migurski:
I’m open to that as well.

ingalls:
Also @iandees you'll need to update the node tests to recognize sub-directories. As it stands this change is not actually being tested by travis.

iandees:
Seems fine to me. I can adjust the PR this afternoon.

migurski:
Is this a rationale for using Python in the unit tests?

iandees:
I'm going to close and delete this branch so I can re-do it from master in one commit rather than bounce all over the place.

PR #1172, opened 2015-08-18
Update ca-on-west_parry_sound REST endpoint to new address layer

slibby:


migurski:
If you [add a `conform`](https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#conform-object) this should be complete. Try this? (Hope it works…)

```
"conform": {
  "type": "geojson",
  "number": "CivicNumber",
  "street": "CivicAddress",
  "city": "Municipality"
}
```

slibby:
I added a conform, let's see what the CI says....

migurski:
Sorry it’s taking a while, there’s a Polish Voivodship in line first.

migurski:
wps

ingalls:
@slibby Thanks for getting this in!

slibby:
@ingalls @migurski I noticed in looking at the latest run of failed datasets on data.oa.org that many of the Esri services have ```"type" : "csv"'``` in the conform. This one seemed to work after I changed it to ```"type" : "geojson" ```-- is this the right way of referencing these Esri layers? I see more that still have "csv" included, including one that you "fixed" here which I believe is failing now: https://github.com/openaddresses/openaddresses/blob/master/sources/us-ca-orange.json

ingalls:
@slibby I've been putting them as type csv as technically that is what they are when they reach the conform step of `machine`. The next iteration I do on this will be to abstract this away. At the moment there are several sources where you can put anything you want as the value as it is ignored.

migurski:
The type really shouldn’t be necessary inside the conform, it’s a bit of a mess until we get that out of the spec.

PR #1173, opened 2015-08-18
Rename French sources

iandees:
Remove accents (which were screwing up git/shell), change to underscore instead of dash.


migurski:
St. Pierre et Miquelon [is a skip](http://data.openaddresses.io.s3.amazonaws.com/runs/11436/output.txt), not a real failure.

PR #1176, opened 2015-08-18
Move sources into country/state directories

iandees:


ingalls:
This is awesome, thanks for getting this merged in!

PR #1181, opened 2015-08-19
Updating/Fixing several CO sources

slibby:
REST service went missing, found zipfile

### DIFF
--- a/sources/us-co-sanmiguel.json
+++ b/sources/us-co-sanmiguel.json
@@ -10,7 +10,7 @@
         "county": "san miguel"
     },
     "attribution": "Sanmiguel County",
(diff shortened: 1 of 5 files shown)


ingalls:
This is fantastic! Glad to have another set of eyes on CO!

sources/us-co-sanmiguel.json
```diff
@@ -24,4 +24,4 @@
         ],
```
ingalls: Looks like oa/hooked failed as one of these fields doesn't exist

slibby:
looks like this now conflicts after all the files were moved to subdirectories -- not  sure if it's easier for a repo owner to fix or to cancel the pr and re-add.

ingalls:
@slibby I'll fix these!

ingalls:
All in, thanks!

PR #1199, opened 2015-08-20
test lat/lng/type removal

ingalls:


migurski:
Bummer.

ingalls:
Yeah I didn't really expect it to pass but now I have a sense of what I need to work on next to get this removed.

PR #1200, opened 2015-08-20
Add conform object for Chattanooga and add license.

seabre:
Added as much as I could. I think everything that's needed is here.


ingalls:
Good stuff!

screen shot 2015-08-21 at 11 40 17 am

PR #1202, opened 2015-08-21
Removed reference to previously-deprecated merge tag in docs.

migurski:


ingalls:
Also here: https://github.com/openaddresses/openaddresses/pull/1202/files#diff-6a3371457528722a734f3c51d9238c13R110

migurski:
Oops good catch.

ingalls:
Looks good to me :ship:

PR #1214, opened 2015-08-26
Create montpellier.json

slibby:


iandees:
Looks like geojson to me... wonder where that tripped up.

slibby:
Working on a fix to conform.py for this

iandees:
:heart: Thank you!

slibby:
This should be fixed by https://github.com/openaddresses/machine/pull/196 -- we'll see once it propagates to the ci workers.

slibby:
Something interesting here where hooked-ci fails with a trailing whitespace in attribution, but succeeds once it's removed. Anyways, this should work now, the previous machine PR took care of the .geojson file extension.

migurski:
That is weird, and I’m having difficulty seeing this change would make the difference.

The [last failing log](http://data.openaddresses.io.s3.amazonaws.com/runs/17454/output.txt) shows what looks like a failed download from the data source, while [the successful one](http://data.openaddresses.io.s3.amazonaws.com/runs/17456/output.txt) seems to find it just fine. I wonder if there is an intermittent problem at `opendata.montpelliernumerique.fr`?

migurski:
Weeeeeeird.

PR #1215, opened 2015-08-26
fix wa-mason link to /data

slibby:


iandees:
Ooo does appending `/data` make it actually forward to an S3 link? We have several other sources like this that could use that trick.

slibby:
@iandees Yes, that's how it works for CSV and zipped shapefile items in ArcGIS Online to switch from item details to download. Note the change in the URL from the /item.html to the sharing API path, all you need is the Item ID.

PR #1238, opened 2015-09-09
Evaluating US skipped sources

geobrando:


geobrando:
Further commits coming in later PR. Merge away.

PR #1251, opened 2015-09-17
Add new ga sources

ingalls:


geobrando:
Hey @ingalls 

Can you explain a bit about the lineage of these cached files?  I'm noticing a lot of `0` character fields

ingalls:
These sources are derived directly from parcel data. Many of the parcel in georgia don't seem to have been assigned addresses resulting in null/0 fields.

geobrando:
I figured these fields were missing data.  But were the zeros introduced during your processing of the data?  I see `0`'s in fields for records containing partial addresses and null fields for records without any address data.  As you know, the two are not the same.  Just trying to make sure we're not adding data.

ingalls:
@geobrando The script I have does almost no processing on the actual data. Any `0` or `0 <Street>` fields are exactly what the source data is returning.

PR #1260, opened 2015-09-20
Remove bibb county, al

slibby:
This is actually bibb-county-ga data.


iandees:
Should this be renamed or is the source already there?

slibby:
The other source is already there, as far as I can tell.

PR #1262, opened 2015-09-20
add conform and update north vancouver

slibby:


iandees:
The conform needs to have a lat and a lon.

slibby:
@iandees yeah just realized that, it lacks it and there's no other format on their site, so I am marking it as "skip" for now

geobrando:
@slibby Closing this for now.  Since there's no spatial data there's no reason to add this.  Taking a look at their portal, it looks like you can join the address csv with a building footprint layer using the `asset_id` field.  There's also a cad parcel and parcel centroid file one could link via an intermediate file, but the footprint layer would seem to be the way to go.

slibby:
@geobrando agreed. I emailed the city asking for info and will update if it comes in.

PR #1263, opened 2015-09-20
update iceland coverage

slibby:


slibby:
@iandees this file is a .dsv and the lat/long values are something funky, I think they are incorrect. I'll look over it another time, it'll probably still fail for now.

PR #1264, opened 2015-09-21
Add more ga sources

ingalls:


ingalls:
Fixed tift source, merging.

PR #1265, opened 2015-09-21
Create bologna.json

slibby:


slibby:
I think this is failing because the embedded file is civici.Shp, not .shp or .SHP, which OGR expects maybe?

ingalls:
@slibby Mind trying the `file` attribute on the `conform` object?

Docs here: https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#processing-tags

slibby:
@ingalls that's what I tried in the second commit... I think?

ingalls:
Darn, I'm sorry, yes you did. Looks like a possible problem with machine. I can take a look tomorrow, or @migurski might know what is going on right away.

slibby:
Changed to a skip, issue started here: https://github.com/openaddresses/machine/issues/257

PR #1268, opened 2015-09-21
Add jefferson parish

ingalls:


ingalls:
@migurski @iandees any ideas on this one? No problem running it locally but hooked doesn't seem to like it.

iandees:
This looks like a transient thing with the ESRI host. The "Unable to complete operation" error usually means some query on their end timed out. In this case it was taking too long to respond to a request for a chunk of address points.

We probably could probably solve this by being less aggressive on the requests to ESRI endpoints.

ingalls:
@iandees :+1: I'm going to merge this for now then. I think we should discuss your PR about going back to geospatial queries against the service.

PR #1273, opened 2015-09-22
bernalillo nm

ingalls:


iandees:
(If you're on a roll and have a sec, it would be helpful to update this doc as you find new sources and add them here: https://docs.google.com/spreadsheets/d/1HFm0YbFDC5YKkHKFt89EKdhpyaCCIo_CiDFjBNq4PoI/edit)

ingalls:
Added this source. Just curious, are we back porting OA => Spreadsheet? I was under the impression we were just doing spreadsheet => OA as we can programmatically generate the spreadsheet from our OA sources instead of doing it by hand.

iandees:
I guess I'm reflecting my workflow on you: usually I pick a chunk of the spreadsheet as a guide when I'm searching for new sources and will mark them as completed when I find something. If you're just "free searching" :smile: then nevermind. We can update the spreadsheet based on sources at some point in the future.

ingalls:
@iandees no worries, yeah I haven't used the spreadsheet yet. Usually get on a roll by finding a state acessors page or a 3rd party vendor and going from there.

PR #1276, opened 2015-09-22
Create zurich.json

slibby:
not sure the split will work...we'll see.


ingalls:
screen shot 2015-09-22 at 6 13 55 pm

PR #1284, opened 2015-09-22
fancy lorain regex

ingalls:


ingalls:
Ran out of time - Merging as this is a machine issue and not a source issue.  cc/ @migurski

PR #1295, opened 2015-09-23
ga

ingalls:


iandees:
GA!

ingalls:
ANSI:INCITS 38:2009 13!

iandees:
Can you make sure you're adding docs/code to describe how you're building these zips that end up in data.openaddresses.io?

migurski:
This makes me think we need a way to detect staleness in the data we are hosting ourselves.

ingalls:
@migurski I agree. We have the `year` tag which I added in the original spec to do just that. But with machine this would be far better to automate and then deprecate the `year` tag.

PR #1302, opened 2015-09-23
Add Several Colorado Counties

slibby:


geobrando:
Hey @slibby. Can you add `county` fields to the coverage objects for these?

Also, `split` tags are deprecated so this will pass but it's preferable to use the regex functions.  For most US sources you can basically just copy and paste the "number" and "street" examples found in Contributing.md (which just use the numeric portion as the number and then everything after that as the street)

slibby:
@geobrando -- look good now? Thanks for the info on "split", didn't realize it was deprecated.

ingalls:
Looks good to me!

PR #1304, opened 2015-09-25
Create costilla.json

slibby:
Testing to see if scraping via proxy will work.


slibby:
apparently doesn't work

iandees:
This was the next thing on my list to add to machine's esri scraper.

PR #1314, opened 2015-09-26
Add Oulu, Finland

hannesj:
The city of Oulu has released its address data. Unfortunately the format is a bit more complex, and can't directly be read by openaddresses/machine. That is why I included a file in scripts, which converts the data to a more standard CSV format.


hannesj:
Updated the source by hosting the converted data by myself.

hannesj:
@iandees, could you check this?

hannesj:
thanks

iandees:
Thanks for doing the work on this!

PR #1316, opened 2015-09-27
Update trinity.json

slibby:


ingalls:
Mind explaining this change? The old url seems to still be active `http://navigator.state.or.us/arcgis/rest/services/Projects/County_Harney/MapServer/4` and looks less ephemeral than the replacement.

slibby:
@ingalls the old url is for Harney County, Oregon, not California, this is now the proper location/service combination.

ingalls:
:+1:

PR #1322, opened 2015-09-29
Cass County MN

mmdolbow:
I'm wondering if this will work even though the split tag is deprecated. If not, I would love to use this example to learn how a regex would apply in this case.


ingalls:
Yes the split tag is deprecated but will still work as expected.  Looks like you have a bad json file in your PR. Do you mind removing -  `sources/us-mn-cass.json`

mmdolbow:
Whoops, killed that errant one.

PR #1325, opened 2015-09-30
Move ca-on-haldimand to proper folder

slibby:


geobrando:
That's weird. Good catch.

Would you mind dropping the `_county` from the filename and also change to a `"county":` in the coverage object?

slibby:
@geobrando done - after looking into it more, canadian sub-terroritories/provinces are kind of a mess, so using the "county" scheme like we do in the US sources probably won't work for all.

geobrando:
@slibby Yeah. It's interesting. According to wiki Halimand is called a county but is a "single-tier municipality" which are often called "megacities".

PR #1346, opened 2015-10-09
Contributing to: us-ga-atlanta.json

landonreed:
Adding/Updating us-ga-atlanta.json

### DIFF
--- a/sources/us-ga-atlanta.json
+++ b/sources/us-ga-atlanta.json
@@ -8,5 +8,14 @@
     "website": "http://gis.atlantaga.gov/gishome/index.php?option=com_content&task=view&id=10&Itemid=52",
     "attribution": "",
     "type": "http",
-    "compression": "zip"
+    "compression": "zip",
+    "conform": {
+        "type": "shapefile",
+        "number": "StNumber",
+        "street": [
+            "Name",
+            "PostType",
+            "PostMod"
+        ]
+    }
 }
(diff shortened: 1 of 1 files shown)


geobrando:
@landonreed Looks like you may have copied image URL when you meant to copy link address.

landonreed:
@geobrando ah, woops!

Fixed.

landonreed:
Also, now that I'm looking at it... I'm not sure that's a stable url (given that it has the last updated date in the path).  @geobrando, is that a dealbreaker for openaddresses?

geobrando:
@landonreed Yeah.  It's kind of an issue.  I believe that currently what happens is that if a url changes, the cached data from the last successful run will be the current data.

This is fairly common.   What some have done is to write the municipality and ask if they could provide a stable url.  In most cases, the date is stored in a metadata file so their naming convention is superfluous.

geobrando:
But we can merge this for now and update later if you want to contact the city and they agree to make that change.

landonreed:
Yea, I'll reach out to the city about the url.  Definitely go ahead and merge for now if that's acceptable.  I'm just happy to have some dots on the map for ATL in the project.  :fireworks:

geobrando:
:+1: 

FYI,  our web app is broken and not creating files in the right place, as we have a new folder structure.  I'll merge and move it.  If you find any additional sources, I recommend avoiding using it.

PR #1348, opened 2015-10-09
[WIP] license tests

### LINKED ISSUE #1347: Update tests to reflect new work on license tag
We are missing tests to reflect changes introduced in recent pull requests:

* https://github.com/openaddresses/openaddresses/pull/1339 - new license dictionary syntax.
* https://github.com/openaddresses/openaddresses/pull/1344 - new license attribution properties.

geobrando:
Giving this a shot.

@migurski Was going to include a new source containing a license object.  Should I wait until machine is ready to go first?

Was also considering adding tests to make sure `license.url` is a valid url, but you'll be testing this in hooked right?

fixes #1347

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,8 @@
   "devDependencies": {
     "glob": "4.2.x",
     "tape": "3.0.x",
-    "tap-min": "^1.0.0"
+    "tap-min": "^1.0.0",
+    "validator": "4.0.x"
   },
   "dependencies": {
     "request": "^2.51.0"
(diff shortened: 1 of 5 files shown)


migurski:
Machine is ready now! If you include a new source that does this, it should be reflected properly.

migurski:
Also, not doing anything to check URL validity right now.

geobrando:
OK.  Figured out that my tests were using the wrong key (I copied from what was there) and this was not caught because `substack/tape` does not intercept uncaught exceptions.  I'm also now properly testing for `"attribution"` to be boolean.  I fixed two other tests that were using the wrong key.

I'm using validator.js to test validity of the attribution url.  We could also use this for website url and email.  Neither are currently used in any way so they are not tested by hooked CI.

Default values for `isURL()` from validator.js

>**isURL(str [, options])** - check if the string is an URL. options is an object which defaults to { protocols: ['http','https','ftp'], require_tld: true, require_protocol: false, require_valid_protocol: true, allow_underscores: false, host_whitelist: false, host_blacklist: false, allow_trailing_dot: false, allow_protocol_relative_urls: false }.

PR #1351, opened 2015-10-15
Add Turku, FI

hannesj:


hannesj:
It seems there is a stray character at the last row, how should this be fixed?

geobrando:
@hannesj Thank you for the contribution. If that's the cause of the failure I would say there are two options:

1. Manually removing the stray character, after which OA could host the file directly
1. Contacting the data provider and asking that they rectify the issue

The latter would be preferable, especially if this data is updated frequently as well as if this stray character is introduced systematically as part of their processes.

tuukka:
@geobrando the data provider has now fixed the issue in their publishing process. Can you restart the checks?

hannesj:
This passes the tests after refreshing them by pushing an updated version.

migurski:
Great!

PR #1355, opened 2015-10-16
Contributing to: slovakia.json

MichalPP:
Adding/Updating slovakia.json

### DIFF
--- a/sources/slovakia.json
+++ b/sources/slovakia.json
@@ -1,0 +1,12 @@
+{
+    "coverage": {
+        "country": "Slovakia",
+        "state": "",
+        "city": ""
+    },
+    "data": "http://epsilon.sk/sk-countrywide.zip",
+    "website": "",
+    "attribution": "Ministerstvo vnútra Slovenske republiky / Ministry of Interior of the Slovak Republic",
+    "type": "http",
+    "compression": "zip"
+}


slibby:
@MichalPP this is great! It needs a couple edits to the JSON to pass and be merged -- would you like to try to update it? It needs a conform object, which would make the final json look like this:

```
{
    "coverage": {
        "country": "Slovakia"
    },
    "data": "http://epsilon.sk/sk-countrywide.zip",
    "attribution": "Ministerstvo vnútra Slovenske republiky / Ministry of Interior of the Slovak Republic",
    "type": "http",
    "compression": "zip",
    "conform": {
        "street": "streetname",
        "number": "streetnumber",
        "type": "csv",
        "file": "sk/countrywide.csv",
        "lat": "lat",
        "lon": "lon",
        "city": "municipalityname",
        "postcode": "postalcode"
    }
}
```

If you want to try updating the file in your own branch, that'd be great, or let me know and I can do it and close these two PRs.

slibby:
+976,735 addresses!

MichalPP:
just to clarify addresses structure, in Slovakia we have two types of addresses (when you want to write s-mail):
1: streetname streetnumber, postalcode city/bigger village
2: č.d. constriptionnumber, postalcode small village (when streetname is null)
(3: streetname constriptionnumber/streetnumber, postalcode city - sometimes)

how should I differentiate this? put some text field with actual address text?

I can also rename the field names, no problem with that.

slibby:
If there are two different sets of data, the only way to format them differently would be to create two separate files, one for each category you described above, and then contribute them separately to OpenAddresses. Currently there is not a logic to say "if it has a streetnumber, use that, if not, use constriptionnumber"

Does that make sense?

MichalPP:
I've updated the file, renaming column names and putting street names so that small villages are treated the same way as cities (so the address would be "smallvillage 25, postcode smallvillage" which is the usual way to write postal address, as I was told by people actually using postal services)

feel free to update and commit the json file, I have not yet mastered any git commit.

any parts I should add to the zip file?

slibby:
@MichalPP Thanks for the update. I have created a new Pull Request here: https://github.com/openaddresses/openaddresses/pull/1366

Thanks again for your contribution, this is great!

PR #1356, opened 2015-10-18
Update douglas.json

mmdolbow:
Think this might work better for Douglas County, which appears not to be appearing on data.openaddreses.io. I found a bunch of WI counties on http://ags1.wgxtreme.com/ArcGIS/rest/services as well.


iandees:
Looks like you're missing a comma in the JSON there.

PR #1357, opened 2015-10-19
Update douglas co Wisconsin

mmdolbow:
Here's a better attempt, I hope. Note the data URL base has more counties to peruse.


slibby:
@mmdolbow unfortunately the ArcGIS Server is version 9.3.1 which I believe is the reason it's having problems caching it.

iandees:
Yep. All of the tricks we use for downloading from ESRI services require at least version 10.

We could probably use a set of spatial queries as a fallback...

geobrando:
Alright. Closing this then.

PR #1359, opened 2015-10-20
Create us/ak/kenai_peninsula_borough.json

slibby:


geobrando:
Looks great @slibby.

Check out the new license object, which you should try to use in the future as the standalone `"attribution":` tag is deprecated.

Also, I would recommend only adding an attribution field if it is stated that attribution is required (and not for public domain data, for example).  Don't know if that is the case with this source.

PR #1363, opened 2015-10-22
Create city_of_canberra.json

slibby:


slibby:
replaces https://github.com/openaddresses/openaddresses/pull/1293

geobrando:
Hey @slibby. What's your thought about license/attribution requirements for data we've scraped from Esri servers?

slibby:
@geobrando it's a good question. Generally, my feeling is that if someone has published it to a public REST service, and they haven't filled in the attribution, credits and usage constraints, the implicit decision is that the data is not sensitive from a privacy perspective. 

I do think we may have a responsibility to notify these organizations and make sure they know their data has been added to OA, but that's both a lot of work and a great way to get a lot of data removed from OA. 

I'd be interested in @ajturner 's take on this too - what is the implicit attribution we should apply to public-facing ArcGIS REST services that do not contain license or attribution metadata.

iandees:
I think this is a conversation worthy of its own thread/ticket. Does someone want to pose this question in a new ticket please?

Issue #1306, opened 2015-09-25
Indiana Statewide


slibby:
Available as REST service here: http://maps.indiana.edu/arcgis/rest/services/Infrastructure/Streets_Address_Points_IDHS/MapServer/0
3,191,230 points reported, probably too many to scrape with machine.

file GDB available in current source json:http://maps.indiana.edu/download/Infrastructure/Streets_Address_Points_IDHS.zip

caching doesn't support filegdb (yet).

ingalls:
@slibby want to add the REST endpoint as a source anyway? There are talks about changing how the timeout works with ESRI values so it would be good to have this ready to test out.

slibby:
Will do.

Sent from my mobile phone.

On Sep 24, 2015, at 10:23 PM, Nick <notifications@github.com<mailto:notifications@github.com>> wrote:


@slibby<https://github.com/slibby> want to add the REST endpoint as a source anyway? There are talks about changing how the timeout works with ESRI values so it would be good to have this ready to test out.

slibby:
closed by https://github.com/openaddresses/openaddresses/issues/1364

PR #1364, opened 2015-10-23
Create indiana statewide_2.json

slibby:
Not sure this will work, both too long (3M rows) and duplicate, but wanted to get it in.

Addresses https://github.com/openaddresses/openaddresses/issues/1306


iandees:
I think the last time I tried this there were oodles of rows with no geometry. Looks like they might have fixed it though.

PR #1367, opened 2015-10-23
Create New Hampshire statewide.json

slibby:
Located a statewide parcel layer. 500,000+ parcels.


iandees:
578,233 features downloaded. Nice!

Issue #942, opened 2015-03-26
Rīga, Latvia


ansis:
https://opendata.riga.lv/visas-datu-kopas.html
https://opendata.riga.lv/visas-datu-kopas.html?file=files/Opendata/Dati-Vectori/Attistiba/Adreses_shp_05_03_15.zip

license:
https://opendata.riga.lv/licence.html
http://opendatacommons.org/licenses/by/1.0/

slibby:
Contributed here: https://github.com/openaddresses/openaddresses/pull/1371

PR #1371, opened 2015-10-26
Create lv/city_of_riga.json

slibby:


slibby:
Addresses: https://github.com/openaddresses/openaddresses/issues/942

PR #1377, opened 2015-10-30
Added town of usa-ct-fairfield

riordan:
usa-ct gets most of its coverage from [this source](https://github.com/openaddresses/openaddresses/blob/usa-ct-fairfield/sources/us/ct/statewide.json), managed by the state, but there are a few missing towns (and appears to not be updated since 2011).


riordan:
[Same source](http://www.ctmetro.org/data-services/maps/) also provides (which I'm happy to take on next):
- Bridgeport, CT (major city, present from main state import)
- Easton (small town, missing)
- Monroe (small town, missing)
- Stratford (mid-size town, present)
- Trumbull (mid-size town, missing)

geobrando:
Looks great.  Could you move the website and note tags outside of the coverage object?

Please.  Feel free to tackle those!

riordan:
@geobrando Happy to! Coverage issues should be resolved.

PR #1378, opened 2015-10-30
Adding the rest of CT METROCOG

riordan:


geobrando:
Thanks @riordan!

A few things:

1. It looks like you're missing a couple commas after the `split` tags in each.
1. The `split` tags have been deprecated.  We can merge this, but you should check out the CONTRIBUTING and try to use the new `function` tags for future contributions.
1. Also for future reference, the attribution information should be included as part of a `license` object (see CONTRIBUTING).  I would also recommend only including it if there is reason to believe that attribution is required by the data provider.

riordan:
Not sure what's up with this. Closing it for now until I can resolve.

iandees:
The column name used in split isn't there. If the column *is* actually there, it looks like the column name is getting lower-cased, so maybe there's a problem there?

geobrando:
@riordan Looks like you used the url for Trumbull for both of these.

riordan:
And here I was thinking I'd have to fight a regex this morning.

It's passing.

geobrando:
:+1:

PR #1379, opened 2015-11-01
us-ny-rochester

riordan:


geobrando:
@riordan Could you rename to `city_of_rochester.json`?

riordan:
👍

PR #1381, opened 2015-11-02
Corrections for Utah Addresses

riordan:
Utah Addresses have several core fields missing, rendering grid addresses in Salt Lake City unsearchable. This means we were missing out on ~ 475,000 addresses for the state.

This cleanup adds:
* Address number suffix (123 **1/2** fake street)
* Prefix Street directionals (123 **S** fake street)
* Suffix Directionals (123 S Fake Street **E**)

Bonus points:
* City
* State
* Parcel ID's

Huge shoutout to @stephenkhess for actually finding this issue, who along with @orangejulius helped write the importer with me.

### DIFF
--- a/sources/us/ut/statewide.json
+++ b/sources/us/ut/statewide.json
@@ -13,11 +13,19 @@
     "conform": {
         "type": "shapefile",
         "file": "AddressPoints/AddressPoints.shp",
-        "number": "AddNum",
+        "number": [
+            "AddNum",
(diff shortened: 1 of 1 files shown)


riordan:
Looks like it's passing now.

We're currently missing 36 addresses that are just "numeric housenumber", "numeric streetname". Some of those are in there, but as fulltext. Maybe worth writing a regex for em?

PR #1385, opened 2015-11-03
added prefix and suffix directionals

stephenkhess:
Trumbull County, OH was missing prefix and suffix directionals

### DIFF
--- a/sources/us/oh/trumbull.json
+++ b/sources/us/oh/trumbull.json
@@ -15,8 +15,10 @@
     "conform": {
         "number": "HOUSENUM",
         "street": [
+            "ST_PREFIX",
             "ST_NAME",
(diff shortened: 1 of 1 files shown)


stephenkhess:
Trumbull County, OH was missing prefix and suffix directionals

migurski:
Looks like the failure here is due to a change in machine: we started to look at `.txt` files and it turns out there are sources with `.shp.txt` extensions.

migurski:
I rolled back the previous `.txt` changes in version 2.7.4 by the way; they were causing problems like this but only fixing minor ones.

stephenkhess:
Excellent, thanks! There are a lot more fixes coming in tomorrow for ohio.

PR #1388, opened 2015-11-04
Fix Washtenaw County, MI, USA

orangejulius:
The data URL pointed to an ESRI endpoint that only contained a list of
buildings on a floodplain. This new URL contains all parcels in the
county.

Note: for some reason I wasn't able to test this URL in QGis but it
looks to contain the right data. I'm still a QGis/ESRI noob so I'm
simply going to rely on the automated tests in Github to see if I got it
right. In other words, other commits may be coming.


orangejulius:
After a URL fix and the addition of a regex to remove the housenumber from the street field, the data from the build looks good now!

PR #1392, opened 2015-11-04
Add street directionals to street name

dianashk:
Allegheny county data has street directional in the ST_PREFIX column that should be added to the street name.


iandees:
Niiice.

PR #1393, opened 2015-11-04
added pre- and post-directionals for .zip-based Ohio counties

stephenkhess:
A large number of Ohio counties were missing pre- and post-directional fields available in the source data.  This PR fixes that.


migurski:
:+1: :+1: :+1:

PR #1399, opened 2015-11-05
spelling correction from "County" -> "Country"

stephenkhess:


stephenkhess:
Looks like Atom removed a space from the end of a line, too.

PR #1404, opened 2015-11-06
Update 3 California county source files

alexfriant:
Mostly added cities, states, zipcodes, and APN's. Los Angeles' data doesn't include a value for State, unfortunately.


iandees:
:100:

PR #1412, opened 2015-11-10
Update Kent County, MI

### LINKED ISSUE #1389: Add parcel data for Kent County (home of Grand Rapids, MI)
Kent County has shapefiles of parcel data for every city in the county, including Grand Rapids.

They're available at https://www.accesskent.com/GISLibrary/#Parcels&Streets

I'll probably try to get to it some time in the next few days, unless someone else really wants to do it :)

orangejulius:
As shown in the [build output](http://data.openaddresses.io.s3.amazonaws.com/runs/33309/output.txt),
it doesn't look like the current URL works: it was run through some
proxy which hopefully isn't needed.

This also changes the URL from an ESRI layer that has only address info,
back to one that has full parcel data, as well as super valuable
zipcodes.

This URL was added in PR #359 but then changed in #1244, so if there was
a reason we can change it back. But it looks like this will give full
polygon and address data, where the other won't.

Fixes #1389

### DIFF
--- a/sources/us/mi/kent.json
+++ b/sources/us/mi/kent.json
@@ -9,20 +9,28 @@
         "state": "mi",
         "county": "Kent"
     },
-    "data": "http://gis.kentcountymi.gov/public/kcsalesviewerweb/proxy.ashx?http://gis.kentcountymi.gov/prodarcgis/rest/services/External/MapServer/6",
+    "data": "http://gis.kentcountymi.gov/prodarcgis/rest/services/External/MapServer/5",
     "type": "ESRI",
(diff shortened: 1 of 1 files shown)


orangejulius:
Hey @iandees, any idea why this is failing? I think the URL is valid, as I can pull from it and even query its size locally in Qgis...

iandees:
It's failing because the server doesn't support one of the queries we're making (even though it says it does). The code that does these queries needs to be fixed to fall back to another method if this one doesn't work.

I'm going to merge this because the source itself is fine and useful. Thanks!

PR #1439, opened 2015-11-11
switched to regexp for more accurate address parsing

stephenkhess:
also added lat, lon, city, city, and postcode


iandees:
The lat and lon fields only need to be set on CSV sources. We automatically pull the lat/lon from the geodata on shapefiles.

Issue #1458, opened 2015-11-17
Update Queensland info


ljbade:
The Queensland data is now available from https://data.qld.gov.au/dataset/property-address-queensland--data-package

Still requires an email flow, but is listed again as CC-BY

ljbade:
Working on a PR. Will upload a new cache of October 2015 data.

A few corrections:
- updated licence field to new format
- ensure the street number post-fix is joined
- add city/district fields

A number of addresses I am not sure how we handle:
- place name instead of number + street, e.g. name of a farm (uses separate column)
- unit numbers, e.g. apartments, 12/33 blah street (unit in separate column from street number)
- street address ranges - block might have a range of street numbers, e.g. 14-15 blah street
- unique id - possibly from plan and lot number columns
- quality/source ids - there are columns for determining how the address was created
- "XXXX" addresses, can find this in the street prefix at times

ljbade:
I also emailed the department that looks after this data to see if they have a better description then the one included in the metadata page. It is missing a number of codes etc. Even if we don't use them we should at least attach to this issue for future reference.

ljbade:
@sbma44 When using COLUMNx for the field name, does it start count at 0 or 1?

ljbade:
Hmm, not sure how to upload a new cache file.

sbma44:
@ljbade I can give you a hand uploading the new cache file

porjo:
http://results.openaddresses.io/ is showing the following error in the log output (instead of 'Show Sample Data' button)
```
2015-12-04 23:15:41,639  WARNING: Error doing excerpt; skipping
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/openaddr/__init__.py", line 167, in conform
    data_sample, geometry_type = task3.excerpt(decompressed_paths, workdir, conform)
  File "/usr/local/lib/python2.7/dist-packages/openaddr/conform.py", line 220, in excerpt
    layer = datasource.GetLayer()
AttributeError: 'NoneType' object has no attribute 'GetLayer'
```

sbma44:
@porjo Australia has announced that it'll be releasing a nationwide dataset in February that will supersede this one; I think we should consider `au-*` deprecated until that arrives.

Issue #1457, opened 2015-11-17
Update Tasmania info


ljbade:
The information on Tasmania should be updated to reflect that it is now possible to easily download all the address points for Tasmania from http://listdata.thelist.tas.gov.au/opendata/index.html#LIST_Address_Points

It is broken up by council boundaries into separate files.

/cc @sbma44

ljbade:
Working on a PR. Not sure how to get this to update data now it is split by council area.

A few corrections:
- updated licence field to new format
- ensure the street number post-fix is joined
- add city/district/postcode/state fields
- ensure the street suffix is joined, eg "SOMEWHERE" "ST" "EAST" (only a few roads have these)
- id

A number of addresses I am not sure how we handle:
- place name instead of number + street, e.g. name of a farm (uses separate column)
- unit numbers, e.g. apartments, 12/33 blah street (unit in separate column from street number)
- street address ranges - block might have a range of street numbers, e.g. 14-15 blah street (end street number in seperate column)
- quality/source ids - there are columns for determining how the address was created

Full details on each field is in http://listdata.thelist.tas.gov.au/public/LIST_Address_Points_information.pdf

ljbade:
Exactly how one can use the new data with our system is not clear as it is split into multiple files by council area.

A cache file will still likely be needed which combines the shape files into a single file.

iandees:
The easiest thing to start with is probably to give each council file its own source in OpenAddresses.

PR #1459, opened 2015-11-17
Update Australia data sources.

ljbade:
For https://github.com/openaddresses/openaddresses/issues/1458 and https://github.com/openaddresses/openaddresses/issues/1457

Still work-in-progress.

Might also take a look at Victoria and NZ.


ljbade:
The columns for the Victorian data are in http://www.depi.vic.gov.au/__data/assets/pdf_file/0003/261651/Vicmap-Address-Data-Model-V5.pdf

Mostly follows the other states, but I should check the data for how street number prefixes are handled.

I have applied similar updates to the sources file as I have done for QLD and TAS

sbma44:
G-NAF is going open in Feb 2015 -- let's plan to load that instead of refreshing. If someone else wants to tackle this prior to then they should be all means reopen!

https://blog.data.gov.au/news-media/blog/geocoded-national-address-data-be-made-openly-available

slibby:
Closing this issue as we now have the full G-NAF in https://github.com/openaddresses/openaddresses/pull/1588

PR #1462, opened 2015-11-26
switch to regexp for more accurate number/street parsing

stephenkhess:


migurski:
Looks good! For more chances like this, could we keep the chance to the JSON fairly localized to just the affected lines?

PR #1463, opened 2015-11-26
mass switch from old style auto_number/auto_street to regexp for more…

stephenkhess:
… accurate number/address extraction


iandees:
It looks like you changed the indent on all these files. Can you switch them back to 4 spaces please?

stephenkhess:
Yep, sorry about that, I wrote a script to handle them so that's easily fixable.

stephenkhess:
GAH!  Fixed the indent but git does not appear to notice the difference.

iandees:
Also, it looks like you're not writing out a newline as the last character. Can you add an extra `\n` to the end?

stephenkhess:
Screw it, I'm starting over, the indentation checkin will trigger all the bad sources to run tests again.

PR #1464, opened 2015-11-26
Add statewide JSON for Puerto Rico

waldoj:
This is using Puerto Rico's statewide address points file, _but it only includes a small subset of the commonwealth's addresses._ There are just 135,000 in all, including San Juan and neighboring Bayamón and Carolina. I have no idea if why PR's statewide address points file only includes these addresses.

### DIFF
--- a/sources/us/pr/statewide.json
+++ b/sources/us/pr/statewide.json
@@ -1,0 +1,32 @@
+
+{
+    "data": "https://data.pr.gov/api/views/h9md-ic24/rows.csv?accessType=DOWNLOAD",
+    "website": "https://data.pr.gov/en/Desarrollo-e-Infraestructura/Direcciones-Estandarizadas-Bay-Car-SJ/h9md-ic24",
+    "license": {
+        "text": "public domain",
+        "attribution": false,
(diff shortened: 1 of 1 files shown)


slibby:
@IanHenshaw looks like the contributor/creator of the data -- maybe he has some insight.

Ian, are you aware of better territory-wide data on address points? Or is the file that Waldo added the best example out there?

slibby:
Found an existing effort to import to OSM too: http://wiki.openstreetmap.org/wiki/Puerto_Rico_Projects/Municipal_Buildings_Addresses_Import

slibby:
I was reviewing this and thinking about how to define the source - it says it's all of PR on the data site, but as you mentioned before, it's really only two specific areas. "statewide" will over-emphasize the coverage on the results page.

Also found this, worthy of keeping an eye on in case a parcel or address point map appears: http://www.satasgis.crimpr.net/crimgis/rest/services

waldoj:
Whether to include a "statewide" resource that is not actually statewide—and what the threshold is for which to include it—is a philosophical matter that is beyond my ken. :)

Issue #1466, opened 2015-12-02
Add unmaintained statewide Wisconsin data?


waldoj:
[The state of Wisconsin published a statewide dataset in 2013](http://www.sco.wisc.edu/find-data/addresses.html):

> A statewide address point dataset from the LinkWISCONSIN EBM Address Point and Parcel Mapping Project is available from the Public Service Commission of Wisconsin’s State Broadband Office. This dataset was aggregated from data received during the fall of 2013 and is not maintained or updated in any form.

([Direct link to the data, a 674 MB ZIP file](http://psc.wi.gov/utilityInfo/tele/broadband/documents/LinkWISCONSIN_EBM_ParcelAddress_Data.zip).) Given its age and lack of updates, should this be included in OpenAddresses?

iandees:
Interesting that they released this… I asked the maintainer if we could have it and they said no as recently as last year. It's definitely worthy of getting added to OpenAddresses, I think.

waldoj:
Great, I'll file a pull request. Thanks!

PR #1468, opened 2015-12-03
Add a statewide Wisconsin dataset

waldoj:
The catch is that it's an address point file geodatabase, which I don't think we can process. This was raised in #1466.


waldoj:
Too many things combined here. Trying again.

Issue #1400, opened 2015-11-05
The Netherlands - suggestion for resource


justb4:
I see that the addresses CSV for The Netherlands is lacking several fields like Postal codes, City, Province, WGS84 coordinates,  etc. I scanned several issues on this subject: #93, #233 and #800. 

From what I noticed is the NodeJS script https://github.com/openaddresses/openaddresses/blob/master/scripts/nl/nl-data.js that parses the raw address-data provided by the Dutch government (Kadaster, PDOK) at http://geodata.nationaalgeoregister.nl/inspireadressen/extract/inspireadressen.zip will not render complete addresses. Worse: as also history is included in the source data (using begin-end dates and other attributes) , it may render many duplicates. 

This is not to blame anyone, since the Dutch government made it *very* hard for us to derive  complete addresses from their source data :-( called the "BAG" (Base Registry for Addresses and Buildings). Apart from the history-issue, the dataset "BAG" comprises both buildings and "addressable objects" in a complex, normalized XML schema but lacking "Gemeentes" (like Counties) and Provinces and is using a local projection (EPSG:28992).

For some years, as Bert Spaan @bertspaan has indicated in one of the issues, we run a local Open Source project called NLExtract (http://nlextract.nl) to provide tools to extract the BAG and other Dutch geo-datasets mainly into PostGres/GIS. For the BAG this is a dedicated Python program with several SQL scripts to enrich the data with Counties, Provinces and more.

As running the NLExtract scripts for addresses may be a bit complex and time-consuming (couple of hours), we are also making the end-results available as downloads. The data is refreshed each month by PDOK via an RSS URL: http://geodata.nationaalgeoregister.nl/inspireadressen/atom/inspireadressen.xml
We have a daemon running that detects that a new BAG-version is available. It will run NLExtract for BAG and make the results available as a PostGIS dump or, what I think is useful for OA a zipped CSV of adresses. The download URL is  http://data.nlextract.nl/bag/csv/
(I now see that the date is in the CSVs, we may change this to "latest").

So all in all I would suggest to harvest Dutch addresses from http://data.nlextract.nl/bag/csv. For this I guess some changes are needed. Let me know if I can help, possibly via a PR.

sbma44:
This sounds like an excellent suggestion, @justb4 -- thank you for taking the time to dig into the problem. Replacing the existing NL source sounds like a good idea.

[...]

PR #1471, opened 2015-12-07
Update nl-countrywide.json

slibby:
Closes https://github.com/openaddresses/openaddresses/issues/1400


slibby:
I think this just took too long, not sure why it failed.

PR #1479, opened 2015-12-23
Update bologna.json

slibby:


slibby:
thanks @migurski -. .Shp looks good now.

Issue #1380, opened 2015-11-02
Add "unit" to accepted conform tags


migurski:
@iandees: “This will be things like apartment number, unit number, etc.”

Context: https://github.com/openaddresses/machine/issues/232

migurski:
With https://github.com/openaddresses/machine/issues/232 closed and machine 2.9.0 released, we can add this to documentation.

slibby:
How's this look: https://github.com/openaddresses/openaddresses/pull/1481

migurski:
Oops, "unit" needs to be present in the unit tests as well. I always forget this step.

PR #1481, opened 2015-12-23
Update CONTRIBUTING.md to include `unit` in conform

slibby:
Attribute | Required? | Description
---------- | --- | ----
`unit`     |     | The Unit number. This is an optional attribute and can be used for uniquely identifying addresses where the street number is the same but a "unit" number is specified, i.e. `300 Willow Valley Lakes Drive #3A` where `3A` can be found as a separate attribute in the source.


iandees:
:+1:

slibby:
closes https://github.com/openaddresses/openaddresses/issues/1380

PR #1485, opened 2015-12-23
Add a statewide Wisconsin file

waldoj:
The catch is that it's an address point file geodatabase, which I don't think we can process. This was raised in #1466.


migurski:
Looks like there are some JSON quotes missing in the coverage object.

waldoj:
This JSON file is 80% missing quotation marks, by weight.

waldoj:
Soo, I think this is going to just keep failing, because it's an address point file geodatabase, which we can't do anything with. It's somebody else's call how to handle it (per #1466).

iandees:
I extracted the address point layer from the GDB you point to here to a CSV and uploaded it to `http://openaddresses.s3.amazonaws.com/us-wi-statewide.csv.zip`. The columns are:

- X
- Y
- RelateID
- ParcelID
- CountyID
- FullMailAdd
- FullAdd
- AddNumPrefix
- AddNum
- AddNumSuffix
- Prefix
- StreetName
- StreetAlias1
- StreetAlias2
- StreetAlias3
- StreetType
- Suffix
- LandmarkName
- UnitType
- UnitID
- PlaceName
- PlaceNameAlternate
- CensusPlaceName
- ZipCode
- Zip4
- State
- PtType
- AddSource
- LoadDate
- EntryDate
- CAI
- Township
- Section
- Range
- GEOID
- SourceType
- AddLifeStaus
- AddressCompleteness
- AddressLongitude
- AddressLatitude

iandees:
This has 2,741,467 points, by the way.

waldoj:
Good Lord.

Issue #1483, opened 2015-12-23
Fix Montana County Parcel downloads


slibby:
It looks like Montana changed their FTP server, the new folders are here ftp://ftp.geoinfo.msl.mt.gov/Data/Spatial/MSDI/Cadastral/Parcels/

I think every MT county will need an update.

PR #1486, opened 2015-12-23
Update MT Sources to new FTP Site

slibby:
MT changed their FTP server address, so I went to their website and checked -- got a new DNS alias and no longer requires a password, so this should work to update all of them.


slibby:
Closes https://github.com/openaddresses/openaddresses/issues/1483

PR #1488, opened 2015-12-23
Added Lancaster, PA unit number

migurski:
…thanks https://github.com/openaddresses/openaddresses/issues/1380


migurski:
Units are in:

units

PR #1495, opened 2015-12-28
[Don’t Merge] Moved Canberra to ESRI hosting

migurski:
* [ ] Wait for https://github.com/openaddresses/machine/issues/272

### DIFF
--- a/sources/au/city_of_canberra.json
+++ b/sources/au/city_of_canberra.json
@@ -4,8 +4,8 @@
         "state": "ACT",
         "city": "Canberra"
     },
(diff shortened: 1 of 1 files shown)


migurski:
Hm, “source has 1 rows” doesn’t sound right.

iandees:
It looks like there this is a single bounding box on this layer and not actual points:

http://www.arcgis.com/home/webmap/viewer.html?url=https://services1.arcgis.com/E5n4f1VY84i0xSjy/ArcGIS/rest/services/Geocoded_Addressing/FeatureServer/0&source=sd

migurski:
Boo to them.

migurski:
Don’t think this new source will work, since it’s not got address points.

Issue #249, opened 2014-07-03
Albemarle County, VA, USA


waldoj:
On [their GIS data page](http://www.albemarle.org/department.asp?department=gds&relpage=3914), Albemarle provides parcel-level data ("Current Parcels") and building polygons ("Buildings"). Neither of these GIS sets have addresses within the data, but the ZIP file for building polygons includes a dBASE table that contains the following fields:

* GPIN
* PIN
* HOUSENO
* UNIT
* STREET
* CITY
* ZIP
* PERMIT
* CREATED
* CREATED_BY
* CAMA_CARD
* ADDR_ID

The GPIN is also found within the buildings geodata. Seems to me that, with a bit of post-processing, these two files could be combined to useful effect.

waldoj:
Yay! (I live in Albemarle.) Thanks!

PR #1496, opened 2015-12-29
Create us/va/albemarle.json

slibby:


slibby:
closes https://github.com/openaddresses/openaddresses/issues/292 and https://github.com/openaddresses/openaddresses/issues/249

Issue #1473, opened 2015-12-10
Isn't it a problem that HERE North America LLC holds a copyright over the zip code data in the New York statewide addresses


dthgeek:
The [source](http://gis.ny.gov/gisdata/inventories/details.cfm?DSID=921) for us/ny/statewide states that "In the Address Points–SAM file, the ZipCode and ZipName attribute data is copyrighted by HERE North America LLC 2013 and licensed to the New York State Office of Information Technology Services. Please refer to the Metadata for more information." and the metadata for  Address Points - SAM - Statewide (http://gis.ny.gov/gisdata/metadata/SAM.address_points_SAM.xml) says "Use_Constraints: ZipCode/ZipName data copyrighted by HERE North America LLC © 2013 and licensed to the New York State Office of Information Technology Services". So doesn't that mean the address data can't be used by openaddresses.io?

slibby:
On my reading that means that the ZipCode attribute should be removed from the scraping effort, but the address points themselves are not restricted.

dthgeek:
Thank you for responding and sorry about the delay. 
Was the ZipeCode attribute from HERE removed and replaced using TIGER data? Also why is the rest of the data from New York Sate it unrestricted?

iandees:
We removed the postcode from the output. Thanks for pointing this out.

PR #1497, opened 2015-12-31
US/NY review

riordan:
Updates to records for NY State.

Primary changes are addition of UNIT, Postalcode, cityname, and complete housenum for statewide dataset, with some minor improvements (addition of parcel ID's, descriptions where possible) to others.


sources/us/ny/statewide.json
```diff
@@ -14,6 +14,14 @@
...
         "street": "CompleteStreetName",
-        "number": "AddressNumber"
+        "number": ["PrefixAddressNumber", "AddressNumber", "SuffixAddressNumber"],
+        "unit": "SubAddress",
+        "city": "ZipName",
+        "region": "State",
+        "postcode": "ZipCode",
+        "district": "CountyName",
+        "id": "NYSAddressPointID"
+
```
iandees: Ditch these blank lines.

sources/us/ny/statewide.json
```diff
@@ -14,6 +14,14 @@
     "conform": {
         "type": "geojson",
         "street": "CompleteStreetName",
-        "number": "AddressNumber"
+        "number": ["PrefixAddressNumber", "AddressNumber", "SuffixAddressNumber"],
+        "unit": "SubAddress",
+        "city": "ZipName",
+        "region": "State",
+        "postcode": "ZipCode",
```
iandees: We can't use the `ZipCode` field because it is apparently filled using proprietary Navteq data. See https://github.com/openaddresses/openaddresses/issues/1473.

riordan: Bah, humbug. Good catch.

Same rights restrictions [apparently also apply to ZipName](http://gis.ny.gov/gisdata/inventories/details.cfm?DSID=921).

Switching that over as well.

slibby:
previous run on hooked-ci ran out of time, interested to see if the changes to ArcGIS Server queries from last week help here: http://data.openaddresses.io.s3.amazonaws.com/runs/49643/output.txt

riordan:
@slibby Yep; you nailed it. Timed out again.

slibby:
This could be random, but in comparing the two runs, it seems to be about half as fast when querying specific fields, as compared to querying all fields. 

http://data.openaddresses.io.s3.amazonaws.com/runs/49643/output.txt -> reaches OBJECTID 1500000 
http://s3.amazonaws.com/data.openaddresses.io/runs/52079/output.txt -> reaches OBJECTID ~700,000

riordan:
Grr, tried updating the source and it failed again. Might be worthwhile to keep this one Statewide, and get the others in as a separate PR.

On the other hand, what's the best way to get the statewide one in?

iandees:
I think we should probably bump the timeouts up by a few hours so the statewide layers finish more reliably.

riordan:
@iandees Is there a way we can add a note to the Statewide file about deliberately not adding the fields that aren't licensed for public use so others will know to stay away?

iandees:
Probably a `note` in the root of the source object.

slibby:
@riordan @iandees any reason this can't be merged in? looks like the last run succeeded but it was some time ago so I wasn't sure if there was a reason to wait.

riordan:
I Could push a trivial change and run it again for verification.

iandees:
I think it's good to go. I was probably waiting for a build to finish and got distracted.

PR #1505, opened 2016-01-04
added unit types/ids for Alaska

trescube:


stephenkhess:
Server seems to get hung up on 15th page of results

slibby:
Wow, 535 addresses in Yakutat Borough must be the smallest county-sized area by address count so far.

PR #1509, opened 2016-01-04
added unit types/ids and address num suffixes

trescube:


slibby:
City of Fayetteville is outputting 0 records to CSV, we'll need to look into why before merging.

slibby:
Looks like it needs to be changed to `"type": "shapefile"` in the conform and `lat`/`lon` removed.

trescube:
Updated per your suggestion, @slibby.  Any ideas on what's wrong?

slibby:
`SUFDIR` -> `SUF_DIR` is what's causing the error.

also would do add the PRE_TYPE attribute for a few `HWY 110` Addresses
```
"street": [
            "PREDIR",
            "NAME",
            "TYPE",
            "SUFDIR"
        ]
```
to
```
"street": [
            "PREDIR",
            "PRE_TYPE",
            "NAME",
            "TYPE",
            "SUFDIR"
        ]
```

trescube:
GRR!!!! they changed fields on us!

PR #1510, opened 2016-01-04
added unit types/ids

trescube:


slibby:
Thanks!

For Glendale, can you add a "region": "STATE" to get the State 2-letter code populated? Otherwise it looks great

slibby:
:+1: Thanks for adding that attribute, looks great.

PR #1514, opened 2016-01-07
Update Waupaca source

iandees:


slibby:
@iandees I think HSN_W_PRE works better for house number than concatenating HSNP, HSN, and HSNS - it includes previx and suffix and shows them with no spaces, as the addresses appear to expect: I.e. `N5335` instead of `N 5335` when we concatenate the three fields. It's a fair question as to whether we should support no-space concatenation when `number` or `street` are comprised of multiple fields.

iandees:
Ah, good catch. I was wondering about that as I was poking through the data.

PR #1516, opened 2016-01-10
fix table layout for "accuracy" line

tyrasd:
this had a superfluous column, which took the actual description to be off-screen when rendered on github.


iandees:
Thanks!

PR #1384, opened 2015-11-03
Contributing to: ua-ua-12-dnipropetrovsk.json

Andygol:
Adding/Updating ua-ua-12-dnipropetrovsk.json

PR #1520, opened 2016-01-11
Contributing to: .json

tyrasd:
Adding/Updating .json

### DIFF
--- a/sources/.json
+++ b/sources/.json
@@ -1,0 +1,12 @@
+{
+    "coverage": {
+        "country": "",
+        "state": "",
+        "city": ""
+    },
+    "data": "",
+    "website": "",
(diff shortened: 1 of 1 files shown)


migurski:
Hi @tyrasd, could this have been an error on our part? Please let us know if something is not working!

tyrasd:
Well, I was looking at the project website and found the [contribute](http://openaddresses.io/contribute/) form which apparently doesn't check if the fields are actually filled out when clicking *save* (I was just playing around, not trying to actually submit anything). I think this can be added to what I've read in https://github.com/openaddresses/openaddresses/pull/1384#issuecomment-153435208: *“The page used to contribute data is in sore need of an update”*.

PR #1528, opened 2016-01-17
added units to applicable `us/co` sources

trescube:


sources/us/co/city_of_fort_collins.json
```diff
@@ -24,6 +24,10 @@
     "conform": {
         "number": "number",
         "street": "strname",
+        "unit": [
+          "BUILDING",
+          "UNIT"
+        ]
```
iandees: missing a comma here

PR #1529, opened 2016-01-17
fixed a number of us/fl sources to include:

trescube:
- unit types/ids
- correct street fields
- city/state/postcodes, where available:


slibby:
all passing CI now - great! Not sure why the previous hooked run failed but there's some work going on to update the ci process as well which may have helped in the interim.

migurski:
:clap: :clap: :clap:

PR #1542, opened 2016-01-22
fixing Berlin, Germany

ubahnverleih:


ubahnverleih:
hm the test failed. I guess the problem is, that i tried to fix the coordinates with regex (replacing ',' with '.')

iandees:
Yep, the `lat` and `lon` attributes in the `conform` object don't support the regex stuff that other fields support.

iandees:
But there's no great reason they couldn't... any thoughts @migurski?

NelsonMinar:
That regex looks complicated. How about just having the conform parser accept comma in lat/lon columns instead of point? I'm surprised this hasn't come up before!

migurski:
Huh, yeah. It’d be nice for the regexep to work, and it would also be nice for the commas to work.

ubahnverleih:
I'm not sure why this test failed. Output looks OK for me. Any idea?

iandees:
Looks like the excerpting process failed, @migurski. I'll merge this in anyway.

Issue #1515, opened 2016-01-09
Support Constants for County/State/Region/Etc.


slibby:
It would be good if we could support adding a value in the `conform` that lets you specify a static City, State, Region, etc, when one is not available in the source. Especially for City data, they generally include number, name, even postcode, but rarely include City Name or State Name or County. 

What if in a `conform`, if any value is "_abcdef_" or "_abcdef_" or "CONSTANT:abcdef" -- then that value gets applied to every row? I think this would improve the completeness of many datasets, though it also involves a lot of re-work on existing sources. Would apply to many of the county datasets too.

migurski:
Could be an [attribute function](https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#attribute-functions), perhaps? 

```
"state": {"function": "constant", "value": "TX"}
```

PR #1543, opened 2016-01-24
added conform to city of cologne

ubahnverleih:


ubahnverleih:
is there a chance to hardcode a city for every address in a dataset? There is no city column in the dataset but all addresses are from cologne.

iandees:
There's discussion about support for constants over here: https://github.com/openaddresses/openaddresses/issues/1515.

slibby:
Nice, that's nearly 1M new addresses! :+1:

PR #1547, opened 2016-02-02
Update to ramsey MN json

mmdolbow:
New official source from the county, with new, more comprehensive field mappings.


iandees:
Thanks!

PR #1553, opened 2016-02-12
Add U.S. conforms

migurski:


migurski:
Orleans Parish source is sad because the Esri server is sending 504 responses. Merge anyway?

iandees:
Yea, I'll just merge it. I poked around with the source just now and got responses for the same query that 504'd, so maybe the server was busy when you were doing this?

PR #1558, opened 2016-02-16
Update washington.json

mmdolbow:
While the "USPS city" might not always be the actual situs city, in this service it is a better choice, because the "City" field says things like "City of Woodbury". Which is how nobody refers to it. 106K records are not null in that field.

### DIFF
--- a/sources/us/mn/washington.json
+++ b/sources/us/mn/washington.json
@@ -21,7 +21,7 @@
             "STREETTYPE",
             "SUFFIX_DIR"
         ],
-        "city": "CITY",
+        "city": "CITY_USPS",
         "postcode": "ZIP"
     }
 }


mmdolbow:
Not sure what's going wrong with this check. I get a successful query with the link below. 
I notice that if I put in the resultRecordCount and resultOffset parameters, like shown in the CI log at http://s3.amazonaws.com/data.openaddresses.io/runs/62755/output.txt, I get a failed query as well.

Successful query:
http://maps.co.washington.mn.us/arcgis/rest/services/Public/Public_Parcels/MapServer/0/query?where=1%3D1&text=&objectIds=&time=&geometry=&geometryType=esriGeometryEnvelope&inSR=&spatialRel=esriSpatialRelIntersects&relationParam=&outFields=BLDG_NUM%2CCITY_USPS%2CPREFIXTYPE%2CPREFIX_DIR%2CSTREETNAME%2CSTREETTYPE%2CSUFFIX_DIR%2CZIP&returnGeometry=true&returnTrueCurves=false&maxAllowableOffset=&geometryPrecision=7&outSR=4326&returnIdsOnly=false&returnCountOnly=false&orderByFields=&groupByFieldsForStatistics=&outStatistics=&returnZ=false&returnM=false&gdbVersion=&returnDistinctValues=false&resultOffset=&resultRecordCount=&f=pjson

iandees:
I think we've previously discovered that there's a bug in certain versions of ESRI's product (when they use a certain database backend) that breaks this. Let me see if I can dig up that issue...

iandees:
On Slack here: https://openaddresses.slack.com/archives/general/p1451946588000416

Looks like we'll need to make some changes on our side for this to work.

mmdolbow:
Bummer. Let me know if there's something I can do to help; the GIS lead for the county is a friend.

PR #1562, opened 2016-02-17
Update ramsey.json

mmdolbow:
PostalCity, while not a "situs address city" but more like a "mailing address city", seems a better choice for this service, because the CityCode makes use of a coded value domain. In other words, a human-readable city doesn't come across, instead a numeric code. This change resolves that, although the "city" listed will often be for bulk-mailing purposes, not the actual city.


slibby:
@mmdolbow while you've got the PR open, can you add `"district": "CO_NAME"` to the conform? I think it's the same for every point, but until we have a better way of designating a "static value" for district or region this helps add more to it.

mmdolbow:
Yeah, I can try. :-)

mmdolbow:
Hopefully that did as intended. I'm kind of caveman-like when it comes to the GitHubs. ;)

slibby:
:+1:

PR #1565, opened 2016-02-18
Updated the source file for Nevada County

alexfriant:
This is my first time modifying a source file for an ArcGIS Server rest endpoint. I looked at other examples to make sure I did all the tagging correctly. Let me know if you have any concerns :100:


alexfriant:
Oops, I see my mistake. I didn't need to put the value for accuracy in quotes. I'll change it and resubmit.

PR #1566, opened 2016-02-18
Updated source file for Nevada County

alexfriant:
They now have an ArcGIS Server rest endpoint to parcels with situs addresses.


iandees:
Thanks for this update!

PR #1572, opened 2016-02-23
added unit type/id fields to:

trescube:
marion_county
statewide

added more street/number fields to:
hamilton
madison

switched st_josephn to regexp function


iandees:
I'm going to merge this, but the statewide layer is a FGDB that we don't support (also, if I remember correctly the FGDB has street name in a separate table from the address data).

PR #1584, opened 2016-03-02
Add New Jersey statewide parcel-based layer

iandees:


iandees:
I'm going to merge this, but we still need to up the timeout so these long-running state-wide data sources will complete.

migurski:
Agreed, it’s on the list: https://github.com/openaddresses/machine/issues/298

Issue #1470, opened 2015-12-07
Australia's G-NAF (national) file


wboykinm:
Per [this excellent news](https://blog.data.gov.au/news-media/blog/geocoded-national-address-data-be-made-openly-available), every address geocode in Australia will become available in pipe-delimited format come February 2016. In the meantime it might be worthwhile to spec the data using samples [available from resellers](http://web.pb.com/datasample-gnafaus/?utm_source=pbs&utm_medium=mapinfodotcom-data-catalog&utm_campaign=mapinfo-data-catalog&lead_source_details=Online%20Marketing) or [this data model document](https://www.psma.com.au/sites/default/files/g-naf_product_description.pdf) from the primary contractor.

@ingalls Does this supersede https://github.com/openaddresses/openaddresses/issues/132? 

Another thing to consider is the wording of the license (vague):

> The G-NAF and Administrative Boundaries datasets will be published under an open data licence at no cost to end users on data.gov.au in February 2016.

Anyone know anybody in the Commonwealth government they could ask about specific licensing?

CloCkWeRX:
OpenStreetMap asked and received explicit permission - see wiki.openstreetmap.org/wiki/Attribution/data.gov.au_explicit_permission

Its likely to be licenced under CC-BY-3.0-AU or better, given AusGOAL (see https://wiki.creativecommons.org/wiki/Government_use_of_Creative_Commons#Australia  )

CloCkWeRX:
> Does this supersede #132?
I think that multiple address sources simply add to the value. The specific sources listed in that ticket are likely more available at a different refresh rate; this is probably going to be updated quarterly. 

AFAIK There was discussion for "GNAF Live" incremental updates somewhere; but that's a lot of effort for small changes / probably wouldn't be opened up just yet.

GNAF is basically assembled from a giant ETL; and as houses get knocked down/new addresses made it can take some time to catch up - its fed from smaller government sources at the state/local level.

wboykinm:
:+1:

sbma44:
Checked in on this -- as of 2/22, they're still working on the release: https://blog.data.gov.au/news-media/blog/update-open-g-naf-and-administrative-boundaries-datasets

maxious:
Checked in again today and it is now available! https://data.gov.au/dataset/geocoded-national-address-file-g-naf

iandees:
This schema is pretty darn complicated. Does anyone have time to look through the docs and join the tables together into something we can more easily use?

[...]

PR #1588, opened 2016-03-05
G-NAF Australia data

sbma44:
cache creation script & source JSON for http://data.gov.au/dataset/geocoded-national-address-file-g-naf

cf https://github.com/openaddresses/openaddresses/issues/1470

### DIFF
--- a/scripts/au/gnaf.sh
+++ b/scripts/au/gnaf.sh
@@ -1,0 +1,72 @@
+#!/bin/bash
+
+set -eu
+
+TMP=${TMP:-'/tmp/gnaf'}
+PGUSER='postgres'
+DBNAME='gnaf'
+PSQL="psql -t -q -U $PGUSER $DBNAME"
+
+# set up postgres, directories
+mkdir -p $TMP/gnaf
+mkdir -p $TMP/gnaf-admin
+dropdb -U $PGUSER --if-exists $DBNAME
+createdb -U $PGUSER $DBNAME
+echo "CREATE EXTENSION postgis;" | $PSQL
+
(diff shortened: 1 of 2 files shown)


migurski:
Awesome.

This might time out, but I’m in the process of making it possible for the machine to run for significantly longer in https://github.com/openaddresses/machine/issues/298.

porjo:
Will this addition appear on http://results.openaddresses.io/ as a matter of course, or does that need to be done manually by someone?

migurski:
It will appear as a matter of course.

iandees:
The dots will appear automatically on the dotmap on https://openaddresses.io, but shouldn't the coverage object in the source file here update the static PNG map at results.openaddresses.io?

migurski:
Yes, once the batch set that started after the merging of GNAF is complete. That’s in progress here: http://results.openaddresses.io/sets/67593/

migurski:
(they generally take the duration of each weekend, wrapping up sometime on Monday)

mishawakerman:
// Excuse my ignorance, I'm not familiar with openaddresses and am just looking at it for an (open-source) project.

As per https://data.gov.au/dataset/ds-dga-19432f89-dc3a-4ef3-b943-5326ef1dbecc/details:

> Updated versions of the G-NAF will be published on a quarterly basis from February 2016.

Does/will openaddresses continually update this dataset or is this something that requires manual work each quarter?

andrewharvey:
@mishawakerman As you saw I do try and update this each GNAF release, but of course there is no SLA. It's a pain doing this manually but worth the effort.

PR #1592, opened 2016-03-09
Add units to us mn

trescube:
Also other fields where appropriate, fixed a few GIS URLs

### DIFF
--- a/sources/us/mn/wilkin.json
+++ b/sources/us/mn/wilkin.json
@@ -24,7 +24,7 @@
             "replace": "$1"
         },
         "city": "CITY",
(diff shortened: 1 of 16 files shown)


iandees:
:clap:

sources/us/mn/wilkin.json
```diff
@@ -23,6 +23,8 @@
             "pattern": "^(?:[0-9]+ )(.*)",
             "replace": "$1"
         },
+        "city": "CITY",
+        "state": "STATE",
```
iandees: This should be `region` key, not `state`.

PR #1595, opened 2016-03-10
Add units to US/MT

trescube:
The majority of the changes are for fixing `number` and `street` for the sources that have the same format (`OwnerAddre` -> `AddressLin`).

`unit` has been aded where found


migurski:
Thanks for making these large-scale edits!

PR #1597, opened 2016-03-10
adds us-ca-hayward and friends

riordan:

PR #1598, opened 2016-03-10
Fix tests for single-element arrays

migurski:
https://github.com/openaddresses/openaddresses/pull/1597 should have passed.
- [x] Fix the actual tests
- [x] “Should” the tests to death.


iandees:
Hehe, I guess this is ONE WAY of doing it :smile:.

migurski:
I need to replace all the unit test language to use the word "should"—it’s never clear in these errors whether it’s telling you what it found, or what it wanted to find.

ingalls:
:+1:

PR #1604, opened 2016-03-12
Add units to US/NY

trescube:


migurski:
Damn, a [timeout I wasn't aware of hiding in the cache code](https://github.com/openaddresses/machine/blob/1c51c0bbcd32035054006b50c0975e020b0b8a26/openaddr/cache.py#L579-L583).

PR #1613, opened 2016-03-16
Support for several language versions of addresses

vesameskanen:
Added the guidelines how to handle address translations to several languages. Two example cases, Brussels  and Helsinki, updated to demonstrate how to use the new language tag.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -279,7 +279,7 @@ Additional metadata helps future proof the project!
 `note`        | A String containing a human readable note.
 `attribution` | **Deprecated:** Where the license requires attribution, add it here. example `CC-BY United Federation of Planets`
 `email`       | This email is used to send automated emails to the data provider if a user changes their data. Do not set unless the data provider wants to receive updates.
-`language`    | ISO 639-1 code for the language of the data. For example, `en`, `fr` or `de`.
+`language`    | ISO 639-1 code for the language of the data. For example: `en`, `fr` or `de`.
 
 #### Example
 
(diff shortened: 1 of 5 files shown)


migurski:
Great start! Looks like the tests need a mention of lanaguage, in here: https://github.com/openaddresses/openaddresses/blob/master/test/sources.js#L47-L57

riordan:
This is really cool. I rather like the approach.

Couple of questions:
- Is this designed to represent the dominant language for an area?
- How would this work for places where streets are named in "blended" ways, bringing multiple languages together (e.g. _Calle de Lago Road_ in the American Southwest)?

vesameskanen:
Hi @riordan, the proposed language tag does not hint anything about language dominance or priorities.That is actually an interesting point of view: for example, in Pelias context, which one of the available languages should be assigned to name.default?  I guess we need a sensible solution for that question.

Blended names are not a problem, as far as I can see. An address used by local Spanish-speaking citizens can be mapped via Spanish metadata, regardless of what kind of words (Spanish or English or Chinese) it contains. If Spanish people use the same form as English folks, then there is no need for a translation at all; declaring it as English is perfectly OK. In other words,  it is the context, not the content, that decides which language is associated tot an address.

vesameskanen:
Thanks @migurski for feedback - the tag validation is now updated.

migurski:
Looks good to me! @iandees, any feedback on this?

iandees:
This looks good to me. Thanks for the suggestion and implementation, @vesameskanen!

PR #1623, opened 2016-03-19
Added coverage geometries for Canadian places

migurski:


migurski:
Can’t find a new URL for the `ca/bc/prince_george` source; otherwise looks good.

Issue #1574, opened 2016-02-25
Pennsylvania Counties


aaronpdennis:
Found a datasource from 2007 for many of the missing Pennsylvania counties, seen below.

<img width="1030" alt="screen shot 2016-02-25 at 1 31 30 am" src="https://cloud.githubusercontent.com/assets/8019997/13311839/c214df2c-db5f-11e5-9d56-d524f7f780bd.png">

I'd like to add these address points to OpenAddresses, and I'm eager to put together the JSON source, but could use some guidance.

Is the coverage here statewide or should it be broken down into individual counties? I would guess the points should be separated into individual counties. Also, some counties already have a more up-to-date source in OpenAddresses than this aggregation done by the state in 2007. Should these counties be left out?

The `.zip` downloaded at the FTP url contains the data in ESRI GDB. Should this be converted to shapefile? Where would that shapefile have to be for use in OpenAddresses?

Data source url: http://www.pasda.psu.edu/uci/MetadataDisplay.aspx?entry=PASDA&file=PAMAP_BuildingPoint2007.xml&dataset=3

migurski:
If it's possible to split this into counties, I think that might be better.

aaronpdennis:
I'll break the FileGDB down into CSV files based on county, leaving out those counties that already have a more up to date source in OpenAddresses (like the counties publishing to an ESRI endpoint), and then put those CSV data files in a GitHub repo, then create source files for them in OpenAddresses. 

This datasource was a onetime project the state GIS office undertook in 2009, so I don't expect they'll be updating it again.

aaronpdennis:
I've processed out individual counties and identified the 18 counties in this dataset that could be valuable to OpenAddresses in this repo: https://github.com/aaronpdennis/pa-county-addresses

PR #1628, opened 2016-03-22
add 17 us/pa county sources

aaronpdennis:
These sources are extracted from the 2009 PAMAP program address Geodatabase. #1574


aaronpdennis:
It's possible and likely that these counties are maintaining a more up-to-date database of addresses. These static files are current up to 2009.

migurski:
Looks great!

screen shot 2016-03-22 at 1 34 41 pm

PR #1634, opened 2016-03-29
Add city of Nashville, TN, USA

orangejulius:
Found via the Nashville GIS website: http://www.nashville.gov/Planning-Department/Mapping-and-GIS/Interactive-Maps.aspx


iandees:
Looks good to me, thanks @orangejulius!

image

PR #1635, opened 2016-04-01
Lat/lon were flipped in the Wisconsin statewide source

iandees:


migurski:
It should be possible for us to detect something like this using the source geometry.

PR #1641, opened 2016-04-07
Oxford County, ON, CA

ingalls:


ingalls:
cc/ @migurski mind looking into this one? I was able to scrape it locally no problem.

migurski:
Huh, looks like a connection failure? Can you try making a trivial change to the JSON and trying it again with a new commit? I’m thinking it might be transient.

slibby:
POSTAL_CODE,PROVINCE don't exist in the service, @ingalls can you remove those and try again?

PR #1648, opened 2016-04-20
Add units to us sc

trescube:
Along with other fields and fixed source `data` URLs where required


iandees:
The City of Rockville regex seems to be failing in at least one case.

trescube:
Yeah, I wrote this regex way too late in the evening.  Will revisit in a bit.

iandees:
@trescube do you want to try fixing this regex again?

PR #1652, opened 2016-04-20
updated `data` URL, added `street`, removed `note`

trescube:


iandees:
I don't think we support extracting lat/lon with regex. I think there was a ticket asking about that elsewhere, but I can't seem to find it...

iandees:
Going to close this because we already have statewide data for Vermont.

PR #1653, opened 2016-04-21
Add units to us tx

trescube:
There were a number of sources that required `data` updates, too.

### DIFF
--- a/sources/us/tx/lubbock.json
+++ b/sources/us/tx/lubbock.json
@@ -5,8 +5,8 @@
         "state": "tx",
         "county": "Lubbock"
     },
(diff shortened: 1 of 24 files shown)


sources/us/tx/lubbock.json
```diff
@@ -5,16 +5,16 @@
         "state": "tx",
         "county": "Lubbock"
     },
-    "data": "http://216.167.160.18:88/2014/rest/services/2015External/MapServer/0",
-    "type": "ESRI",
+    "data": "http://ewebmap.ci.lubbock.tx.us/currentdata/Parcels/ParcelOwnerLanduseAddress.zip",
+    "type": "zip",
```
iandees: This should be `http` and the `conform` object's `type` should be `shapefile-polygon`.

trescube:
superseded by #1887

PR #1676, opened 2016-04-28
Add new us tx sources

trescube:
Found a bunch of new sources here:

http://mapsatbis.com/bisgis/rest/services


ingalls:
Hey @trescube these look great!! Super excited to get these in.

About half of them failed on openaddresses/hooked but it looks like the server that we were hitting simply couldn't keep up and timed out. Going to merge these for now and watch the results the next time they are cached.

PR #1684, opened 2016-04-29
Prescott russell

ingalls:


slibby:
Canada wins for having the least understandable standard geographies, the "United Counties of Prescott and Russell".

PR #1699, opened 2016-05-25
Extend the nz source

s01ipsist:
This adds some missing attributes and clarifies the license terms.

### DIFF
--- a/sources/nz/countrywide.json
+++ b/sources/nz/countrywide.json
@@ -8,18 +8,24 @@
     },
     "data": "http://data.openaddresses.io/cache/lds-nz-street-address-electoral-SHP.zip",
     "website": "https://data.linz.govt.nz/",
-    "year": "2014",
     "license": {
-        "text": "CCBY3.0",
+        "url": "http://creativecommons.org/licenses/by/3.0/nz/",
+        "text": "CC BY 3.0",
         "attribution": true,
+        "attribution name": "Land Information New Zealand",
         "share-alike": false
     },
-    "note": "This data is currated and produced by the New Zealand LINZ Data Service",
+    "language": "en",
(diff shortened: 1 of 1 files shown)


slibby:
Cheers!

s01ipsist:
No problem. The cached data file at http://data.openaddresses.io/cache/lds-nz-street-address-electoral-SHP.zip appears to be from Jan 2015 according to the timestamp of the files inside. Is it a manual process to update that? The latest data has 40,000+ more address points. Happy to help, not sure how.

migurski:
It is a manual process—I can push a new file there, if you’ve got something available at a URL?

andrewharvey:
According to https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#processing-tags the `conform` process only works with a single shapefile. The source data is at https://data.linz.govt.nz/layer/779-nz-street-address-electoral/ but you end up with a ZIP file containing two shapefiles.

I've gone ahead and downloaded the latest source data and combined it into a single shapefile at https://www.alantgeo.com.au/share/lds-nz-street-address-electoral-SHP.zip

The script I used to combine the two shapefiles from the source is at https://github.com/andrewharvey/openaddresses/blob/master/scripts/nz/lds-nz.sh.

I'm not 100% sure that this process of combining the two shapefiles is necessary and my script is okay, if all good I can create a pull request to merge in the script.

s01ipsist:
@andrewharvey :+1: 
I had done almost the same thing yesterday but not yet uploaded. 
LINZ say that the outputs will be changing sometime this year, but we'll have to deal with that when it happens.

andrewharvey:
@s01ipsist Two people doing it independently helps pick up any issues in either ones process. :wink:

PR #1704, opened 2016-05-31
added statewide Florida gathered by Mapwise/@b-may

### LINKED ISSUE #1633: us-fl proposal to replace loveland data with mapwise data
I realize this isn't a Pelias issues forum, but... I am proposing to replace the loveland version of Florida address data with data from MapWise. 

I've implemented Pelias address search into our web-based mapping app and have been researching geocoding problems. Our app just covers Florida, so I'm focusing exclusively on Florida. Two of the biggest problems traces back to the loveland address data. 

1) The loveland data does not include zip codes. This seriously confuses the geocoder when a zip code is included in an address search. Basically, zip codes cannot be used at all for searching, and that really confuses people!

2) The loveland data covers all of Florida, except Miami-Dade. The Pelias data importer is using the loveland data first, and ignoring the many counties of address point data available.

I have statewide parcel data sourced from Florida Department of Revenue (FDOR) and many counties. I've created a new Florida address file that includes zip codes, pin numbers and address types (single family residential, commercial, etc). In addition, unit numbers are separated out for several counties.

This version does not include counties where address point data is included in OA. The reasoning is the data would be redundant and the parcel centroid version of address data is not as accurate and complete as the address points data produced by counties.

There's a few counties with address point data "on-deck" to be added to OA, and I plan to update this version to remove those as they come on line in OA.

Another difference between the two data sets is this only includes parcel centroids that have an address number. I figured OA prefers actual addresses, but I can include all parcel centroids if that is preferred.

The data is available here: https://www.mapwise.com/data/florida_mapwise.zip

Thanks for your consideration.

trescube:
Fixed #1633

### DIFF
--- a/sources/us/fl/statewide.json
+++ b/sources/us/fl/statewide.json
@@ -1,0 +1,28 @@
+{
+    "coverage": {
+        "US Census": {
+            "geoid": "12",
+            "state": "Florida"
+        },
+        "country": "us",
+        "state": "fl",
+    },
+    "data": "https://www.mapwise.com/data/florida_mapwise.zip",
+    "note": "https://github.com/openaddresses/openaddresses/issues/1633#issuecomment-202908172",
+    "compression": "zip",
+    "type": "http",
+    "year": "2016",
+    "conform": {
+        "type": "csv",
+        "id": "id",
+        "lat": "lat",
+        "lon": "lon",
+        "number": "number",
+        "street": "street",
+        "unit": "unit",
+        "city": "city",
+        "district": "district",
+        "region": "region",
+        "postcode": "postcode"
+    }
+}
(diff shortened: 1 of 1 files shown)


trescube:
This gives OA coverage in 44 additional counties (as far as I can tell there's little to no overlap with existing sources):

BAKER
BRADFORD
CALHOUN
COLUMBIA
DESOTO
DIXIE
DUVAL
ESCAMBIA
FLAGLER
FRANKLIN
GADSDEN
GILCHRIST
GLADES
HAMILTON
HARDEE
HERNANDO
HOLMES
INDIAN_RIVER
JACKSON
JEFFERSON
LAFAYETTE
LEVY
LIBERTY
MADISON
MARION
MARTIN
MONROE
NASSAU
OKALOOSA
OKEECHOBEE
OSCEOLA
PASCO
POLK
PUTNAM
SANTA_ROSA
SEMINOLE
ST_JOHNS
ST_LUCIE
SUMTER
SUWANNEE
TAYLOR
UNION
WAKULLA
WASHINGTON

PR #1715, opened 2016-06-08
potential sources for koochiching and marshall counties in MN

mmdolbow:


migurski:
screen shot 2016-06-18 at 8 01 41 am

Curious about those scattered addresses—they reach pretty far out.

mmdolbow:
That's something not done in this project, is it? (Checking the results against the border of the source.) They are probably corrupted records that the source isn't even aware of.

iandees:
Nope, we take the source data as-is with the hope that downstream users will find and report errors.

We've been thinking about ways to include basic checks like this as part of data transformation steps.

mmdolbow:
Understood. That would be an interesting challenge to undertake; since there's often not high-precision boundaries available, it would be tough to rule out a lot of records based solely on their coordinates. The best approach might be a simple bounding box to get rid of the worst potential errors.

PR #1719, opened 2016-06-13
Dockerize NZ LDS data

migurski:
Retrieved a fresh copy of NZ data, previously stale since January 2015. Wrapped download script in Docker context and added descriptive README.


migurski:
screen shot 2016-06-13 at 4 36 48 pm

PR #1722, opened 2016-06-14
Point to New Jersey's REST address layer

iandees:


ingalls:
```
Could not retrieve this chunk of objects from ESRI source: HTTP 502
```

Maybe try restarting this one? Looks like their server hiccuped

migurski:
Looks like [we are considered abuse](http://s3.amazonaws.com/data.openaddresses.io/runs/89096/output.txt). Might make sense to cache this dataset using a much slower download process.

iandees:
It might be worth adding a retry mechanism with backoff?

migurski:
Yeah good idea, I’ll put that on the list.

iandees:
Pulling in the GDB is easier on us and their server, so I'm going to close this.

Issue #1726, opened 2016-06-17
us/ar/statewide.csv not showing street names


jeffklee:
There seems to be a problem with this file. The street names are coming in blank. I think the issue is that in the JSON mapping I see:

"street": "PSTR_FULNA",

It should be:

"street": "PSTR_FULNAM",

PR #1729, opened 2016-06-17
Updated US/AR with new column names

migurski:
“PSTR_FULNA ” is actually okay, but [the log](http://s3.amazonaws.com/data.openaddresses.io/runs/87439/output.txt) shows some other column name errors. Closes https://github.com/openaddresses/openaddresses/issues/1726.


migurski:
screen shot 2016-06-17 at 2 19 54 pm

PR #1737, opened 2016-06-20
Add nevada sources

trescube:
Carson City, NV is a county-equivalent: https://en.wikipedia.org/wiki/Carson_City,_Nevada


migurski:
screen shot 2016-06-20 at 8 36 50 am

PR #1754, opened 2016-06-22
updated `data` URL to arcgis-hosted service, appears to be updated mo…

trescube:
…re often


migurski:
Oh Henry.

screen shot 2016-06-21 at 10 01 48 pm

PR #1759, opened 2016-06-29
Generate a proper unique id for Finnish metadata entries 

vesameskanen:
Conform objects's id field for all Finnish source entries used a building identifier of the original data source. A large building can open to many streets, so the id was not unique. Now the source combines the building id with the index of each address assigned to it. The combined id should be truly unique.

A side note: 'npm test' complains about use of join function for id. However, looking at the specs for contributing, it should be OK. I 'm a bit confused about this contradiction.

### DIFF
--- a/sources/fi/ahvenanmaa-fi.json
+++ b/sources/fi/ahvenanmaa-fi.json
@@ -29,7 +29,11 @@
         "number": "COLUMN10",
         "lat": "COLUMN5",
         "lon": "COLUMN6",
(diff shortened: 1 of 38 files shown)


migurski:
This is great, thank you! We’ll have a look at the failing tests, maybe the tests or spec themselves are at fault.

migurski:
Replacing in favor of https://github.com/openaddresses/openaddresses/pull/1763.

PR #1763, opened 2016-06-30
Generate a proper unique id for Finnish metadata entries, with fixed tests

migurski:
This new PR updates @vesameskanen's https://github.com/openaddresses/openaddresses/pull/1759:

> Conform objects's id field for all Finnish source entries used a building identifier of the original data source. A large building can open to many streets, so the id was not unique. Now the source combines the building id with the index of each address assigned to it. The combined id should be truly unique.

Also, fixes tests to allow functions for `id` column.


vesameskanen:
Thank you for quick processing!

PR #1764, opened 2016-06-30
Update peel-region.json

kevin-peel:
Added municipality field to the data.  Modified data to change the MUNICIPALI field from the first letter of the municipality to the full lower tier municipality name to improve usefulness in geocoding.


ingalls:
This is fantastic! Thrilled to merge this!

PR #1773, opened 2016-07-06
added city of Vicksburg, MS, USA

trescube:


ingalls:
screenshot 2016-07-06 00 04 45

Issue #1797, opened 2016-07-06
us-fl-duval


b-may:
Duval County FL just brought online a new ArcGIS server with address points.

`http://maps.coj.net/coj/rest/services/DuvalMaps/duvalProperty_2012/MapServer/34`

trescube:
I'm having trouble opening the data in QGIS right now, it appears to stall about halfway thru, and I can't tell for certain what the fields are without seeing the source data.  I'll keep an eye on it.

b-may:
I scraped 1000 records into shapefile format with all of the fields included - you can download the sample here:

`https://www.mapwise.com/data/address_points_duval_test.zip`

trescube:
Thanks!  I'll check it out in a few minutes.

PR #1798, opened 2016-07-06
added Duval County, FL, USA

trescube:


trescube:
Fixed #1797

trescube:
I can't get this to load in QGIS (it stalls at ~60%) so I can't include a screenshot.

PR #1799, opened 2016-07-06
switched to more reliable fields

trescube:
The `PREFIX` field doesn't contain any data and the directional in the `SUFFIX` field could be the prefix, so addresses with a prefix directional were being formatted incorrectly.  This swaps out for more reliable fields prefixed with `COL_`.


trescube:
For example, `-81.9077064,28.0247194,1440,Combee Road South,,,,,33801,,` has the directional incorrectly placed, so it should actually be `-81.9077064,28.0247194,1440,South Combee Road ,,,,,33801,,`.

Issue #1720, opened 2016-06-14
Los Angeles County Address Source


johnhickok:
The data source for Los Angeles County addresses referenced in this project is over 4 years old:
https://github.com/openaddresses/openaddresses/blob/master/sources/us/ca/los_angeles.json

Los Angeles County provides a countywide source of address points and lines here:
http://egis3.lacounty.gov/dataportal/2014/06/16/2011-la-county-street-centerline-street-address-file/

he 250 MB download, a file geodatabase containing line and point address feature classes, is updated weekly:
http://egis3.lacounty.gov/dataportal/wp-content/uploads/ShapefilePackages/CAMS.zip

The City of Los Angeles' OPen Data Portal hosts this file - not sure if there is overlap:
https://data.lacity.org/A-Well-Run-City/Addresses-in-the-City-of-Los-Angeles/4ca8-mxuh

trescube:
Wow, they need to update their site, that file is actually 454MB and not downloading terribly fast. 

I should have an update to the source sometime today.

trescube:
Ah, this is GDB which OA doesn't support yet.  I'll check to see how much difference there is to see if it's worth caching an export on s3.

johnhickok:
L.A. County chose the gdb format due to the file size. The streets and
address points cover 88 cities and over ten million residents.

The plan is to get all the public agencies within the county to participate.

It's ambitious; it might have a long way to go.

John Hickok, PLS, GISP

*Try paperless...*

ingalls:
Working on getting GDB supported - No timeline yet as it's been a bit since I've jumped into the code but as long as the GDAL driver works this should be fairly straighforward

johnhickok:
If it helps, I wrote some gdal code for that file gdb here:

http://egis3.lacounty.gov/eGIS/wp-content/uploads/2011/10/baselayers_toolkit.zip

It was for a presentation.

John Hickok, PLS, GISP

*Try paperless...*

ingalls:
LA data has since been added (As well as GDB support).
Also just added parcel support here: https://github.com/openaddresses/openaddresses/pull/5816

PR #1807, opened 2016-07-07
GDB Support [Don't Merge]

ingalls:
Don't merge until I get a machine branch that actually supports this.

Refs: https://github.com/openaddresses/openaddresses/issues/1720


migurski:
New GDB stuff is live on Machine.

Issue #1831, opened 2016-07-11
Update us/ma/statewide to use statewide address points instead of parcels


iandees:
New layer is here: http://gisprpxy.itd.state.ma.us/arcgisserver/rest/services/AGOL/MassGIS_Master_Address_Points/MapServer/0

iandees:
If someone takes this, I'd love to see a step-by-step guide written about how to update an existing source with a new URL.

trescube:
I can do that

trescube:
It might be a few days.

erictheise:
I've been looking for a way in to OA and would be glad to take this on but frankly I don't know enough to tell whether this is straightforward or has some hidden complexity.

trescube:
Eric, you appear to have deleted your comment but if you're still watching it would be great to work with you on updating this data source.  It would be a great introduction to contributing to OA.

trescube:
Great, sounds good!

slibby:
closed via https://github.com/openaddresses/openaddresses/pull/1832

erictheise:
I was still thinking about @iandees' ask, which was to contribute a step-by-step write up, presumably to https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md. I started in on a blog post which, when finished, I'd strip back to something less chatty.

The thing is, the bulk of it is about coaxing data out of the ESRI server and that's not really specific to a change in data source. Maybe I should let it go. Thoughts?

iandees:
@erictheise I'd be curious to see your write-up including how you ended up coaxing data out of ESRI. I have my methods of doing it (as do others that contribute here) and I want to see if your methods are similar.

The eventual goal for me is to build a web app that let's people submit their own data (including an ESRI endpoint), see a sample of the source data, build the conform/mapping information, and preview the output result. Documenting the manual process is good first step towards making it automated.

erictheise:
Okay, @iandees, I'll finish up my post in the next couple of days and see how if the info could be shoehorned into CONTRIBUTING.md. Will let you know. Maybe reopen the issue?

PR #1832, opened 2016-07-12
Update us/ma/statewide to use statewide address points instead of par…

erictheise:
…cels. #1831 

My first contribution, so please treat with caution. For this
```
"street": [
    "PRE_MOD",
    "PRE_DIR",
    "BASE",
    "POST_MOD",
    "POST_DIR"
],
```
I've verified that the order of `PRE_MOD` and `PRE_DIR` are correct with respect to the source's aggregate field, `STREETNAME`. There seem to be no records having entries for both `POST_MOD` & `POST_DIR` so I've made my best guess at their best concatenation order.

### DIFF
--- a/sources/us/ma/statewide.json
+++ b/sources/us/ma/statewide.json
@@ -7,28 +7,26 @@
         "country": "us",
         "state": "ma"
     },
-    "data": "https://www.dropbox.com/s/15bzesrcesp5wt2/SHPStatewideParcels.zip?dl=1",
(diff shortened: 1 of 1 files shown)


trescube:
I noticed that `STREETNAME` field in the source data already contains the concatenated street fields instead of piecing together the other 5 components, so you could use that instead.  Other than that, it looks great!

erictheise:
I started with `STREETNAME` but when I asked @iandees about it in Slack he said 

> i usually like to concatenate the street names by specifying the parts in a list so that if we add the individual columns in the future it'll be easier to spot the ones that have the columns broken out already

so I went with his approach.

trescube:
Ah, makes sense.

trescube:
massachusetts

PR #1835, opened 2016-07-15
LA (First GDB Source)

ingalls:
cc/ @migurski


migurski:
Sorry for the delay. 

I believe GDAL should now be up-to-date throughout the live machine code. I added `ppa:ubuntugis` to the machine server and upgraded libgdal-dev and python-gdal to 1.11.2. Then I did the same thing to the AutoScale AMI.

This might require a re-bump.

migurski:
FYI, the change in 88b49f6 reverted the source back to the version from before debec0d, so it was still using a cache. New change should be a true cache-buster.

migurski:
The data worked, but the sample did not. I forgot there was a separate code path for sampling a source. 

screen shot 2016-07-16 at 9 24 59 am

migurski:
Hm, that should have worked.

ingalls:
Huh that's strange, looks like it worked but reported a fail...

migurski:
Yeah. It seems to be related to how GDB’s are a directory rather than a file. Need to do some work on that machine PR.

migurski:
PHEW.

tl;dr on this process is that adding a new supported file format requires updates in both the record-sampling and conform code paths, because they are different. GDB in particular was challenging, because unlike other formats you `ogr.Open()` the directory name, and not a file.

PR #1873, opened 2016-07-21
fixed `number` and `street` field name

trescube:


trescube:
massac-il

trescube:
Incorrect casing of field name resulted in no addresses.

PR #1888, opened 2016-07-24
Addresses for August Town, Jamaica

milovanderlinden:
A local mapping initiative on Jamaica called "Interactive Community Mappers" is currently in the process of mapping parts of Jamaica, starting with August Town. We would like to make these addresses available to OA, that is why we are tagging all the addresses with source=icm_jamaica in OpenStreetMap. We then periodically run an overpass query:
```
/*
This is an example Overpass query.
Try it out by pressing the Run button above!
You can find more examples with the Load tool.
*/
node
  [source=icm_jamaica]
  ({{bbox}});
out;
```
To extract these addresses and provide them as GeoJson here.

### DIFF
--- a/sources/ja/august_town.geojson
+++ b/sources/ja/august_town.geojson
@@ -1,0 +1,1274 @@
+{
+  "type": "FeatureCollection",
+  "generator": "overpass-turbo",
+  "copyright": "The data included in this document is from www.openstreetmap.org. The data is made available under ODbL.",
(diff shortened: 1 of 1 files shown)


migurski:
Thank you! I’ve uploaded the data to our S3 cache, and issued a replacement PR in #1889 which includes our preferred JSON source format.

screen shot 2016-07-25 at 3 34 27 pm

PR #1893, opened 2016-08-04
Added Emilia-Romagna in Italy

sergiyprotsiv:
Hi,

I noticed that some regions in Italy have posted open addresses on the INSPIRE portal, will try to find useful data. The first one is the region of Emilia-Romagna: http://geoportale.regione.emilia-romagna.it/it/download/dati-e-prodotti-cartografici-preconfezionati/database-topografico/gestione-viabilita-e-indirizzi/dati-preconfezionati-dbtr-civico-ncv_gpt

The regional hierarchy in Italy has both regions (with 2-digit codes, e.g. 45 for Emilia-Romagna) and provinces (with 2-letter codes, e.g. BO for Bologna). I assumed that the region would be the "state" and the province would be the "county". Provinces are gradually losing their role, so I think it makes sense to have regions as the main hierarchical unit.

I think this data can replace it-45-bologna.json,  it-45-ferrara.json, and it-er-rimini.json.

Let me know if I am doing anything wrong, this is my first experience with GitHub pull requests.

Cheers,
Sergiy


migurski:
Thank you Sergiy! In order for these files to work, their names must end with ".json" instead of ".txt". Can you update your pull request with this change?

sergiyprotsiv:
Sure, this should work.

iandees:
image

Another huge swath of Italy, thanks!

migurski:
👏 👏 👏

PR #1895, opened 2016-08-04
Added Friuli-Venezia Giulia

sergiyprotsiv:
Added the regionwide layer for Friuli-Venezia Giulia in Italy.


iandees:
image

Looks good, thanks!

PR #1896, opened 2016-08-04
Added Toscana

sergiyprotsiv:
Another region-wide source, though with two issues:
- Share-Alike license
- The zip that gets downloaded contains another 7zip file which finally unzips to the folder with files. Not sure if the machine can handle it.


iandees:
Yea, looks like we don't support .7z files natively. There might be some extra work there.

migurski:
screen shot 2016-08-09 at 9 07 18 pm

PR #1898, opened 2016-08-04
Added Piemonte

sergiyprotsiv:
Added Piemonte, which appears to have very good coverage.
This is the WMS server at http://geomap.reteunitaria.piemonte.it/ws/taims/rp-01/taimsbasewms/wms_sfondo_cart_rif_bn?request=getcapabilities&service=WMS (layer AccessoCarraio), which I converted to shape file and put on Dropbox. 

Again, let me know if there is a better procedure.

PR #1897, opened 2016-08-04
Added Sardegna

sergiyprotsiv:
Adding the data from most of Sardegna. 

There are two files to be dowloaded from the page indicated  (strato03 and strato09) and a trivial join between house, street, location and municipality tables is needed.

I made the processing offline in QGIS and uploaded the processed file to dropbox. Let me know if another procedure is preferred.

Closes #1477.

PR #1894, opened 2016-08-04
Added Lombardia in Italy

sergiyprotsiv:
This covers most of Lombardia, though some municipalities are missing, most notably the city of Milan. There seems to be ongoing work to bring the data from all municipalities online, so hopefully this gets updated in the near term.

PR #1899, opened 2016-08-05
Added Umbria

sergiyprotsiv:
Added a good dataset for Umbria.

It is generally available via the REST service (http://geo.umbriaterritorio.it/ArcGIS/rest/services/Public/ECOGRAFICO_CATASTALE1_WGS84/MapServer/0), but the resulting files contain only the code for municipality, not the name. So this is the data downloaded from the REST service and joined with municipality data from Istat.


nvkelso:
Re-run this please

trescube:
What's the next step with this one?  It sounds like the REST data is missing the city name which must be looked up elsewhere in a post-processing step, but the standard operating procedure is to register a script to recreate the data in `/scripts`?

sergiyprotsiv:
I have all the scripts for the Italian regions, but did not find time to figure out exactly how to do this via Github (a bit of a novice here). If you can suggest the proper way, I'll try to put this up during the weekend. Or I can send these via email.

trescube:
I've not had to handle custom scripts, but looking at other things in `/scripts`, it looks like we're favoring creating a directory in `/scripts`, adding all the custom scripots, then creating a docker file to run the exporting scripts.  See: 
https://github.com/openaddresses/openaddresses/tree/master/scripts/us/tn

If you want to tackle it, please do so, otherwise email everything to stephen@mapzen.com and I can take a crack at it.

trescube:
The same probably applies to #1900, #1898, #1897, and #1894.

migurski:
Yeah — the docker part is optional, but helps in cases like Australia’s data where it's necessary to run a complex set of pieces like PostGIS, etc.

iandees:
I'd rather we pull straight from the source and not manually merge in city information. For me, pulling from a primary source trumps more-complete data.

sergiyprotsiv:
Re-run this please

sergiyprotsiv:
This is ready to be merged, I think

PR #1901, opened 2016-08-09
Added Italy countrywide

sergiyprotsiv:
Here is a big one for Italy: countrywide coverage, though with a few issues (see below).

This data has been collected by the Italian Ministry of Environment and Protection of Land and Sea and has been made available via their WFS server.

The license is CC-BY-SA 3.0, according to the dataset’s page on the INSPIRE geoportal, and I was not able to find other license information on the website of the publisher, so I assumes this license.

The data has consistent, though incomplete coverage. The total number of addresses is about 14 million, whereas you would expect at least 20-25 million given Italy’s population. Some towns and parts of cities are missing compared to OSM, but overall it covers the whole territory fairly evenly.

Closer inspection makes it clear that the addresses have been interpolated from road segments, sometimes rather disastrously (when the houses follow a strange numbering pattern, e.g. 6500 followed by 8400, it would add 1900 points). So, I would suggest that other sources should take priority where they are available (I put 4 in the accuracy field).

On the plus side, this is the only datasource in Italy I found that has postal codes, which in itself could be a great benefit and complement other sources.

I am adding it as a countrywide source, though it could be split into regions or provinces. However there are a lot of addresses without the regional attribute, so not clear where to put those (I could do spatial matching against provincial boundaries, but that seems like too much processing for open addresses as it seems most users do their post-processing anyway).

To prepare the dataset I downloaded all addresses from the WFS server (which is fairly slow, it took several days to download, so would be great if making it available here would keep the server from being hit too much and keep providing other nice data). I then combined all the query results in a PostGIS database to do some basic checks and remove the duplicates (there were about 200K, all other them due to the usual issue of some records dropping the leading zero in their IDs…). No further processing other than putting it into a single shapefile and uploading to Dropbox.

Let me know if you need more info or would like me to change a few things.

Cheers,
Sergiy

### DIFF
--- a/sources/it/countrywide.json
+++ b/sources/it/countrywide.json
@@ -1,0 +1,30 @@
+{
+    "coverage": {
+        "ISO 3166": {
+            "alpha2": "IT",
+            "country": "Italy"
+        },
+        "country": "it"
+    },
+    "data": "https://www.dropbox.com/s/sut0ldbf6vzukiw/civici_italia20160807.zip?dl=1",
+    "website": "http://wms.pcn.minambiente.it/ogc?map=/ms_ogc/wfs/Numeri_Civici_2012.map&service=wfs&request=getCapabilities",
+    "license": {
+        "url": "https://creativecommons.org/licenses/by-sa/3.0/",
+        "attribution": true,
+        "share-alike": true
+        },
+    "type": "http",
+    "compression": "zip",
+    "language": "it",
+    "conform": {
+        "number": "CIVICO",
+        "street": "NOME",
+        "city": "COMUNE",
+        "postcode": "CAP",
+        "district": "PROVINCIA",
+        "region": "REGIONE",
+        "id": "ID",
+        "type": "shapefile",
+        "accuracy": 4
+    }
+}


migurski:
This is exciting, thank you! Would you be able to provide the script you used to download the data? For large caches like this, it’s been nice to keep those available. Here's an example: https://github.com/openaddresses/openaddresses/tree/master/scripts/us/tn

napo:
this dataset has a little problem: the frequency of update (I think that it's no more updated)
I suggest you to take a to this work made by Cesare Gerbino with the collection of all the open data sources related to the civic numbers.
https://cesaregerbino.wordpress.com/2015/09/27/numeri-civici-open-data-in-italia-disponibile-la-release-3-0-della-raccolta/
... there is always a problem between licenses

nvkelso:
Re-run this please

nvkelso:
@migurski Bot is out of memory and is hungry:

From http://s3.amazonaws.com/data.openaddresses.io/runs/129262/output.txt:

```
2016-11-22 18:42:10,516     INFO: Processed data in /tmp/worker-9oixq4lk/work-6ah21aro/out/process_one-3qo6n0yn/out.csv
2016-11-22 18:44:59,721  WARNING: Error in process_one.process()
  File "/usr/local/lib/python3.4/dist-packages/openaddr/process_one.py", line 70, in process
    preview_path = render_preview(conform_result.path, temp_dir, mapzen_key)
  File "/usr/local/lib/python3.4/dist-packages/openaddr/process_one.py", line 100, in render_preview
    preview.render(csv_filename, png_filename, 668, 2, mapzen_key)
  File "/usr/local/lib/python3.4/dist-packages/openaddr/preview.py", line 31, in render
    points = project_points(iterate_file_points(src_filename))
  File "/usr/local/lib/python3.4/dist-packages/openaddr/preview.py", line 187, in project_points
    points.append((geom.GetX(), geom.GetY()))
MemoryError
```

sergiyprotsiv:
I guess this file is too large, my local machine was struggling with it as well. I can split the addresses into provinces as part of the processing (and hence create several sources), but that would have to be based on polygons, not text attributes, so might be harder to handle automatically in the future. 

That said, as napo mentioned this source is not being actively updated, so I would view this source as a placeholder until better sources become available across the country. If it is split into provinces, it would be even easier to handle this process.

I am sorry for not providing a good writeup of the Italian sources, had too many commitments, will try to do so soon.

As for Cesare Gerbino's site, it was one of the main gateways I used to get many of the sources, but I thought it was always better to link to official repositories rather than to pre-processed work due to clearer provenance and licensing.

migurski:
We're doing a whole-country run in #2157 too, so I’d love to fix the previews to work.

migurski:
I have a possible fix open in https://github.com/openaddresses/machine/pull/456; I'll re-run this when it’s deployed.

migurski:
Re-run this please.

migurski:
🎉 🇮🇹💥

nvkelso:
Wow!

What does this mean for the other open Italy PRs? Are they using another data source than this, or same and should they be closed?

sergiyprotsiv:
I would not delete any of the others, since this data has several drawbacks:
- It is interpolated from street segments (at least in the places I checked back when creating the source)
- It is no longer updated (I think it was a one-time task the ministry took before the 2011 census, but I might be wrong)
- It has a share-alike license

The other sources I have added in the summer mostly do not share these drawbacks, but they are only available for about half the country (in terms of population).

cesaregerbino:
Hi all! It's fine to see now Italy with more addresses on OpenAddresses .... :-)

As @napo said above I've collected several open datasets about addresses in Italy (published from several public administrations like Regions, Provinces and Cities): you can see them here http://www.cesaregerbino.com/ItalyOpenAddresses/ItalyOpenAddresses-V03.html (in Italian but shoud be quite clear .... ). I've also used  the WFS service by the Italian Ministry of Environment and Protection of Land and Sea (in my map you can see this dataset using the max map detail activating the "Civici Geoportale Nazionale - (WFS)" layer .... ). 

In my collections there are some datasets that are more updated and more detailed that those of the WFS service above: you can see the links at the original sources in a column in this CSV file ... http://www.cesaregerbino.com/ItalyOpenAddresses/List/ItalyOpenAddressesList-V03.csv

Could be possible to integrate these data in OpenAddresses ? Their licences should allow it .... 

Any chance to collaborate?

sergiyprotsiv:
Hi Cesare, when creating these sources back in the summer I have indeed used your website a lot as the gateway to individual data sources. The data source for the map in this PR is exactly the "Civici Geoportale Nazionale - (WFS)" that you mention.

I have also separately added official sources for Toscana, Piemonte, Emilia-Romagna, Umbria, Friuli, Lombardia, Sicilia, Sardegna. Most of those are better quality than this source, but they do not have full coverage and they rarely have postal codes. Also, most of OpenAddress data does not have the ShareAlike license, while this dataset (and a few provincial ones) have it - I would say that at least it would be good to separate the two.

There seem to also be open data sources for at least Abruzzo and Puglia, maybe for Liguria too. If you can help adding those (was hard for me without knowing Italian), would be fantastic.

Otherwise, I think it is up to the maintainers of OpenAddresses to decide on the best way to structure individual sources (link to the consolidated source by Cesare or create new sources as new data becomes available). My current understanding is that the coverage is roughly equivalent.

migurski:
Generally, OA is fine with overlapping sources like this, especially if they have license difference such as a share-alike requirement.

sergiyprotsiv:
Technically, this source can be sources straight from the REST server: http://www.pcn.minambiente.it/arcgis/rest/services/Civici_Italia_2012/MapServer/0
It is however 14 million addresses (and the server itself is flaky), so I do nothing it is likely to finish in one go and need some hand holding. Would be great to increase the default timeout for esri2geojson to 2 mins or so - many large sources take more than 30 seconds to just return the count.

PR #1905, opened 2016-08-16
added Alabama statewide

trescube:


trescube:
The server keeps timing out after ~1/3 of the data, whether running in the machine or dumping with pyesridump.  Will keep trying.

trescube:
Closing w/o merge, the data is too unreliable.

PR #1909, opened 2016-08-18
Updated source and conform for Carlton, MN

mmdolbow:
Got a better source for Carlton. I'm certain I don't have the regex quite right for the city and region tags because the typical contents are "<city> MN" - only a space between the city name and the state, no comma like the regex I have here. (I'm crummy at regex and trying to learn more.)


trescube:
The `PHYSCITY` field doesn't contain commas though your regex is expecting them.  Instead of `^([^,]+)`, you can probably use `^(.*) MN$` for `city` and `^.* (mn)$` for `region`.

mmdolbow:
Thanks for the tips, I'll try that!

sources/us/mn/carlton.json
```diff
@@ -9,11 +9,31 @@
...
+        "street": {
+            "function": "regexp",
+            "field": "PHYSADDR",
+            "pattern": "^(?:[0-9]+) (.*)",
+            "replace": "$1"
+        },
+        "city": {
+            "function": "regexp",
+            "field": "PHYSCITY",
+            "pattern": "^(.*) MN$"
```
trescube: Sorry, I forgot to mention that you'll have to include `"replace": "$1"` in both `city` and `region` to replace then entire value with the first captured group.  It's very close, but the `city` and `region` fields in the output are blank.

sources/us/mn/carlton.json
```diff
@@ -9,11 +9,31 @@
...
+        },
+        "region": {
+            "function": "regexp",
+            "field": "PHYSCITY",
+            "pattern": "^.* (MN)$"
+        },
```
trescube: Add `"replace": "$1"` here, too.

migurski:
I see some trailing spaces in the sample data, e.g. `"ANYTOWN MN       "`; something like `"^(.*) MN\\s*$"` may fix it.

mmdolbow:
OK, tried that fix.

iandees:
I'm going to close this as these two sources changes appear to have been added to the master branch already.

PR #1910, opened 2016-08-18
Create Douglas County

mmdolbow:
Pretty sure I got this one right

### DIFF
--- a/sources/us/mn/douglas.json
+++ b/sources/us/mn/douglas.json
@@ -1,0 +1,31 @@
+{
+    "coverage": {
+        "US Census": {
+            "geoid": "27041",
+            "name": "Douglas County",
+            "state": "Minnesota"
+        },
+        "country": "us",
+        "state": "mn",
+        "county": "Douglas"
+    },
+    "data": "http://dc-web.co.douglas.mn.us/arcgis/rest/services/AGOL/Data_Layers/MapServer/41",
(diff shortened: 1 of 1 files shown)


iandees:
Looks good, thanks!

trescube:
The `number` and `street` fields are already split apart so you don't have to use the regexes.  The source fields are `HNUM_HALF` (there's also `HNUM` but some addresses are 1/2's) and `STREET`.

mmdolbow:
I saw those splits, but when using their interactive map, the fields I specified were what popped up in an "Identify" window, so I thought those were more likely to be complete. For example, there are some records with additional Unit numbers in the ADDRESS line, with separate locations, like these two:
http://dc-web.co.douglas.mn.us/arcgis/rest/services/AGOL/Data_Layers/MapServer/41/2
http://dc-web.co.douglas.mn.us/arcgis/rest/services/AGOL/Data_Layers/MapServer/41/3

trescube:
There's a separate field for `unit`that shouldn't be included in the `street` or `number` field.  Unfortunately this source doesn't separate the unit field out individually like it does for house number and street but you can use a `regexp` to extract it.  Something like this should work: `^.* #(.*)$/`.

mmdolbow:
Thanks. Gonna do another PR to change this - looks like the "Unit" info is available in the "Location" field

PR #1911, opened 2016-08-19
Update Douglas with specific fields

mmdolbow:
Thanks to @trescube suggestion


trescube:
Thanks @mmdolbow !

douglas

PR #1912, opened 2016-08-19
Suggestions for contributing.md

mmdolbow:
Since I'm trying to learn regex, I figured a bit more explanation around the examples shown in the regex area might help. Happy to phrase it a different way if desired!


migurski:
This is great, thanks Mike!

PR #1914, opened 2016-08-22
removed duplicate fields from NY statewide

trescube:
Removed duplicate fields and fixed `number` field


trescube:
The source is timing out repeatedly (probably due to 4.3M records) but I'm merging anyway.

PR #1917, opened 2016-08-29
adding Spanish postcodes

thatdatabaseguy:
There is a postcode field in the processed CSVs for Spain but it wasn't getting mapped during the transform.


migurski:
Thank you, Al!

PR #1918, opened 2016-08-29
adding Austrian postcodes

thatdatabaseguy:
All of the Austrian files include postcode as well in the PLZ field.


migurski:
Thanks Al!

PR #1919, opened 2016-08-29
adding Brussels postcodes

thatdatabaseguy:
Postcodes are available in the PZNC field for Brussels.


migurski:
Looks good to me.

PR #1920, opened 2016-08-29
using number and street fields for Windsor, Ontario

thatdatabaseguy:
Was seeing some weird street names in the city_of_windsor file. 

Looks like it's splitting the full address field, using the first numeric token as the house number and the rest of the field as the street. Some of the addresses in this data set look like "239, 251 DROUILLARD RD", which splits into "239" and ", 251 DROUILLARD RD" as the street name. However, it looks like the source already contains separate number and street fields, so using those instead.


migurski:
screen shot 2016-08-29 at 11 23 34 am

PR #1923, opened 2016-08-29
adding postcodes for Iceland

thatdatabaseguy:
Available in the POSTNR field


migurski:
screen shot 2016-08-29 at 11 21 19 am

PR #1925, opened 2016-08-30
Netherlands house numbers - letter should come after number

thatdatabaseguy:
Was looking at Dutch house numbers in OSM and they usually take form "number + letter". The OpenAddresses config was using the inverse.

Format-wise, it didn't seem from [conform.py](https://github.com/openaddresses/machine/blob/master/openaddr/conform.py) that there was a way to do a string join with more than one separator, but it might be worth considering that for the conform spec. In the Netherlands, as far as I can tell, when there's a "huisletter", it would usually be concatenated onto the number with no space e.g. "112B", whereas the "huisnummertoevoeging" ("house number addition", usually also numeric) is joined using a hyphen e.g. "112-11". If all three fields exist, it's "112B-11".

### DIFF
--- a/sources/nl/countrywide.json
+++ b/sources/nl/countrywide.json
@@ -13,7 +13,7 @@
 	"conform": {
 		"lon": "lon",
 		"lat": "lat",
-		"number": ["huisletter", "huisnummer","huisnummertoevoeging"],
+		"number": ["huisnummer", "huisletter", "huisnummertoevoeging"],
 		"street": "openbareruimte",
 		"postcode": "postcode",
 		"city": "woonplaats",


migurski:
I’m not sure multiple separators is doable with a regexp, even. Worth some thinking.

thatdatabaseguy:
Thanks for the all the quick merges.

Re: multiple separators, one possibility would be to define a new key, "separators" perhaps, which takes a list of length len(fields) - 1, where the _i_-th separator in inserted before field i + 1 as long as one of the fields <= i exists (to allow for NULL/empty values). The resulting conform would look like:

```json
"conform": {
    "number": {
        "function": "join",
        "fields": ["huisnummer", "huisletter", "huisnummertoevoeging"],
        "separators": ["", "-"]    
    }
}
```

A more flexible alternative (though probably in YAGNI territory) would be to define a format function:

```json
"conform": {
    "number": {
        "function": "format",
        "fields": ["huisnummer", "huisletter", "huisnummertoevoeging"],
        "format": "$1$2-$3"
    }
}
```

and sub in the variables from there (using shell-style variables here because that should be familiar to non-Python developers, but could as easily be Python format strings, which would also allow type casting e.g. for data sets that store house numbers as floats).

migurski:
I like the "format" syntax, a lot!

iandees:
Same! The `format` version looks really great.

PR #1929, opened 2016-08-31
Adding postcodes to several US counties/cities

thatdatabaseguy:
Went through the US counties that didn't already have postcodes and added them to over a dozen county/city files, including a few larger cities like Seattle and Spokane.

Also removed postcodes from a few files that were using the ZIP from the property owner's mailing address rather than the situs ZIP.


iandees:
Awesome, thank you!

Are you interested in joining us on our Slack team to chat about this stuff? Also, I'm happy to add you to the org so you don't have to maintain a fork for PRs.

thatdatabaseguy:
Think I'm already on the Slack, but will pop in. Sure, I'd join the Github org.

Re: build, think I've fixed everything on my side, errors are about downloads/timeouts (or: the Central Limit Theorem of county GIS servers).

thatdatabaseguy:
Ok, fixed some of the failing sources and removed a few files from the PR where the sources couldn't be tracked down.

Thanks for adding me to the org, will merge this in.

PR #1936, opened 2016-09-02
using a "/" to delimit numero civico from esponente in Italian house numbers

thatdatabaseguy:


thatdatabaseguy:
This passes, but will want to rerun after the next machine deploy now that https://github.com/openaddresses/machine/pull/380 is in master.

migurski:
I’ll tickle this when I version + deploy machine.

PR #1937, opened 2016-09-03
Replaced several uses of "split" with non-deprecated regexp function

migurski:


migurski:
None of the failures are new, so I’m merging this.

PR #1938, opened 2016-09-05
Added more-accurate New York State coverage polygon

migurski:
Using county list from http://gis.ny.gov/gisdata/inventories/details.cfm?DSID=921.


migurski:
Merging this because we know that New York State source works.

PR #2041, opened 2016-09-15
added `unit`, corrected case for `number` and `street`

trescube:
`number` and `street` field names were incorrectly cased so no data was coming out


trescube:
accomack

Issue #2030, opened 2016-09-15
Add addresses in Lafayett, LA


nvkelso:
Status:

- [ ] **Research sources** and propose best source (and it's data license)
- [ ] **Review proposed source**
- [ ] **Approve source** based on review comments
- [ ] **Add** to OpenAddresses

_NOTE: Proposed source notes should include basic column mapping to OpenAddress data model._

davidchiles:
ArcGIS REST Services Directory: http://lcgmaps.lafayettela.gov/arcgis/rest/services

iandees:
Here's an address layer: http://lcgmaps.lafayettela.gov/arcgis/rest/services/BaseLayerServices/LCG_Address_Points/MapServer/0

slibby:
closed via #2050

PR #2050, opened 2016-09-20
Change source for us-la-lafayette

davidchiles:
This source has 102,341 features while the previous source had 49,672. 

A possible problem is the `CITY` field uses `LAF PARISH` which looks like shortened "Lafayette Parish".

Related #2030.


migurski:
Looks like an ESRI server timeout. Asking machine to
re-run this, please.

PR #2053, opened 2016-10-03
Updated CA province paths and coverages

migurski:


migurski:
screen shot 2016-10-02 at 6 01 38 pm

PR #2054, opened 2016-10-03
Updated Bucks County ESRI token

davidchiles:
Not sure how often this changes.

### DIFF
--- a/sources/us/pa/bucks.json
+++ b/sources/us/pa/bucks.json
@@ -9,8 +9,9 @@
         "state": "pa",
         "county": "Bucks"
     },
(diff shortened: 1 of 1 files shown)


migurski:
Hm, [for Tennessee we set up a script to do the token out of band](https://github.com/openaddresses/openaddresses/tree/master/scripts/us/tn). Might be necessary here.

davidchiles:
I used https://gisweb.co.bucks.pa.us/apps/floodplainviewer/ and then looked at the network requests to find the token.

migurski:
Makes sense! I verified that I could find the token and get the data. Would you mind making two changes to this PR? 

1. Update the data URL to https://s3.amazonaws.com/data.openaddresses.io/cache/us/pa/bucks.zip, a GeoJSON cache of the live server.
2. Add [a website tag](https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#optional-tags) pointing to https://gisweb.co.bucks.pa.us/apps/floodplainviewer/

davidchiles:
Thanks @migurski  for the help. I think I got it now.

PR #2058, opened 2016-10-07
rewrite specifying minimum amount of info needed

trescube:
I was talking with a friend who works with non-profit and the county government of York, PA about the project and asked him to look over the contribution example.  He recommended pointing out the bare minimum amount of information that we're looking for to allay any concerns about privacy.


migurski:
👍

iandees:
This looks good, thanks for the changes.

PR #2061, opened 2016-10-13
Create grant.json

mmdolbow:
Grant County has a new interactive map up. No addresses, but parcels. They aren't perfect, for example vacation homes don't have a true property address, so the best field is blank for those records.


migurski:
screen shot 2016-10-12 at 11 38 41 pm

migurski:
Don’t mind me, just testing new 3.x release:

Rerun, this please.

PR #2072, opened 2016-10-19
added Grant County, Washington

trescube:


sources/us/wa/grant.json
```diff
@@ -0,0 +1,31 @@
...
+    "data": "https://gis.grantcountywa.gov:6443/arcgis/rest/services/EveryoneData/Addresses/MapServer/0",
+    "website": "http://grantcountywa.maps.arcgis.com/home/index.html",
+    "type": "ESRI",
+    "conform": {
+        "type": "geojson",
+        "number": "ADDRNUM",
+        "street": [
+            "PREADDRNUM",
+            "FULLNAME",
+            "ADDRNUMSUF"
```
iandees: This is probably for the street number, not street name.

trescube:
grant

sources/us/wa/grant.json
trescube: It's correct as-is, I went thru the data to verify.  Weird, I know.

PR #2079, opened 2016-10-20
Add Bucks County Docker script

migurski:
Expands on #2054 with helpful instructions.


migurski:
I updated the machine’s github token, so testing this again:

Re-run this, please.

iandees:
I think I'd prefer to close this and direct people to use `esri2geojson` directly. Any thoughts?

migurski:
I agree!

PR #2080, opened 2016-10-21
adding encoding=utf-8 for Belgium

thatdatabaseguy:


migurski:
screen shot 2016-10-20 at 11 06 43 pm

f

thatdatabaseguy:
So these jobs run on a cached CSV but the encoding error appears to get introduced when that file's being generated.

Downloaded the [original GMLfile](https://github.com/openaddresses/openaddresses/blob/master/sources/be/flanders.json#L4) and everything is UTF-8 encoded:
```
<agiv:STRAATNM>Aimé De Graevestraat</agiv:STRAATNM>
```

whereas in the cached CSV it's:
```
Aim? De Graevestraat
```

Is there a way to re-run from scratch with the new encoding?

migurski:
Yeah, we can regenerate the cache from source, or stop using the cache and go from source directly?

migurski:
Looks like you're on top of it in #2082!

PR #2086, opened 2016-10-24
Add us/mo/bates.json

slibby:


slibby:
@justinelliotmeyers here's the PR I created from that sample data (with a few typos in the commits for fun!)

iandees:
👏 

Thanks for putting this together @slibby and @justinelliotmeyers.

I just put the Dropbox file in our S3 bucket:
http://data.openaddresses.io.s3.amazonaws.com/cache/us/mo/bates.zip

iandees:
...and then updated the source to point to the new URL.

migurski:
Maybe time to add S3 cache upload to Machine so this is easier?

iandees:
I think pointing to existing docs for `awscli` or `s3cmd` or `Transmit.app` would be better than building it into machine.

migurski:
In machine it'd be possible to rely on Github org membership auth rather than distributing AWS keys & secrets, though.

iandees:
Huh, interesting. Let's move that convo to a new ticket.

PR #2094, opened 2016-10-29
Updated us/pa/wayne source and website.

davidchiles:


slibby:
Replaces broken source 👍

PR #2095, opened 2016-10-30
Fix broken sources

iandees:


migurski:
St. Louis, Cecil, Woodford, and Scott:

preview-st_louis

preview-cecil

preview-woodford

preview-scott

iandees:
Looks like the FTP cacher is broken maybe?

migurski:
Goodhue:

preview-goodhue

migurski:
`TypeError: 'str' does not support the buffer interface` sounds like a Python 2/3 unicode issue. Could we make a test case for this in machine?

migurski:
Also, Champaign:

preview-champaign

PR #2111, opened 2016-11-05
Add city of Flint, MI

### LINKED ISSUE #1988: Add addresses in Flint, MI
Status:
- [ ] **Research sources** and propose best source (and it's data license)
- [ ] **Review proposed source**
- [ ] **Approve source** based on review comments
- [ ] **Add** to OpenAddresses

_NOTE: Proposed source notes should include basic column mapping to OpenAddress data model._

iandees:
Fixes #1988.

### DIFF
--- a/sources/us/mi/city_of_flint.json
+++ b/sources/us/mi/city_of_flint.json
@@ -1,0 +1,33 @@
+{
+    "coverage": {
+        "geometry": {
+            "type": "Point",
+            "coordinates": [
+                -83.766,
+                43.012
+            ]
+        },
+        "US Census": {
+            "geoid": "2629000",
+            "name": "City of Flint",
+            "state": "Michigan"
+        },
+        "country": "us",
+        "state": "mi",
+        "city": "Flint"
+    },
+    "data": "http://services5.arcgis.com/lqqWNtSxx8Akj04A/ArcGIS/rest/services/Flint_Parcels_2016_v2/FeatureServer/0",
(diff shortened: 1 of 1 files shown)


migurski:
Testing https://github.com/openaddresses/machine/pull/420 with a larger source:

Re-run this please.

migurski:
Flint preview

nvkelso:
Would be nice to click on this preview image and link to a live map of same area to explore cause of density patterns (e.g. Mapzen's Bubble Wrap). 

For instance, I'm guessing downtown and industrial areas have bigger buildings but less address density here in Detroit.

migurski:
Oh yeah, that's a good idea. Right now, the complete dotmap on openaddresses.io is being hosted by Mapbox. Presumably it should be possible to display those vector tiles in Tangram?

nvkelso:
A link to that existing dot map at the same map extent as the data source preview would be most the way there :)

iandees:
The dotmap on the website is only updated once a week or so. Showing the points on a slippy map as part of the build process (before it gets Incorporated into the final output product) would require running the cached stuff through tippecanoe or something and serving the vector tiles from machine or something.

This would be a bunch more work but I think it'd be useful for when we want to show local governments what their data will look like as part of them posting their data to us.

migurski:
This fits in with a goal I’ve had, to show the status over each source over some kind of timeline.

Issue #1995, opened 2016-09-15
Add addresses in suburbs of Detroit, MI


nvkelso:
Status:
- [x] **Research sources** and propose best source (and it's data license)
  - [x] Wayne county
  - [x] Oakland county
  - [ ] Macomb county
- [ ] **Review proposed source**
- [ ] **Approve source** based on review comments
- [ ] **Add** to OpenAddresses

_NOTE: Proposed source notes should include basic column mapping to OpenAddress data model._

justinelliotmeyers:
Contacted Wayne County requesting data 10/2/2016

justinelliotmeyers:
Contacted Oakland County requesting data 10/2/2016

justinelliotmeyers:
Contacted Macomb County requesting data 10/2/2016

justinelliotmeyers:
Oakland County response:
Justin,
 
I received your request below, Oakland County GIS data can be purchased through the Oakland County One Stop Shop
One Stop Shop Business Center
Economic Development & Community Affairs Oakland County, Michigan
248.452.2058
 
Cost is $.01 per record and there are 491,901 records ($4,919.01).  OSS office charge of $30.  Total would be $4,949.01.
We also require a signed user agreement (see attached).
 
Let me know if you have any questions.
Thank you,
Janine
 
Janine Daily GISP, IT Business Analyst
Oakland County Department of Information Technology
1200 N. Telegraph Road | Building 49W | Pontiac, MI  48341
248.858.0123 direct  |  Fax 248.452.9128

justinelliotmeyers:
As per Oakland County, might be a better idea to contact individual cities to see if they will share without such a large cost.

nvkelso:
Yikes! Let's see if individual cities have better price options (hopefully
free).

justinelliotmeyers:
Macomb County Update from Dan Durkee (dan.durkee@macombgov.org )
1st email response: Hi Justin, we do not have a distribute-able point address data set to provide you.  thanks for your interest

I emailed him back asking how I could succesfully achieve the task without the counties data.

2nd response: Hi Justin, you could ask each community's tax assessor(addressing authority) in Macomb County for a list of addresses to geocode. 
We work primarily with address ranges(address approximation) in street center-lines.

dexterslu:
@justinelliotmeyers @nvkelso Oakland County rest service parcel points here:
[...]

PR #2117, opened 2016-11-07
Add Oakland, MI

iandees:
For #1995.


migurski:
Detroit is from before the launch of previews, but here is Oakland, MI: 

Oakland

nvkelso:
wow, those lakeside houses!

PR #2133, opened 2016-11-14
Add Washington statewide parcels

iandees:
The data is pretty dirty, so I'm only pulling in address number and street.

### DIFF
--- a/sources/us/wa/statewide.json
+++ b/sources/us/wa/statewide.json
@@ -1,0 +1,26 @@
+{
+    "coverage": {
+        "US Census": {
+            "geoid": "53",
+            "state": "Washington"
+        },
+        "country": "us",
+        "state": "wa"
+    },
+    "data": "http://geoservices.wa.gov/arcgis/rest/services/Parcels/ECY_Washington_State_Parcel_Database_2015/MapServer/0",
+    "type": "ESRI",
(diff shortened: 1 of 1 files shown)


iandees:
Re-run this please.

nvkelso:
Re-run this please.

sources/us/wa/statewide.json
```diff
@@ -0,0 +1,26 @@
+{
+    "coverage": {
+        "US Census": {
+            "geoid": "53",
+            "state": "Washington"
+        },
+        "country": "us",
+        "state": "wa"
+    },
+    "data": "http://geoservices.wa.gov/arcgis/rest/services/Parcels/ECY_Washington_State_Parcel_Database_2015/MapServer/0",
```
nvkelso: This might need to be:

`"data": "http://geoservices.wa.gov/arcgis/rest/services/Parcels/ECY_Washington_State_Parcel_Database_2016/MapServer/0",`

(note 2016 instead of 2015)

iandees:
I'm going to close this – looking closer and sampling more data it looks like there's too much of a range of data in the "SITUS_ADDR" column. All of this data came from individual counties, though, so if there's dirty data in here for a county there's probably cleaner data available directly from that county.

PR #2138, opened 2016-11-15
Skip some Missouri data sources at the request of local governments

iandees:
Jason Warzinik asked us to stop using their Esri REST endpoints for OpenAddresses. He asked to include contact information, which I will include below:

Boone County, Missouri:
```
Jason Warzinik, GISP, GIS Manager
Boone County
801 E. Walnut, Room 220
Columbia, MO 65201-4890
Phone: (573) 886-4325
E-mail: jwarzinik@boonecountymo.org
www: www.showmeboone.com
USNG 15S WD 5824 1181
```
 
City of Columbia, Missouri:
```
Matthew J. Gerike, PhD 
Geospatial Services Manager | City of Columbia, Missouri 
Matt.Gerike@CoMo.gov | 573.874.7304 
GIS Office On Call Phone: 573.825.8036
```

### DIFF
--- a/sources/us/mo/boone.json
+++ b/sources/us/mo/boone.json
@@ -11,12 +11,18 @@
     },
     "skip": true,
     "note": "Jason Warzinik asked us via email on 2016-11-16 to stop using their Esri REST endpoint",
-    "type": "ESRI",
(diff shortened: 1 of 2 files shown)


migurski:
Can I suggest that we cache these sources, instead of skipping them? It would have the same effect, but it might allow us to continue to experiment with new output transforms.

PR #2161, opened 2016-11-26
switch nz/countrywide to LINZ AIMS Street Address

andrewharvey:
As per https://data.linz.govt.nz/layer/3353-nz-street-address/ the current source for nz/countrywide in OA is being replaced.

> This dataset contains the core components of an address to support simple searching or use of this layer to provide context and is a replacement for the NZ Street Address (Electoral) dataset. This dataset has been sourced from LINZ’s Address Information Management System (AIMS), a centralised database for the management of national addresses, including for electoral purposes.
>
> The existing address and road datasets will be available for 6 months after AIMS go-live to allow users further time to transition.This version will be replacing NZ Street Address (Electoral) dataset.

This is my attempt to swap OA over to the new source data. The only issue I've noticed is this new data seems to be lacking the `region`.

/cc @s01ipsist hinted this change in https://github.com/openaddresses/openaddresses/pull/1699#issuecomment-222382268


scripts/nz/README.md
```diff
@@ -39,7 +39,7 @@ apt-get install docker.io.
 Getting Data
 ----
 
-1.  Start by [finding the Street Address (Electoral) dataset](https://data.linz.govt.nz/search/?q=street+address+electoral)
+1.  Browse to the [NZ Street Address dataset](https://data.linz.govt.nz/layer/3353-nz-street-address/)
     on [LINZ Data Service](https://data.linz.govt.nz/), and select “download” from
     the drop-down menu:
```
s01ipsist: We should update the screenshots to match the new functionality.
image

I also see a significantly different size (360MB).

s01ipsist: ah, estimate is just widely off. actual download is 160mb.

scripts/nz/lds-nz.sh
```diff
@@ -1,24 +1,24 @@
 #!/bin/bash
 
-# 1. Download the zip file from https://data.linz.govt.nz/layer/779-nz-street-address-electoral/
+# 1. Download the zip file from https://data.linz.govt.nz/layer/3353-nz-street-address/
 #    into this directory
```
s01ipsist: this prior instruction `into this directory` isn't correct. Let's change it to match the working flow. I think it should say `into the /work folder`

scripts/nz/README.md
andrewharvey: I've updated the screenshots. For me the export still says 168.2MB.

scripts/nz/lds-nz.sh
andrewharvey: good idea, fixed.

andrewharvey:
> Minor suggestions, 2 of the readme screenshots should be updated to make current.

I've updated the commit with these changes. Thanks!

> Loss of the region could be addressed in a separate PR. Regions should be https://en.wikipedia.org/wiki/ISO_3166-2:NZ and can be integrated from the New Zealand 2016 12 mile generalised (NZTM) (183MB) file on http://www.stats.govt.nz/browse_for_stats/Maps_and_geography/Geographic-areas/digital-boundary-files.aspx

Are you suggesting doing a point in polygon test to merge the region? Should that be done as a pre-processing stage into OA, or should OA users do this merge as a post processing? I suppose it'll be more convenient for data consumers if it was done transparently by OA.

s01ipsist:
I run a similar process to set the region when processing the data for http://www.addressable.co.nz/.  I put everything through a PostGIS db so setting the region is just 1 SQL update. I'm not sure how to do it without a db intermediary but it would be ideal for the data consumer to do it upfront. Definitely not required in this PR. 
Somebody with access to the data cache will need to upload the new data for the build to succeed. Looks good to me.

andrewharvey:
> I run a similar process to set the region when processing the data for http://www.addressable.co.nz/. I put everything through a PostGIS db so setting the region is just 1 SQL update. I'm not sure how to do it without a db intermediary but it would be ideal for the data consumer to do it upfront. Definitely not required in this PR.

Should be possible as an ogr2ogr step in the script with `-sql` and `-dialect sqlite`. I'll give it a go when I get a chance.

migurski:
Thanks for doing this, @andrewharvey!

andrewharvey:
> Thanks @andrewharvey for the PR. You are missing the region as recommended by @billgeo.

From what I can tell this new NZ Street Address (AIMS) data doesn't contain the `region` https://en.wikipedia.org/wiki/Regions_of_New_Zealand#List_of_regions. Hence it's omitted at the moment. @s01ipsist suggested to join it back from Stats NZ data, @billgeo @palmerj would that be your recomendation to get the region included?

andrewharvey:
@billgeo @palmerj Also I've omitted the `town_city` because I'm not sure which it maps to in OA  if any.

OA has three administrative levels for addresses per https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#attribute-tags

Tag | Note
-|-
city | Name of the City or Municipality in which the address falls
district | District/County/Sub-Region in which the address falls
region | State/Region/Province in which the address falls

If you could share your data/local knowledge on what makes sense to map to where, that would be much appreciated!

In Australia a typical address is something like:

    <UNIT> <STREET ADDRESS> <STREET NAME>
    <SUBURB> <STATE> <POSTCODE>

    Unit 101, 2b Main Street
    ULTIMO NSW 2007

`<CITY>` is really part of the address but we map `<SUBURB>` to the OA `city`and `<STATE>` to `region`.

andrewharvey:
Upstream news about the new LINZ dataset at http://mailchi.mp/e19f4f5e99ac/changes-to-address-and-roads-data-978653

nvkelso:
Is this the new “good” one to use?

https://data.linz.govt.nz/layer/3353-nz-street-address/data/

palmerj:
@nvkelso yes

andrewharvey:
> https://data.linz.govt.nz/layer/3353-nz-street-address/data/

@nvkelso That's the one this PR uses. I think it should be good to go, just some of the attributes from the source data aren't used.

billgeo:
@andrewharvey, in regards to how to map city, district, region I would suggest:

| tag | source |
|-----|------|
| city | 'suburb_locality' data in https://data.linz.govt.nz/layer/3353-nz-street-address/data/ |
| district | Stats NZ [Territorial Authority boundaries](https://koordinates.com/layer/4241-nz-territorial-authorities-2012-yearly-pattern/) are probably the closest to this | 
| region |	Stats NZ [Regional Council boundaries'](https://koordinates.com/layer/4240-nz-regional-councils-2012-yearly-pattern/) probably the closest to this |

Note: Stats NZ boundaries are administrative / political boundaries so not always the same as common usage in postal addresses. But the closest we have in a publicly available dataset I believe. 

And like you mentioned you'd have to link Stats and LINZ data together (spatially).

andrewharvey:
I think I've got the script updated based on the advice at https://github.com/openaddresses/openaddresses/pull/2161#issuecomment-302059696.

More testing is needed though before it's ready to merge.

In the meantime though #3010 was created and merged which has created merge conflicts here.

andrewharvey:
the ogr2ogr based spatial join implemented here is too slow to be practical and after trying many different options to speed it up, I can't get it to work in a reasonable amount of time. We might need to load it into psql.

andrewharvey:
Closing this PR as the new NZ Address Dataset was merged at https://github.com/openaddresses/openaddresses/pull/3010, and I can't get the spatial join here to work with ogr, and https://www.nzpost.co.nz/personal/sending-within-nz/how-to-address-mail/correct-address-formats-envelope-layouts says "Don’t use a province, region, district or territory name instead of (or in addition to) the town or city name.".

If there are issues with the current state, best to open a new issue for that.

PR #2162, opened 2016-11-26
add au/countrywide accuracy field

andrewharvey:
GNAF provides a few fields measuring geocode:

- `reliability` a measure of how accurate the geometry of the geocode is, did the geometry come from the site boundary, road, locality?
- `confidence` metric based on how many different address databases it appears in
and type.
- `geocode_type` what is this address/geometry representing. eg. is it the building centroid or the front door to the building? or is it the property parcel centroid, or does it come from offsetting from the road centerline at a distance along the road? etc.

OA only provides [*`accuracy`*](https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#accuracy).

Looking through which fields accually appear in the GNAF data and comparing the field definitions to the OA `accuracy` field, this is my suggested mapping. Comments welcome.

### DIFF
--- a/scripts/au/gnaf.sh
+++ b/scripts/au/gnaf.sh
@@ -61,7 +61,21 @@ SELECT
     latitude AS lat,
     locality_name AS city,
     locality_postcode AS postcode,
-    state AS region
+    state AS region,
+
+    (
+        CASE geocode_type WHEN 'BUILDING CENTROID' THEN 1 -- rooftop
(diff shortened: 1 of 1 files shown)


migurski:
These changes are now reflected in http://data.openaddresses.io/cache/au-nov2016.zip.

andrewharvey:
> These changes are now reflected in http://data.openaddresses.io/cache/au-nov2016.zip.

Awesome. thanks.

PR #2166, opened 2016-11-28
Added us/ca/tehama

davidchiles:
This source has a lot of inconsistent data. Parcels are missing some  or all of the fields. I left out the `SITCITST` field because sometimes it contains city information or city + state or some other information like Granny Unit.


migurski:
Looks good!

PR #2172, opened 2016-11-29
added docs for prefix_number and prefix_street

trescube:


CONTRIBUTING.md
```diff
@@ -221,6 +253,8 @@ _Example_
 }
 ```
 
+The source data should be examined to determine if the shorthand methods prefix_number and prefix_street could be used instead of regexp.
```
iandees: Can you add \`'s around the function names for clarity.

CONTRIBUTING.md
```diff
@@ -147,7 +147,9 @@ This list gives a brief summary of what each function does. Examples can be foun
 
 Function | Note
 -------- | -----
-`regexp` | Allow regex find and/or replace on a given field. Useful to extract house number/street/city/region etc when the source has them in a single field
+`prefix_number` | Allow number to be extracted from a single field using an internal convenience regex.
```
iandees: A minor nit: do you think it would make sense to remove reference to it being a regular expression behind the scenes? This way we can swap out the backend implementation with something non-regex-ey later. Or maybe it doesn't matter...

Issue #2179, opened 2016-11-30
Texas OA Data


justinelliotmeyers:
![texas_oa_yeeeehawwww](https://cloud.githubusercontent.com/assets/5324316/20774587/f810ac68-b724-11e6-9f0b-baf46cae0461.jpg)

After filing an official PIA, this is what I was sent.  Cheers!

Mr. Meyers,
 
The information responsive to your request can be downloaded via the following link:  https://geo-comm.sharefile.com/d-s72fbc0849fe49de9.  The Road Centerline and Site Structure Address Points (SSAP) layers include the fields responsive to your request.  SSAP includes fields for latitude and longitude, however, pursuant to the CSEC NG9-1-1 GIS Data Standard, latitude and longitude are currently optional fields so not all SSAP may have x/y information.  Please let me know if you need further assistance.
 
Thank you.
 
Patrick Tyler
Public Information Coordinator
Commission on State Emergency Communications (CSEC)

iandees:
Wahoo! This is great, thanks for chasing this, @justinelliotmeyers.

I uploaded the data here: http://data.openaddresses.io.s3.amazonaws.com/cache/us/tx/20161129_CSEC_SSAP_RCL.zip

justinelliotmeyers:
**They plan on having complete coverage by February 2017 - I will put in my calendar to do another formal PIA request so we can fill in the missing counties.

PR #2180, opened 2016-11-30
Add a source for Texas statewide

iandees:
Fixes https://github.com/openaddresses/openaddresses/issues/2179


migurski:
Nice. I’m going to add a footprint polygon for this, based on a list of included counties.

migurski:
Does this look right? 

screen shot 2016-11-30 at 8 42 54 pm

migurski:
Also, we'll need a website property here to be able to track this later, since we're hosting a cached version ourselves.

nvkelso:
@migurski I've noted **7 counties to remove** from the Texas footprint in the GIF below. The blue ? turns out to be  a military base that spans multiple counties so that's why there's a weird hole there (no change in footprint proposed there). The projections aren't quite the same between the images, but close enough.

texas_footprint

migurski:
Thanks, here's a new one:

screen shot 2016-12-01 at 5 16 41 pm

PR #2193, opened 2016-12-06
added municipality of Kalmar, Sweden

### LINKED ISSUE #2188: Kalmar, Sweden
kalmar_se

http://gis.kalmar.se/arcgisserver/rest/services/Bakgrundskartan/Bakgrundskarta_500k/MapServer/3

trescube:
Swedish municipalities seem to break down further into localities (at least according to wikipedia) so I wasn't sure how much detail to add to `coverage`.  Let me know if more is required.

Fixes #2188

### DIFF
--- a/sources/se/municipality_of_kalmar.json
+++ b/sources/se/municipality_of_kalmar.json
@@ -1,0 +1,14 @@
+{
+    "coverage": {
+        "country": "se",
+        "city": "Kalmar"
+    },
+    "data": "http://gis.kalmar.se/arcgisserver/rest/services/Bakgrundskartan/Bakgrundskarta_500k/MapServer/3",
+    "type": "ESRI",
+    "conform": {
+        "type": "geojson",
+        "number": "ADRPLATS",
+        "street": "ADROMRADE",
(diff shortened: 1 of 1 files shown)


sergiyprotsiv:
Hi, Swedish localities (tatörter) are not really an administrative division - they are defined every five years based on the density of built-up areas, and thus of course thy often are spread over many municipalities. their main purpose is statistical.

So the municipalities (kommuner) are definitely the right choice for the coverage (they are similar to US counties in scope). They are also the ones responsible for maintaining addresses, so hopefully we will see more of them publishing the data online.

Last week I talked to Lantmäteriet (Swedish Cadastre) regarding their plans of opening up the address data nationwide (they had it with a question mark in some of their presentations). Unfortunately, they said that they have not yet secured any funding for that yet, so this remains a plan without any specific date.

So, for now the best bet in Sweden will be the individual municipalities. Good news is that I have found open address data for Stockholm, will try to publish it later today.

nvkelso:
Thanks for the background, this is super helpful :)

PR #2217, opened 2016-12-07
added city of Kharkiv, Ukraine

### LINKED ISSUE #2176: Kharkiv Ukraine
http://cdr.citynet.kharkov.ua/arcgis/rest/services/gis_ort_stat_general/MapServer/1

I tried to quick pull some down and verify - looks like points plot, but didn't want to come in WGS84 - just a fyi

trescube:
Fixes #2176

### DIFF
--- a/sources/ua/63/city_of_kharkiv.json
+++ b/sources/ua/63/city_of_kharkiv.json
@@ -1,0 +1,24 @@
+{
+    "coverage": {
+        "country": "UA",
+        "state": "UA-63",
+        "city": "Kharkiv"
+    },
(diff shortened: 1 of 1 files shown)


sergiyprotsiv:
Hi,
In my test last week there was an issue with Kharkiv. Basically the coordinate system it seems to be following is wrong and in my tests Kharkiv ended up deep inside Kenya. Could you check if this happens for you too? I am pretty sure it is using one of the Pulkovo projections, but the offsets from origin are clearly wrong.
Sergiy

trescube:
Records still appear to be landing in Kenya.  We'll investigate.

migurski:
Yeah that preview map should be a tip-off.

trescube:
Oh yeah, good point.  I see a lot of dots and assume that it's correct.

Issue #2221, opened 2016-12-08
us/ga/forsyth


davidchiles:
http://gisp2.forsythco.com/arcgis/rest/services/TaxAssessor/CAMA_ParcelMap/MapServer/0

davidchiles:
@trescube 😕 This is for Forsyth County. The pull request was for Douglas County.

trescube:
D'OH!  Sorry about that.

trescube:
Some might say that I had Georgia on my mind :rimshot:

justinelliotmeyers:
https://geo-forsythcoga.opendata.arcgis.com/datasets/siteaddress

PR #2222, opened 2016-12-08
Added us/ga/douglas

davidchiles:


trescube:
Thanks for adding this!  I looked thru the data and saw that some of the street fields have embedded units, using the terms `UNIT`, `APT`, `BLDG`, and `STE`.  Modifications to the regexes from https://github.com/openaddresses/openaddresses/blob/master/sources/us/nd/stutsman.json should do it.

davidchiles:
@trescube thanks

trescube:
Fixes #2221

PR #2239, opened 2016-12-11
Stockholm, Sweden

sergiyprotsiv:
Adding Stockholm municipality, which is about 10% of Sweden's population. 

It has an API that one could query to get all streets in the city, and then all addresses for the street. I am also attaching the script file. 

Getting the API key is easily done here: http://openstreetgs.stockholm.se/Home/Key. Put an email in the box on the left and preferably select the WGS 84 coordinate system.

I have not found the reference to a specific license, but I am reasonably sure from the text on the site that it is not share-alike and at most requires attribution.


sergiyprotsiv:
Fixed the bad header (it had leading spaces in column names). Ready to commit.

migurski:
Thanks!

PR #2240, opened 2016-12-11
Aachen, Germany

sergiyprotsiv:
Should be a simple WFS dump to GML...

### DIFF
--- a/sources/de/nw/city_of_aachen.json
+++ b/sources/de/nw/city_of_aachen.json
@@ -1,0 +1,23 @@
+{
+	"coverage": {
+		"country": "de",
+		"state": "nw",
+		"city": "Aachen" }
+	},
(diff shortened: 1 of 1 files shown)


sources/de/nw/city_of_aachen.json
```diff
@@ -0,0 +1,23 @@
+{
+	"coverage": {
+		"country": "de",
+		"state": "nw",
+		"city": "Aachen" }
```
migurski: Looks like a stray closing JSON curly brace here.

migurski:
Re-run this please

sergiyprotsiv:
Not sure where the error is: the machine is getting an uncompressed gml file (without extension), but tries different extensions, says it has decompressed it, and then fails to find the gml inside...

migurski:
I think it's a mis-map between the mime-type and an assumed filename extension. Definitely a bug in machine code.

PR #2249, opened 2016-12-14
Add docs for remove_prefix and remove_postfix functions

trescube:


migurski:
Ack, hold up — it’s not necessarily safe to merge this until we have actually deployed the changes from  https://github.com/openaddresses/machine/pull/495. They're not yet in a release. It's no big deal for now, but we have to do these in serial so that our docs are accurate. Right now, trying to use these functions will fail.

trescube:
D'oh!  Sorry, I even remembered to hold off on #2195 until the machine was updated but got overly eager on this one.

PR #2253, opened 2016-12-15
Add json schema

trescube:
@migurski and @iandees (and others), could you take a look at this PR and give me some feedback?  

# Current State of the Schema

The schema is 99% complete based on the tests in sources.js and docs in CONTRIBUTING.md.  The only thing that I can say doesn't work right now is failing when any of `csvsplit`, `headers`, and `skiplines` are specified when `type` is not `csv`.  v4 of json-schema doesn't appear to support conditional logic where a field can only be present when another field is a certain value.  There are ways to do it for simple cases by defining separate types that are very similar but in the context of the `conform` object, it would make the schema very cumbersome and convoluted.  In this case it may not be a big deal since those fields will just be ignored when `type` is something other than `csv`.  The v5 draft propsal does include support for conditional schema logic but it has to be enabled specifically in [ajv](https://www.npmjs.com/package/ajv) and I wasn't sure if it would be worth it for something relatively minor.  

The new schema imposes a lot more restrictions on sources listed in [CONTRIBUTING.md](https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md) than are tested in sources.js.  

# Tests

There are 2 types of tests:

* validate every source against the schema (sources_validator.js)
* validate the schema itself using mock sources (schema_validation.js)

## Source Validation

Switching to testing sources via schema validation over individual tape tests saves a lot of build time.  The former takes ~3 seconds whereas the latter takes almost a minute.  

All sources pass schema validation, though the `format: uri` for the `data` property has been disabled until the issue with decoding URLs mentioned in #2254 is fixed.  

## Schema Validation

I've also started a test suite for the schema itself so that, for example, a source missing a `type` property should produce an error.  There is only a smattering of tests in there right now as proof of concept before I got feedback from the team.  Since there are a relative handful of things that can go wrong (missing properties, incorrect types, value errors, etc), I added some convenience functions for these scenarios.  Due to testing one condition but potentially causing multiple errors, rather than fail based on the first element of the errors array, it looks thru all for the expected failure condition.  

Other than the schema itself, I'd appreciate feedback on how these tests are structured before continuing.  

# Infrastructure

## node 

This PR also upgrades from node 0.10 to 4/6.  0.10 is well past end-of-life and a new package introduced for testing needs at least node 4.  

## tape/tap

`npm test` used to run `tape test/*.js | tap-min` which only works because there was a single test file in `test`.  Having more that one test .js file causes the command to fail due to process return codes.  The solution is to introduce the `multi-tape` package which can handle a globbed list of tests by running each in a separate node process.  A side-effect of switching to `multi-tape` is that it produces output incompatible with `tap-min`, like:

```
#
# test/schema_validation.js
#
TAP version 13
# bare minimum source should pass
ok 1 type http should pass
ok 2 type ftp should pass
ok 3 type ESRI should pass

```

The first 3 lines are comments and don't appear to be valid either due to format or prepending `TAP version 13` (`tape` itself doesn't do this).  `tap-min` doesn't deal well with this output format and causes all the tests to pass but the overall run to fail.  The switch to `tap-dot` has fixed this so it appears to be more forgiving of where comments are.  Both are functionally equivalent, only the latter outputs a green `.` to indicate success and a red `x` is a failure.  It runs in travis just fine.


migurski:
This is exciting!

schema/source_schema.json
```diff
@@ -76,9 +76,84 @@
...
+        },
+        "county": {
+          "type": "string"
+        },
+        "city": {
+          "type": "string"
+        },
+        "geometry": {
+          "description": "from https://github.com/fge/sample-json-schemas/blob/master/geojson/geometry.json",
+          "oneOf": [
```
iandees: Instead of copying bits of the GeoJSON schema into our schema you can reference it like here: https://github.com/osmlab/editor-layer-index/blob/gh-pages/schema.json#L14

trescube: Ah, good point!

trescube:
@migurski and @iandees (and others), could you take a look at this PR and give me some feedback?  

# Current State of the Schema

The schema is 99% complete based on the tests in sources.js and docs in CONTRIBUTING.md.  The only thing that I can say doesn't work right now is failing when any of `csvsplit`, `headers`, and `skiplines` are specified when `type` is `csv`.  v4 of json-schema doesn't appear to support conditional logic where a field can only be present when another field is a certain value.  There are ways to do it for simple cases by defining separate types that are very similar but in the context of the `conform` object, it would make the schema very cumbersome and convoluted.  In this case it may not be a big deal since those fields will just be ignored when `type` is something other than `csv`.  The v5 draft propsal does include support for conditional schema logic but it has to be enabled specifically in [ajv](https://www.npmjs.com/package/ajv) and I wasn't sure if it would be worth it for something relatively minor.  

The new schema imposes a lot more restrictions on sources listed in [CONTRIBUTING.md](https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md) than is tested in sources.js.  

# Tests

There are 2 types of tests:

* validate every source against the schema (sources_validator.js)
* validate the schema itself using mock sources (schema_validation.js)

## Source Validation

Switching to testing sources via schema validation over individual tape tests saves a lot of build time.  The former takes ~3 seconds whereas the latter takes almost a minute.  

All sources pass schema validation, though the `format: uri` for the `data` property has been disabled until the issue with decoding URLs mentioned in #2254 is fixed.  

## Schema Validation

I've also started a test suite for the schema itself so that, for example, a source missing a `type` property should produce an error.  There is only a smattering of tests in there right now as proof of concept before I got feedback from the team.  Since there are a relative handful of things that can go wrong (missing properties, incorrect types, value errors, etc), I added some convenience functions for these scenarios.  Due to testing one condition but potentially causing multiple errors, rather than fail based on the first element of the errors array, it looks thru all for the expected failure condition.  

Other than the schema itself, I'd appreciate feedback on how these tests are structured before continuing.  

# Infrastructure

## node 

This PR also upgrades from node 0.10 to 4/6.  0.10 is well past end-of-life and a new package introduced for testing needs at least node 4.  

## tape/tap

`npm test` used to run `tape test/*.js | tap-min` which only works because there was a single test file in `test`.  Having more that one test .js file causes the command to fail due to process return codes.  The solution is to introduce the `multi-tape` package which can handle a globbed list of tests by running each in a separate node process.  A side-effect of switching to `multi-tape` is that it produces output incompatible with `tap-min`, like:

```
#
# test/schema_validation.js
#
TAP version 13
# bare minimum source should pass
ok 1 type http should pass
ok 2 type ftp should pass
ok 3 type ESRI should pass

```

The first 3 lines are comments and don't appear to be valid either due to format or prepending `TAP version 13` (`tape` itself doesn't do this).  `tap-min` doesn't deal well with this output format and causes all the tests to pass but the overall run to fail.  The switch to `tap-dot` has fixed this so it appears to be more forgiving of where comments are.  Both are functionally equivalent, only the latter outputs a green `.` to indicate success and a red `x` is a failure.  It runs in travis just fine.

migurski:
Is it feasible to express the JSON schema in a format other than JSON, ideally one that supports comments? I'm a bit troubled by how opaque it looks.

trescube:
Not with json-schema itself.  A `description` field can be added virtually anywhere to describe what something does.

migurski:
This does seem a bit cleaner than the JS code.

trescube:
Oddly, when a field can only have one value, there's no way in v4 to say "it can only have this value", the value has to be wrapped in a single-element `enum`.  v5 does allow for `const`, however.

raizlin:
[Sc2ReplayStats.zip](https://github.com/openaddresses/openaddresses/files/659346/Sc2ReplayStats.zip)



- @sesam

raizlin:
> _``_

trescube:
@raizlin what is this file?  It doesn't appear to have anything to do with OA.

migurski:
Account created today; I'm going to guess it’s for phishing. Deleted.

trescube:
Interestingly, the account creator fixed 3 sources in #2261.

migurski:
Huh, interesting! Well, maybe they’ll explain the file.

trescube:
Ah, no they didn't, those were @iandees commits, @raizlin just added an empty comment.

schema/source_schema.json
```diff
@@ -0,0 +1,416 @@
...
+        "accuracy": {
+          "description": "https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#accuracy",
+          "type": "integer",
+          "minimum": 0,
+          "maximum": 5,
+          "exclusiveMinimum": true
+        },
+        "srs": {
+          "type": "string",
+          "regex": "EPSG:[0-9]{4}"
```
iandees: EPSG codes can be longer or shorter than 4 digits

schema/source_schema.json
```diff
@@ -0,0 +1,416 @@
...
+            { "$ref": "#/definitions/base_conform_field_value" },
+            { "$ref": "#/definitions/function_regexp" },
+            { "$ref": "#/definitions/function_remove_prefix" },
+            { "$ref": "#/definitions/function_remove_postfix" },
+            { "$ref": "#/definitions/function_join" },
+            { "$ref": "#/definitions/function_format" }
+          ]
+        },
+        "district": {
+          "description": "doesn't include prefixed_number or postfixed_street",
```
iandees: These `description`s should be more useful information about the field, not descriptions about the list in `oneOf`.

iandees: e.g. "The state/province for the address row"

iandees:
This looks good. Just a couple small changes and I think we're ready to merge.

iandees:
Thanks for making this change, @trescube!

migurski:
💯

PR #2254, opened 2016-12-15
Update uris for formatting

trescube:
Spaces in URLs is causing URI format validation errors in #2253


iandees:
We need to update machine so that it decodes these URLs before sending them to the FTP client code we use.

migurski:
I can do this.

PR #2260, opened 2016-12-17
WIP: updated Czech Republic countrywide

### LINKED ISSUE #2238: New Nation-wide Czech Address Data!!
cz_addr_12092016

Current data below is green.  Red is new data.  (I like mapping during the holidays... X-Mas colors...).
cz_addr_12092016_compare_to_current

For the past week I have been working with our current Czech Address data.  Looks amazing!  I did notice a few issues with the data however.  
1)there is a shift compared to current data from http://services.cuzk.cz/ (Adresy )
2)the locations seem to be triplicated (essentially each address has three identical points)

I have done the following steps to create a new Czech Dataset:
1)Downloaded all 6,258 files here: http://services.cuzk.cz/gml/inspire/ad/epsg-4258/

2)Extract

3)convert all xml files to shapefile in FME (*UTF-8 using fanout method); **554782.xml errors out because the 2GB dbf limit - so this was manually converted in QGIS

4)merge files together in one feature class in ESRI

5)reproject to WGS84

6)add X Y

7)export out X, Y, and alternativ fields (alternativ has address information)

I need to know if I should do a spatial join to current administrative boundaries?  Please lt me know.

trescube:
Fixes #2238 

This will probably take some regex finessing.

### DIFF
--- a/sources/cz/countrywide.json
+++ b/sources/cz/countrywide.json
@@ -13,6 +13,7 @@
     "data": "https://www.dropbox.com/s/fhopgbg4vkyoobr/czech_addresses_wgs84_12092016_MASTER.zip?dl=1",
     "note": {
         "address formats": "http://www.upu.int/fileadmin/documentsFiles/activities/addressingUnit/czeEn.pdf",
+        "description": "according to format, there are cities w/o streets, just house numbers",
         "examples": {
             "street+house+neighbourhood+postcode+city": [
                 "Jiřská 9/10, Hradčany, 11900 Praha 1",
--- a/sources/cz/countrywide.json
+++ b/sources/cz/countrywide.json
@@ -17,14 +17,22 @@
         "number": {
             "function": "regexp",
             "field": "alternativ",
-            "pattern": "^(.+?) ([0-9]+(?:/[0-9]+[a-z]?)?), .*$",
-            "replace": "$2"
+            "pattern": ".*?((?:č\\.?(p|ev)\\.?)?\\s*\\d+(?:\/\\d+[a-z]?)?)"
         },
         "street": {
             "function": "regexp",
             "field": "alternativ",
-            "pattern": "^(.+?) ([0-9]+(?:/[0-9]+[a-z]?)?), .*$",
-            "replace": "$1"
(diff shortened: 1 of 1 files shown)


migurski:
This is good to merge, yes?

trescube:
I need to check the resulting data.

iandees:
@trescube still checking?

trescube:
Just got back to it, regexs need some tweaking, working on it now.

thatdatabaseguy:
For the way the regex function behaves, I've found it's usually better to add only one capture group per output field and no "replace" as this will simply exclude records that don't match the regex. Otherwise machine will insert ```group(0)``` i.e. the entire source field.

trescube:
This source is pretty tricky due to the variations in the data formats.  For example, sometimes the format is street+house+city+postcode as in `Jiřská 9/10, Hradčany, 11900 Praha 1`.  Other times, the format is house+postcode+street as in `č.p. 111, 27042 Skryje` but this could also be interpreted as house+postcode+city (as seen in [google maps](https://www.google.com/maps/place/49%C2%B057'46.7%22N+13%C2%B045'53.1%22E/@49.9631823,13.7643259,19z/data=!4m5!3m4!1s0x0:0x0!8m2!3d49.962963!4d13.764754))

trescube:
Ah, according to http://www.upu.int/fileadmin/documentsFiles/activities/addressingUnit/czeEn.pdf, addresses like `č.p. 111, 27042 Skryje` are in towns that don't employ street names, so it's parsed as house + postcode + city.

trescube:
Unless @thatdatabaseguy can work some regex magic, it would probably behoove us to split the data into multiple files with multiple sources for that various formats.  Hopefully it's only 2.  Thoughts?

trescube:
Turns out `Jiřská 9/10, Hradčany, 11900 Praha 1` is parsed as street+house+neighbourhood+postcode+city (Praha is Czech for Prague).

trescube:
In light of that last one, it's probably not necessary to split.

thatdatabaseguy:
Maybe try adding the initial ```^``` to those regexes.

trescube:
Thanks to @thatdatabaseguy for regex help!

PR #2267, opened 2016-12-19
Added us/il/mclean

davidchiles:


thatdatabaseguy:
Think this is McLean County, Kentucky rather than Illinois.

davidchiles:
@thatdatabaseguy I made a mistake mistake. Somehow I reused the link from Warren County Ky.

PR #2268, opened 2016-12-19
Added us/mi/st_clair

davidchiles:


nvkelso:
Exurbs north of Detroit 👍

PR #2275, opened 2016-12-20
added city of Wuppertal

trescube:


iandees:
Based on [this StackOverflow answer](http://gis.stackexchange.com/questions/61823/ogr-projection-transformation-error) the error that GDAL is giving here:

> Failure 6: No translation for Mercator_Auxiliary_Sphere to PROJ.4 format is known

Is because the exporter they're using (probably FME or similar) is writing out an invalid .prj file. We could probably fix this by allowing the `srs` attribute on `conform` to override the projection of the source, even if its specified in a .prj. This should probably be a `machine` bug/feature.

iandees:
See also https://github.com/openaddresses/machine/issues/423.

PR #2287, opened 2016-12-22
added Asotin County, WA

trescube:


migurski:
Hm, seeing zero points in this one, but a meaningful data sample. I'll check whether machine code is missing something important.

trescube:
Talked to @iandees about it, apparently it's a difference between geojson and esrijson

PR #2296, opened 2016-12-28
reverted to regexps for number+street

trescube:
Pull `unit` from single value field, added note warning others about switching to component fields.


iandees:
Looks like some regex problems here still?

PR #2301, opened 2016-12-29
WIP: added Jamaica countrywide

### LINKED ISSUE #2256: Jamaica Parcel data
http://gaia.globejamaica.com/NLA_Flex/rest/services/Maps/LandBase_NLA/MapServer/37

Not sure if this will work or is of any value, but wanted to share

trescube:
Fixes #2256


iandees:
Did you esri2geojson this yourself because it took so long?

I wonder if we could change how `machine` runs its periodic updates so that sources that take a long time can be processed more slowly 🤔  without manual intervention like this.

trescube:
You can probably answer this question better than I can due to your familiarity with the problem but while trying to tackle [Richmond, ND](https://github.com/openaddresses/openaddresses/tree/master/scripts/us/nd) I tried introducing sleeping between requests but didn't have any luck even with 5 minutes between.

trescube:
Though the problem may have been that source and not true of slow sources in general.

thatdatabaseguy:
Maybe try the field name lowercased? Appears all the field names except LAT and LON get lowercased in machine: https://github.com/openaddresses/machine/blob/master/openaddr/conform.py#L934

trescube:
Weird! I wrote that function, i wonder what i was thinking... Anyway, thanks for taking a look!

thatdatabaseguy:
No worries, wasn't your function in any case. It has to do with case smashing the configs in machine. Making a pull request on that repo.

trescube:
Oh, i see now, sounds like an end-to-end test would've caught this but units weren't enough.

trescube:
Ah, the source that drove the function implementation has lowercased fields already.

migurski:
That is some bold regexping.

thatdatabaseguy:
City names looking good, after said regexping (oy to the vey).

In cases where the city is something like "Kingston 5" the "5" is technically a postcode (discussion here: https://github.com/openvenues/libpostal/issues/113), but it's also fine to call it city for most purposes.

@trescube ok to merge on your end?

trescube:
There's a depressingly small amount of good data in there:  only 19,146 out of 188,794 have a house number.

Issue #2305, opened 2017-01-01
Document support for layer names in conform


migurski:
- [x] Wait for https://github.com/openaddresses/machine/pull/510 to be merged
- [x] Wait for changes to be released to all workers
- [x] Describe use of new `layer` conform attribute
- [ ] Update New Jersey statewide source, as noted in https://github.com/openaddresses/machine/issues/500#issuecomment-269878745

iandees:
You read my mind. I was just thinking about how to explain this in the docs...

migurski:
Want to take a crack at this one? I'm just about ready to merge and release.

 _…waits for CI tests after another Python 2 strings/bytes bug._

migurski:
Released. Changes will be on all workers after [the current batch set](https://results.openaddresses.io/sets/141476/) is complete. Should be tomorrow sometime.

PR #2306, opened 2017-01-01
Add docs for the `layer` tag in `conform`

iandees:
For https://github.com/openaddresses/openaddresses/issues/2305


CONTRIBUTING.md
```diff
@@ -102,6 +102,7 @@ They are called [Processing Tags](#processing-tags) and [Attribute Tags](#attrib
 ---------------- | --- | ----
 `type`           | Yes | The type properties stores the format. It can currently be one of `gdb`, `shapefile`, `shapefile-polygon`, `csv`, `geojson`, or `xml` (for GML).
 `srs`            |     | Allows one to set a custom source srs. Currently only supported by `type:shapefile`, `type:shapefile-polygon`, and `type:csv`. Should be in the format of `EPSG:####` and can be any code supported by `ogr2ogr`. Modern shapefiles typically store their project in a `.prj` file. If this file exists, this tag should be omitted.
+`layer`          |     | The `gdb` source type allows multiple layers of geodata in a single input file. Use the `layer` tag to specify which of those layers to use. It can either be the string name of the layer or an integer index of the layer.
```
migurski: This will work for any OGR source that supports layers — is there another we can use an example here, to clarify that it's not just GDB? Or will GDB practically be the only one that this works for?

iandees: Yea, I'm pretty sure that gdb is the only example of a source type we support that has support for multiple layers.

PR #2308, opened 2017-01-02
Liechtenstein countrywide

sergiyprotsiv:
Another full-country coverage:) Again in Dropbox. Please add me to the Openaddresses organization, will upload all the sources to S3.

This one comes from the official website: http://www.llv.li/#/11694/internet-kartendienstgeowebservices However, the Addresses layer there strangely contains only geometry and no attributes. So I am getting the data from another WFS server of theirs, the GDAL command is:

`ogr2ogr -f "ESRI Shapefile" -sql 'select * from "ms:HADR"' li_countrywide.shp WFS:"http://geodaten.llv.li//mapserver/geoportal"`

Was unable to find the license, but this is all freely accessible.

### DIFF
--- a/sources/li/countrywide.json
+++ b/sources/li/countrywide.json
@@ -6,7 +6,7 @@
         },
         "country": "li"
     },
-    "data": "https://www.dropbox.com/s/t77mtmevrdddd4q/li_countrywide.zip?dl=1",
+    "data": "https://s3.amazonaws.com/data.openaddresses.io/cache/uploads/sergiyprotsiv/3b8bdb/li_countrywide.zip",
     "website": "http://geodaten.llv.li/geoportal/gebaeudeidentifikator.html",
     "type": "http",
     "compression": "zip",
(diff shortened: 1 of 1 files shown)


migurski:
I’ve invited you to the Github team here; when you accept you should be able to upload to the cache from https://results.openaddresses.io/upload-cache. Thanks for this contribution!

migurski:
Also, the preview looks like the addresses are spatially offset to the northeast by 50-100m, so we may want to check the projection and override it if necessary.

sergiyprotsiv:
The file SRID is 2056, maybe the importer does not pick it up from the shapefile? It looks okay locally
screen shot 2017-01-02 at 09 58 47

migurski:
It must not be be quite EPSG:2056 in the .prj definition, because I get two different interpretations when I convert:

screen shot 2017-01-02 at 10 09 43 am

(green dots are `ogr2ogr -t_srs EPSG:4326 green.shp li_countrywide.shp`, gray dots are `ogr2ogr -t_srs EPSG:4326 -s_srs EPSG:2056 gray.shp li_countrywide.shp`)

migurski:
The gray ones are correct, so I think this is fixable by adding a `"srs": "EPSG:2056"` to the source.

migurski:
Looks better, great find!

PR #2310, opened 2017-01-03
Stockholm postal codes

thatdatabaseguy:
This source has a postal code field as a five digit number, so just needed to be formatted according to https://en.wikipedia.org/wiki/Postal_codes_in_Sweden.


migurski:
👍

PR #2312, opened 2017-01-04
Basel-Stadt, Switzerland

sergiyprotsiv:
Here are the addresses in the Swiss canton of Basel-Stadt. As with many sources in Switzerland, the canton has a "geoshop" where one can order the data. It then comes to the mailbox as a link to download the file (which I then upload to S3).

### DIFF
--- a/sources/ch/basel-stadt.json
+++ b/sources/ch/basel-stadt.json
@@ -14,7 +14,7 @@
     },
     "conform": {
         "type": "shapefile",
-        "encoding": "latin1",
+        "encoding": "ISO-8859-1",
         "file": "DM_Datenmarkt/DatenmarktAdressen.shp",
         "number": "NAUSNR",
         "street": "STR_NAME",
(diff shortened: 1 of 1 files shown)


sergiyprotsiv:
Re-run this please.

sergiyprotsiv:
Re-run this please.

sergiyprotsiv:
Something is wrong either with file encoding or with how machine is treating it: even after providing "latin1" as the encoding in the conform it tries to parse it as utf-8.

migurski:
I think this is related to https://github.com/openaddresses/machine/issues/512. I have a fix completed, but it might be a bit fiddly to deploy so I want to make sure I do it when I have 20-30 minutes set aside to get the sequencing right.

migurski:
Re-run this please.

migurski:
Okay that worked, though it does look a bit offset. Setting an SRS might help.

sources/ch/basel-stadt.json
```diff
@@ -19,6 +19,7 @@
         "number": "NAUSNR",
```
thatdatabaseguy: Think the field name is HAUSNR

PR #2313, opened 2017-01-04
Basel-Land, Switzerland

sergiyprotsiv:
A similar data source, only here any hope of automation is doomed by the presence of a Java site. Again, ordered the addresses, uploaded the received file to S3.

### DIFF
--- a/sources/ch/basel-land.json
+++ b/sources/ch/basel-land.json
@@ -1,0 +1,27 @@
+{
+    "coverage": {
+        "country": "ch",
+        "city": "Basel-Land"
+    },
+    "website": "https://www.geo.bl.ch/servlet/redirector/standard/geoshop/BLGeoShop.igs?user=public&password=public",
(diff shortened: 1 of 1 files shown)


migurski:
Re-run this please.

migurski:
Data is good, but it doesn’t want to do a preview image. Possibly one or more lat/lons are null or invalid; I will look.

sergiyprotsiv:
Thanks a lot @simonpoole for the comments on licensing, having the input from the local is appreciated. For Basel-Land the restricted terms appear to be related to the data under "Zugangsberechtigungsstufe B", the definition of which is not obvious to find. I saw this link: https://www.geo.bl.ch/index.php?id=118 for example and interpreted that "Zugangsberechtigungsstufe B" refers to the data that can only be accessed by cantonal and other Swiss officials and not the general public. Please correct me if I am wrong.
Most sources do require attribution though, this a definite oversight on my part.

PR #2314, opened 2017-01-04
Schaffhausen, Switzerland

sergiyprotsiv:
Another Geoshop order (here: http://gis.sh.ch/GIS_SH/BM3.asp -> Ausgabe -> GeoShop). They seem to have WMS servers as well at http://www.sh.ch/Geodienste.3480.0.html, but I could not connect to them via WFS.

### DIFF
--- a/sources/ch/schaffhausen.json
+++ b/sources/ch/schaffhausen.json
@@ -17,6 +17,7 @@
         "number": "HAUSNUMMER",
         "street": "STRASSE",
         "city": "ORTSCHAFT",
-        "postcode": "PLZ"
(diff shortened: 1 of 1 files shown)


sergiyprotsiv:
Same encoding issue here, I think.

migurski:
Re-run this, please.

migurski:
Hm, looks offset just like https://github.com/openaddresses/openaddresses/pull/2312#issuecomment-270543851. Hope this is a source problem, and not a bug introduced in GDAL 2.1.0.

PR #2311, opened 2017-01-03
Distrito Federal, Brazil

astoff:

Issue #2303, opened 2016-12-30
Brazil census data


astoff:
Census data could be used to generate a country-wide address list in Brazil, but it would require a dedicated script to parse and process. I will describe the situation so you can decide if this makes sense for openaddresses.

There are two datasets that would need to be combined:
- An address list in each municipality (or, sometimes, a district within a municipality). For rural addresses, a GPS reading is included. For urban addresses, the id number of a city block and sidewalk segment is given. (This is at ftp://ftp.ibge.gov.br/Censos/Censo_Demografico_2010/Cadastro_Nacional_de_Enderecos_Fins_Estatisticos/)
-  Vector maps with the sidewalk segments. (This is at ftp://geoftp.ibge.gov.br/recortes_para_fins_estatisticos/malha_de_setores_censitarios/censo_2010/base_de_faces_de_logradouros/)

Thus, at least the block corresponding to each address can be determined.

I can't say much about the quality of the data; I suppose it varies a lot throughout the country. One issue is that the geospatial information is often not precisely aligned with OSM, see the pictures below. In any case, this is the best data one can hope to gather in Brazil (apart from a few bigger cities that may release their own addresses) at least until the next census comes out, in 5 to 10 years.

![1](https://cloud.githubusercontent.com/assets/6500902/21565981/905a47d2-ce9e-11e6-8ea7-8f3360b0c930.jpg)
![2](https://cloud.githubusercontent.com/assets/6500902/21565985/9d5aaf94-ce9e-11e6-95b4-01ce8e412eea.jpg)

nvkelso:
This sounds similar to US Census interpolated address ranges ala Tiger.

+1 for OpenAddresses adding support for this data type (at least interpolation points at each end of line).

astoff:
@nvkelso These are _not_ address ranges. They are more like the Japanese dataset: individual addresses geolocated up to city block (as opposed to parcel or rooftop).

nvkelso:
Interesting!

Can you provide a fully formed Brazil example address, and what this dataset would allow? Is it 1:1 or missing something?

astoff:
I have attached a little example of what one can get (a part of Curitiba)
[c-uniq.csv.gz](https://github.com/openaddresses/openaddresses/files/680067/c-uniq.csv.gz).  Judging by the amount of dots, I'd say the dataset is fairly complete.

[...]

PR #2315, opened 2017-01-04
Brazil census addresses

astoff:
This PR is preliminary. The dataset covers the whole country, but I have included only São Paulo state as a sample. If you would like to accept it, please let me know how to upload the preprocessed data to OA's cache. I will then prepare the data for the whole country (one source file per state) and update the pull request.

There are a few choices I made in the preprocessing script that I will be happy to discuss if necessary.


thatdatabaseguy:
São Paulo data LGTM. About 60% have lat/lon but that's still ~740k addresses, virtually all of which have house number (or SN for sem numero), street, city, and postcode.

@migurski there seem to be a few different patterns for caching in the existing scripts. Is there a preference and are these scripts automated or run manually?

migurski:
The scripts are not automated, they get run manually. There's not a strong preference, but if it’s possible to wrap it up in a container like Docker, it may make them easier to run in the future. [Tennessee, New Zealand, and Australia follow that pattern](http://mike.teczno.com/notes/openaddr/docker-address-data.html).

astoff:
I have updated the script to reflect the discussion on https://github.com/openaddresses/openaddresses/issues/2303. I'd say it's ready to merge.

I also have the preprocessed data for the whole country ready to upload (just tell me where). If you prefer to run the script on your end, use https://gist.github.com/astoff/5dc8f0df2e5a63a6750ce981ee1414b5 to generate the JSONs.

nvkelso:
This is super exciting, thanks for your hard work on the PR with repeatable scripts and 1-time data builds :)

The OpenAddress S3 uploader will take files up to 256 mb in size:

- https://results.openaddresses.io/upload-cache

To use it you'll need to be logged in via Github **AND** publicize your membership in the OpenAddresses organization.

nvkelso:
Connects with https://github.com/openaddresses/openaddresses/issues/2303.

nvkelso:
In the GIST you say:

`#        ('53', 'Distrito Federal',         'DF'), #This has a funny address format, needs special conform`

Has that been fixed subsequently, or will we need a followup issue to resolve the special conform for the capital's address format?

migurski:
Fantastic! FYI @nvkelso, I fixed the bug where membership must be public. It can be private now and the upload will still work. Can also raise the limit above 256MB if needed.

astoff:
@nvkelso There are two things with Distrito Federal: First, there is a better statewide dataset (https://github.com/openaddresses/openaddresses/pull/2311, but it is failing to download). Second, I am not really sure how to map Brasilia's address scheme to the usual number + street scheme—i'm trying to check with a local.

thatdatabaseguy:
Looked into addresses in Brasilia a little while back. The city uses a grid system that's unlike anywhere I've ever seen. The [UPU example](http://www.upu.int/fileadmin/documentsFiles/activities/addressingUnit/braEn.pdf) is:

```
Sr. Luis Carvalho
SBN – Quadra 13 – Bloco B – 8º andar
BRASÍLIA – DF
70002–900
BRAZIL
```

Here "SBN" would be a sector of the city, which together with "Quadra 13" (usually combined and shortened, so "SBN 13") is sort of like a neighborhood, but also the name of the principal street within said neighborhood. Then "Bloco B" is a named/numbered building (and "8º andar" = 8th floor).

For OpenAddresses it probably makes the most sense to call "SBN 13" the street and "Bloco B" the house number.

astoff:
@thatdatabaseguy That's how I did it in https://github.com/openaddresses/openaddresses/pull/2311. In the census, everything is concatenated in the street field, so there's no realiable way to split out the "bloco".

thatdatabaseguy:
Ah ok. Would a regex in the transform work? Haven't looked at the raw census files for DF yet but it seems like the addresses there are very structured so something like this might do the trick if everything's concatenated:

```json
    "number": {
        "function": "regexp",
        "field": "street name",
        "pattern": "^(?:.*?)(?:\\s*(Bloco\\b.*))?$"
    },
    "street": {
        "function": "regexp",
        "field": "street name",
        "pattern":  "^(.*?)(?:\\s*Bloco\\b.*)?$"
    }
```

astoff:
Well, there are many variants (bloco, bl, lote, lt) so it would need testing. But I think the other PR (local gov data) is much better anyway. Should we not worry?

astoff:
All right, so (assuming it builds) I will let somebody else do the merging, so others get a change to see if it looks OK.

Please squash and merge—there is a large deleted file in an intermediate commit.

thatdatabaseguy:
Looks good to me! It's probably worthwhile to investigate the other source for Distrito Federal.

migurski:
🇧🇷 a whole new country of data! 🇧🇷

nvkelso:
The data was only for this one region as a sample? I think there may be
some more work to do to get everything in?

migurski:
I do see 30 files in this PR!

thatdatabaseguy:
For whatever reason the map only shows São Paulo, but [all the states were built as part of the PR](http://results.openaddresses.io/jobs/5ad2e0f1-9cfa-4a81-bfba-73d0564c596e).

nvkelso:
Thanks for the links, I see them now :)

migurski:
Ah yeah, sorry, the part of `machine` that decides which map to post back is not the most reliable thing. That’s a bug for me.

PR #2317, opened 2017-01-05
Solothurn, Switzerland

sergiyprotsiv:
Another Swiss canton. This time no registration was required (though one still needs to accept the terms, etc). The main shape file with addresses only contained the id of the district (gemeinde), not the name, so I used QGIS to join the address shape file with the gemeinden shape file.
The original address file is here: https://s3.amazonaws.com/data.openaddresses.io/cache/uploads/sergiyprotsiv/2565b3/sogis_1483513972.zip
The original district file is here: https://s3.amazonaws.com/data.openaddresses.io/cache/uploads/sergiyprotsiv/198a8f/sogis_1483514338.zip

### DIFF
--- a/sources/ch/solothurn.json
+++ b/sources/ch/solothurn.json
@@ -4,7 +4,7 @@
         "city": "Solothurn"
     },
     "website": "http://geoweb.so.ch/geodaten/index.php",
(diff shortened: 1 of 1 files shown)


migurski:
👍

PR #2321, opened 2017-01-05
Grisons/Graubünden, Switzerland

sergiyprotsiv:
Canton of Grisons/Graubünden. 

They have a WMS server (http://geogr.mapserver.ch/shop/?q=de/prod_av, see WMS), but you need to register to get the password to use it. The license file prohibits the distribution of the password, but regarding the data it only says that when the data is published it has no legal value (I assume this largely refer to parcels, ownership, etc).

This is a common theme across the Swiss datasets: they often require registration and acceptance of the terms, but the main issue they bring up in the terms is that the data they supply cannot be used in court, etc. I am assuming this is fine with OA, but maybe someone with better knowledge of German and legal issues can take a look.

In any case, the file I uploaded is the straight dump of the HADR layer of the WFS (which is the common name for house number layers across most cantons who have WFS).

### DIFF
--- a/sources/ch/grisons.json
+++ b/sources/ch/grisons.json
@@ -4,7 +4,7 @@
         "city": "Grisons"
     },
     "website": "http://geogr.mapserver.ch/shop/?q=de/prod_av",
-    "data": "https://s3.amazonaws.com/data.openaddresses.io/cache/uploads/sergiyprotsiv/ecee94/ch-gr.csv.zip",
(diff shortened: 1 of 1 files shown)


sergiyprotsiv:
Ok, the csv seems to have lost the coordinates. Will bring them back shortly.

sergiyprotsiv:
The WFS for some reasons gave MULTIPOINT geometries, which all consisted of a single point, so I converted it to POINT and then exported again.

migurski:
Looks great!

PR #2325, opened 2017-01-05
Glarus, Switzerland

sergiyprotsiv:
Canton of Glarus, Switzerland.

The main data is sourced from the nationwide website with open geodata (https://www.geodienste.ch/services). So far they only have three open datasets of addresses (two I have covered yesterday from respective cantons), but hopefully they will add more in the future.

The data did not have municipality or postcode, so I did a spatial join with the layer of postal code areas (PLZ-Gebiete) from the official Canton of Glarus WFS: https://wfs.geo.gl.ch/Public?VERSION=1.1.0.

### DIFF
--- a/sources/ch/glarus.json
+++ b/sources/ch/glarus.json
@@ -3,7 +3,7 @@
         "country": "ch",
         "city": "Glarus"
     },
-    "website": "http://geogr.mapserver.ch/shop/?q=de/prod_av",
+    "website": "https://www.geodienste.ch/services",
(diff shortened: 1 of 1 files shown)


sergiyprotsiv:
@simonpoole the data is coming straight from here: https://geodienste.ch/db/av/deu?SERVICE=WFS&REQUEST=GetCapabilities
It is one of the sources that do not require any registration and is explicitly public. The conditions on use are here: http://kkgeo.ch/fileadmin/vrstckt/geodienste_terms_de.pdf 
Could you let me know what the proper attribution from that would be?

PR #2326, opened 2017-01-06
Uri, Switzerland

sergiyprotsiv:
This is a WFS dump of the layer "t19_gebaeudeeingang_ims" available at http://service.lisag.ch/wms?request=getCapabilities. More info here: http://geo.ur.ch/#GEODATEN


migurski:
Sergiy, thanks for all your recent work on Switzerland.

PR #2330, opened 2017-01-06
Removing canton abbreviation suffixes from Swiss cities

thatdatabaseguy:
See #2290

### DIFF
--- a/schema/source_schema.json
+++ b/schema/source_schema.json
@@ -308,7 +308,7 @@
         },
         "pattern": {
           "type": "string",
-          "format": "regex"
+          "format": "string"
         },
         "replace": {
           "type": "string"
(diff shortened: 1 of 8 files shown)


thatdatabaseguy:
Re-run this please

thatdatabaseguy:
Yep @trescube that worked. Error wasn't entirely clear, but may have to do with different regex flavors in Javascript vs. Python.

So this commit also changes validation of the regex field to "string" instead of "regex". The webhook would still fail quickly if the regex doesn't compile on the Python side.

thatdatabaseguy:
This should also fix #2311.

Any objections?

trescube:
Nope, merge away!

PR #2333, opened 2017-01-07
Nordrhein-Westfalen, Germany

sergiyprotsiv:
Here's the biggest state of Germany (>4M addresses), apparently they have just released their whole cadastre and more as open data: https://www.opengeodata.nrw.de/produkte/geobasis/lika/alkis_sek/.

The file I am using here is the direct download from https://www.opengeodata.nrw.de/produkte/geobasis/lika/alkis_sek/gebref/gebref_EPSG4647_ASCII.zip
However, that file requires merge over three fields to get to city names. If someone wants to write a small script to do that, I think this source can be kept rather up-to-date.

Currently on Dropbox since S3 refused a 160MB file. If you could lift the limits, can upload there. Otherwise this could be a temporary solution.

### DIFF
--- a/sources/de/nw/statewide.json
+++ b/sources/de/nw/statewide.json
@@ -3,7 +3,7 @@
         "country": "de",
         "state": "nw"
     },
-	"data": "https://www.opengeodata.nrw.de/produkte/geobasis/lika/alkis_sek/gebref/gebref_EPSG4647_ASCII.zip",
+	"data": "https://www.dropbox.com/s/qxeod9xt7oysyar/de-nrw.csv.zip?dl=1",
 	"website": "https://www.opengeodata.nrw.de/produkte/geobasis/lika/alkis_sek/",
 	"type": "http",
 	"compression": "zip",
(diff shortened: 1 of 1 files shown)


sergiyprotsiv:
If successful, this also supersedes at least my older PR: https://github.com/openaddresses/openaddresses/pull/2240
And possibly others, like https://github.com/openaddresses/openaddresses/pull/2275. 

It has to be noted though that this dataset does not have the postal code, so maybe others are still relevant.

iandees:
Wooo! Look at all those dots!

migurski:
I’ve raised the size limit to 256MB so it should now be possible to upload this file.

sources/de/nw/statewide.json
```diff
@@ -0,0 +1,24 @@
+{
+	"coverage": {
+        "country": "de",
+        "state": "nw"
+    },
```
migurski: Can we add an ISO code here? Should look like this [based on the coverage docs](https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#coverage-object):

```JSON
"ISO 3166": { "alpha2": "DE-NW" }
```

PR #2337, opened 2017-01-10
Helsingborg, Sweden

sergiyprotsiv:
Municipality of Helsingborg, Sweden. REST server, sometimes the street name has additional qualifiers, I tried to remove them with regex. Alternatively can parse "ADDRESSLABEL" into street and number.

### DIFF
--- a/sources/se/municipality_of_helsingborg.json
+++ b/sources/se/municipality_of_helsingborg.json
@@ -15,6 +15,6 @@
         },
         "city": "CITY",
        	"postcode": "POSTCODE",
-        "id": "BALID",
+        "id": "BALID"
     }
 }
--- a/sources/se/municipality_of_helsingborg.json
+++ b/sources/se/municipality_of_helsingborg.json
@@ -14,12 +14,12 @@
             "pattern": "^(.*?)(?: \\|.+)?$"
         },
         "city": "CITY",
-       	"postcode": {
+        "postcode": {
             "function": "regexp",
             "field": "POSTCODE",
             "pattern": "^([\\d]{3})([\\d]{2})$",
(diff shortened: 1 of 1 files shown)


sergiyprotsiv:
The sample produced by the machine seems fine. Maybe there is some issue with the regex, could you check it out? Basically some streets follow the pattern "Streetname | Qualifier", so I was trying to remove everything starting with " |".

sources/se/municipality_of_helsingborg.json
```diff
@@ -0,0 +1,20 @@
...
+    },
+    "data": "http://geodata.helsingborg.se/arcgis/rest/services/Trafik/Skolskjuts_160914/MapServer/37",
+    "type": "ESRI",
+    "conform": {
+        "type": "geojson",
+        "number": ["ENUMERATOR","DESIGNATIONNUMBERLETTER"],
+        "street": {
+            "function": "regexp",
+            "field": "STREETNAME",
+            "pattern": "(^.*?)( \\|.+)?$"
```
thatdatabaseguy: Try using a non-capturing group for the second expression: "^(.*?)(?: \\|.+)?$"

sergiyprotsiv:
That worked, thanks!

thatdatabaseguy:
Data looks good. There was one minor thing I noticed which is postcode formatting. This source uses 5-digit postcodes e.g. "12345" whereas Sweden typically uses "123 45". This should work to add the space:

```json
"postcode": {
    "function": "regexp",
    "field": "POSTCODE",
    "pattern": "^([\\d]{3})([\\d]{2})$",
    "replace": "$1 $2"
}
```

PR #2338, opened 2017-01-10
Vaxjö, Sweden

sergiyprotsiv:
Municipality of Vaxjö, Sweden fro their REST server.


migurski:
Looks very remote. :)

PR #2340, opened 2017-01-10
Savsjö, Sweden

sergiyprotsiv:
Savsjö municipality, Sweden. This looks like a distant suburb of what you thought was remote. Cannot check the rest responses myself (QGIS cannot display results), but it is sufficiently similar to the previous sources that I think this should be fine.

### DIFF
--- a/sources/se/municipality_of_savsjö.json
+++ b/sources/se/municipality_of_savsjö.json
@@ -3,7 +3,7 @@
         "country": "se",
         "city": "Savsjö"
     },
-    "data": "https://s3.amazonaws.com/data.openaddresses.io/cache/uploads/sergiyprotsiv/f3824c/se-savsjo.json.zip",
+    "data": "https://s3.amazonaws.com/data.openaddresses.io/cache/uploads/sergiyprotsiv/797599/se-savsjo.zip",
     "website": "http://giswww.savsjo.se/arcgis/rest/services/Savman/SavmanBakgrund/MapServer/3",
     "type": "http",
     "compression": "zip",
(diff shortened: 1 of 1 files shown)


sergiyprotsiv:
Looks like an issue with ESRI downloader. Doing it via the web interface to the server works fine.

iandees:
Curious: I thought all Esri layers had to have an ID column. This is the first one I've seen that doesn't.

This will need some work in `machine` to support this.

slibby:
@iandees it's a broken service, it's really an fluke that it's even able to "appear" published as we do always (apparently not) require an ID field. [Basic queries at REST](http://giswww.savsjo.se/arcgis/rest/services/Savman/SavmanBakgrund/MapServer/3/query?where=1%3D1&text=&objectIds=&time=&geometry=&geometryType=esriGeometryEnvelope&inSR=&spatialRel=esriSpatialRelIntersects&relationParam=&outFields=&returnGeometry=true&returnTrueCurves=false&maxAllowableOffset=&geometryPrecision=&outSR=&returnIdsOnly=false&returnCountOnly=false&orderByFields=&groupByFieldsForStatistics=&outStatistics=&returnZ=false&returnM=false&gdbVersion=&returnDistinctValues=false&resultOffset=&resultRecordCount=&f=html) do work,  but I'm not sure there's any way to have `machine` programatically go through 1000 rows at a time if there's no logical offset with `id`

sergiyprotsiv:
Is it maybe possible to set the "order by" clause to order by for example "XKOORD, YKOORD"? If not, I can probably do this manually and cache it.

iandees:
It looks like the `FNR` column is an ID-like thing that we could de-dupe on. It's just not specified as the Esri ID type so `machine` doesn't use it.

sergiyprotsiv:
I tried ordering by FNR and got duplicates. I think FNR is a parcel ID, which I can imagine host multiple addresses, though I did not check

PR #2345, opened 2017-01-15
Kristinehamn, Sweden

sergiyprotsiv:
Municipality of Kristinehamn, Sweden. Hopefully the regex does its job...


sergiyprotsiv:
Apologies for the mess with the commits, got two sources pulled togged in one PR. Still not 100% obvious to me how git works.
If anyone could suggest the proper regex to extract the house and street from strings like "Almvägen 24A", would be glad.

sergiyprotsiv:
Will re-add these separately

PR #2346, opened 2017-01-15
Update recife.json

astoff:
Does anybody know if this CKAN link refers to a particular version of the data or always points to the latest? (The latter option would be the preferable one, of course.)


simonkas:
Toronto, ontario, Canada real address, using ip scrambling browser, dont understand this anymore or know what to do.

astoff:
Is the tiny polygon (invisible at results.openaddresses.io) in the geometry field really preferable to a point?

iandees:
> Is the tiny polygon (invisible at results.openaddresses.io) in the geometry field really preferable to a point?

Now that we're covering more small areas around the rest of the world we should probably rethink the results.openaddresses.io coverage map. That's a primary source of reward for this work, so seeing a coverage point or polygon show up is important.

migurski:
I agree! I could do something automagic with the map, such as showing a fat dot if the area is below a certain size.

astoff:
Why not use a wikidata link in lieu of coverage? From there one can get a coordinate and, potentially, links to the geometry (and a bunch of other info as well, such as localized names).

migurski:
I was thinking to use Who’s On First for this; it’s Mapzen’s gazetteer. It’ll have GeoJSON links. Maybe we use both? Also Geonames I guess.

migurski:
Some candidate links:

- https://www.wikidata.org/wiki/Q48344
- https://whosonfirst.mapzen.com/spelunker/id/101948669/ ([geojson](https://whosonfirst.mapzen.com/data/101/948/669/101948669.geojson))
- http://www.geonames.org/3390760/recife.html

migurski:
I’ve added an issue about Wikidata, Geonames, and other sources: https://github.com/openaddresses/machine/issues/524

PR #2349, opened 2017-01-15
Kristinehamn, Sweden

sergiyprotsiv:
This one has the regex issue to extract street and house number. Help appreciated.

### DIFF
--- a/sources/se/municipality_of_kristinehamn.json
+++ b/sources/se/municipality_of_kristinehamn.json
@@ -1,0 +1,29 @@
+{
+    "coverage": {
+        "country": "se",
+        "city": "Kristinehamn"
+    },
+    "data": "http://kartor-va.kristinehamn.se/arcgis/rest/services/Bakgrundskartor/Bakgrundskarta_fargad_enkel/MapServer/112",
+    "type": "ESRI",
+    "conform": {
+        "type": "geojson",
+        "number": {
+            "function": "regexp",
+            "field": "Name",
+            "pattern": "([0-9]+([A-Z])?)$"
+        },
+        "street": {
+            "function": "regexp",
+            "field": "Name",
+            "pattern": "^(.*?)(?: [0-9]+([A-Z])?)?$",
+            "replace": "$1"
+        },
+        "city": "PostCity",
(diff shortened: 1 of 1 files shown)


sources/se/municipality_of_kristinehamn.json
```diff
@@ -0,0 +1,29 @@
...
+        "city": "Kristinehamn"
+    },
+    "data": "http://kartor-va.kristinehamn.se/arcgis/rest/services/Bakgrundskartor/Bakgrundskarta_fargad_enkel/MapServer/112",
+    "type": "ESRI",
+    "conform": {
+        "type": "geojson",
+        "number": {
+            "function": "regexp",
+            "field": "Name",
+            "pattern": "([0-9]+([A-Z])?)$"
```
thatdatabaseguy: Might be an issue with the inner capture group. Also, shouldn't this be identical to the regex below just with different captures? I'd try:  "^(?:.*?)(?: ([0-9]+[A-Z]?))?$"

sources/se/municipality_of_kristinehamn.json
```diff
@@ -0,0 +1,29 @@
...
+        },
+        "street": {
+            "function": "regexp",
+            "field": "Name",
+            "pattern": "^(.*?)(?: [0-9]+([A-Z])?)?$",
```
thatdatabaseguy: Same deal here: "^(.*?)(?: [0-9]+[A-Z]?)?$" should work.

sources/se/municipality_of_kristinehamn.json
```diff
@@ -10,12 +10,12 @@
         "number": {
             "function": "regexp",
             "field": "Name",
-            "pattern": "([0-9]+([A-Z])?)$"
+            "pattern": "^(?:.*?)(?: ([0-9]+[A-Z]?))?$"
```
thatdatabaseguy: Oh, maybe the captured group shouldn't be optional so it won't return None: "^(?:.*?)(?: ([0-9]+[A-Z]?))$"

sergiyprotsiv:
Great, thanks, that worked!

PR #2354, opened 2017-01-17
added Vitebsk, Belarus

trescube:


iandees:
`number` and `street` both point to `ADDRESS` column?

trescube:
No, but I can't get the data to load in qgis or via URL (/0/1, /0/2, etc) and it's not cooperating immediately in pyesridump so I'm loading via machine to hopefully get some metadata from the result.

slibby:
@trescube I think this still needs some Regexing, the street number should be alone in NUMBER but the rest seems to differ depending on whether there is a city and district listed. confusing but I think it shouldn't be merged until it's a little cleaner, what do you think? 
image

trescube:
Agreed, I was poking around with the regexes last night but these are pretty tricky given the varying number of comma-delimited fields (there can be 3, 4, or 5).  @thatdatabaseguy, care to take a look?

trescube:
Thanks, @thatdatabaseguy!  If you're happy with the result, let me know and I'll merge.

thatdatabaseguy:
Cool, will check when it's finished building. In general Russian addresses are a classic example of why _not_ to use regex for parsing (the old Soviet form was written largest-first and is still in use, whereas the current standard is smallest-first similar to most Western addresses), though here it should be ok because the data's from a single source and uses relatively consistent conventions.

The regexes I provided can probably result in a few false negatives, but no false positives in a sample of the first 1000 records. The latter is probably more important. Note that these only allows addresses with street names (the others are basically village + house number and are still valid, see the [addr:place](http://wiki.openstreetmap.org/wiki/Key:addr:place) examples in OSM, but perhaps a bit more difficult to parse out).

For having set out to save the world from regexes I sure end up writing a lot of regex :-).

trescube:
Cool, sounds good.  How would libpostal have fared against this dataset?  Or are there machine functions that could be written to tackle this since the regexes are pretty complex?

Oh, I suppose you have permission to merge, too.

thatdatabaseguy:
@trescube probably not very well because of the smaller admins like "Богушевский с/с" (village council, not in OSM and I'm not even sure what to map it to in libpostal) or "д. Коленьки" (village is in OSM but doesn't use the деревня prefix). I did recently start adding the "г." prefix to Russian-language cities but that may need to be population-based so "д." gets used for villages, etc.

@sergiyprotsiv can correct me if I'm wrong, but now that this is written, I'd guess it can be reused for just about any data set from a post-Soviet state where the addresses are written largest-to-smallest (it would also be trivial to write a version that's smallest-to-largest).

If that's the case and there are more data sets like this one, it might make sense to add "post_soviet_{street, house_number, unit, city}" functions to machine.

thatdatabaseguy:
Data LGTM. Merging.

PR #2357, opened 2017-01-18
Israel countrywide

sergiyprotsiv:
Here is a great quality source for Israel. It is available in the open, however I cannot fully determine its provenance.

Has both Latin and Hebrew versions of addresses (as separate files). Many building include "entrances", which I coded as "unit". This is usually a numeral in Hebrew, but I kept it in the Latin source too.

The geometry is oddly a linestring, hopefully the machine handles those too.

### DIFF
--- a/sources/il/countrywide-hebrew.json
+++ b/sources/il/countrywide-hebrew.json
@@ -5,6 +5,7 @@
     },
     "data": "http://www.sol-israel.co.il/arcgis/rest/services/Amidar_Vector/MapServer/9",
     "type": "ESRI",
+    "skip": true,
     "language": "HE",
     "conform": {
         "type": "geojson",
--- a/sources/il/countrywide-hebrew.json
+++ b/sources/il/countrywide-hebrew.json
@@ -7,6 +7,7 @@
     "type": "ESRI",
     "language": "IW",
     "conform": {
+        "type": "geojson",
         "city": "SETL_NAME",
         "number": "HOUSE_NUM",
         "unit": "ENTRY_LETR",
--- a/sources/il/countrywide-hebrew.json
+++ b/sources/il/countrywide-hebrew.json
@@ -5,12 +5,17 @@
     },
     "data": "http://www.sol-israel.co.il/arcgis/rest/services/Amidar_Vector/MapServer/9",
     "type": "ESRI",
-    "language": "IW",
+    "language": "HE",
+    "skip": true,
     "conform": {
         "type": "geojson",
         "city": "SETL_NAME",
-        "number": "HOUSE_NUM",
+        "number": {
+            "function": "format",
+            "fields": ["HOUSE_NUM", "ENTRY_LETR"],
+            "format": "$1$2"
+        },
         "unit": "ENTRY_LETR",
-        "street": "STR_FLNM"
+        "street": "STR_FULNAM"
     }
 }
--- a/sources/il/countrywide-hebrew.json
+++ b/sources/il/countrywide-hebrew.json
@@ -6,7 +6,6 @@
     "data": "http://www.sol-israel.co.il/arcgis/rest/services/Amidar_Vector/MapServer/9",
     "type": "ESRI",
     "language": "HE",
-    "skip": true,
     "conform": {
         "type": "geojson",
         "city": "SETL_NAME",
--- a/sources/il/countrywide-hebrew.json
+++ b/sources/il/countrywide-hebrew.json
@@ -1,0 +1,15 @@
+{
+    "coverage": {
+        "country": "il",
+        "ISO 3166": {"alpha2": "IL", "country": "Israel"}
+    },
+    "data": "http://www.sol-israel.co.il/arcgis/rest/services/Amidar_Vector/MapServer/9",
+    "type": "ESRI",
+    "language": "IW",
+    "conform": {
+        "city": "SETL_NAME",
+        "number": "HOUSE_NUM",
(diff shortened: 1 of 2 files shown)


slibby:
This is the source company: http://www.sightvision.co.il/

thatdatabaseguy:
Gadol!

Entrance letter appears to be [a part of house number for Israeli addresses](https://en.wikipedia.org/wiki/Address_(geography)#Israel). The source indicates this as well under "Labeling Info": "Label Expression: [HOUSE_NUM] CONCAT [ENTRY_LETR]".

The "format" function can be used to concatenate multiple fields without space:

```json
"number": {
    "function": "format",
    "fields": ["HOUSE_NUM", "ENTRY_LETR"],
    "pattern": "$1$2"
}
```

Also, for some reason only seeing street name for the Latin data set, not in Hebrew. Any other fields that it could be? The most common street type is "שדרות" or "Sderot" in Hebrew if that's helpful (although there is also a small town named "Sderot", yay for ambiguity)..

nvkelso:
Are they a general GIS services company that put together this data for a
gov’t agency with CC0 or CC-BY license, or did we happen to find an
unprotected commercial endpoint?

slibby:
@nvkelso my guess is a unprotected endpoint used to support their other applications.

sergiyprotsiv:
Yes, I was not sure about this one, happy close it. Shall I just open and issue in case there are questionable source like this one before creating a PR?

nvkelso:
We should still import it, but you’ll need to include `skip: true` on it
until we can verify the license.

(Even if the license still doesn’t allow use, we can keep this source
around with `skip: true` still set – that flag should prevent it from
showing up in the download builds.)

sources/il/countrywide-hebrew.json
```diff
@@ -0,0 +1,16 @@
...
+    },
+    "data": "http://www.sol-israel.co.il/arcgis/rest/services/Amidar_Vector/MapServer/9",
+    "type": "ESRI",
+    "language": "IW",
+    "conform": {
+        "type": "geojson",
+        "city": "SETL_NAME",
+        "number": "HOUSE_NUM",
+        "unit": "ENTRY_LETR",
+        "street": "STR_FLNM"
```
thatdatabaseguy: Ah, that's why it's not showing up. The Hebrew field name is "STR_FULNAM"

sources/il/countrywide-hebrew.json
```diff
@@ -0,0 +1,16 @@
...
+        "ISO 3166": {"alpha2": "IL", "country": "Israel"}
+    },
+    "data": "http://www.sol-israel.co.il/arcgis/rest/services/Amidar_Vector/MapServer/9",
+    "type": "ESRI",
+    "language": "IW",
+    "conform": {
+        "type": "geojson",
+        "city": "SETL_NAME",
+        "number": "HOUSE_NUM",
+        "unit": "ENTRY_LETR",
```
thatdatabaseguy: Mentioned this in the PR,  but there was other licensing discussion going on as well. Entrance letter is part of the house number, not the unit. The "format" conform function can be used to concatenate multiple fields without space:

```json
"number": {
    "function": "format",
    "fields": ["HOUSE_NUM", "ENTRY_LETR"],
    "pattern": "$1$2"
}
```

sources/il/countrywide-latin.json
```diff
@@ -0,0 +1,15 @@
...
+        "country": "il",
+        "ISO 3166": {"alpha2": "IL", "country": "Israel"}
+    },
+    "data": "http://www.sol-israel.co.il/arcgis/rest/services/Amidar_Vector/MapServer/9",
+    "type": "ESRI",
+    "conform": {
+        "type": "geojson",
+        "city": "SETL_NAME_LTN",
+        "number": "HOUSE_NUM",
+        "unit": "ENTRY_LETR",
```
thatdatabaseguy: Same as above re: ENTRY_LETR

sources/il/countrywide-hebrew.json
```diff
@@ -0,0 +1,16 @@
+{
+    "coverage": {
+        "country": "il",
+        "ISO 3166": {"alpha2": "IL", "country": "Israel"}
+    },
+    "data": "http://www.sol-israel.co.il/arcgis/rest/services/Amidar_Vector/MapServer/9",
+    "type": "ESRI",
+    "language": "IW",
```
thatdatabaseguy: Language code should be "HE"

thatdatabaseguy:
Re-run this, please.

thatdatabaseguy:
Hm, it seems skip: true also prevents it from building here. Can we try one build without it so the correct data files are built and then add it in for merging?

nvkelso:
Good point about skip and map previews, filed: https://github.com/openaddresses/machine/issues/522

thatdatabaseguy:
Hm, failed on a timeout.

Re-run this please.

thatdatabaseguy:
Ok, that time the Latin version worked. Probably has to do with concurrent requests to their server. @sergiyprotsiv try adding skip: true to the Latin version (now that we have a copy) and not the Hebrew one and let's see if it builds.

sergiyprotsiv:
Re-run this, please.

slibby:
@sergiyprotsiv based on the discussion above, I think `countrywide-hebrew` needs the `skip:true` added back to it, and then this can be merged.

sergiyprotsiv:
Sure, I did not realize it was gone from that file.

PR #2367, opened 2017-01-19
Almaty, Kazakhstan

sergiyprotsiv:
Here is a layer for Almaty, the largest city in Kazakhstan. This is sourced from the public map of the Attorney General GIS website infopublic.pravstat.kz (which they use to display crime rates, traffic collisions, etc).

I do not think there is any license attached, but it is in public access from a governmental website specifically created for public purpose. If you think it's fine, I'll format the rest of the country.

### DIFF
--- a/sources/kz/ala-statewide.json
+++ b/sources/kz/ala-statewide.json
@@ -1,0 +1,17 @@
+{
+    "coverage": {
+        "country": "kz",
+        "ISO 3166": {
+          "alpha2": "KZ-ALA", 
+          "country": "Kazakhstan",
+          "subdivision": "Almaty"}
+    },
+    "data": "http://infopublic.pravstat.kz:8399/arcgis/rest/services/Kazakhstan/MapServer/20",
+    "attribution": "Комитет по правовой статистике и специальным учетам Генеральной прокуратуры Республики Казахстан",
+    "type": "ESRI",
+    "conform": {
+        "city": "Алматы",
+        "number": "NUMBER_1",
+        "street": "STREET_NAM"
+    }
+}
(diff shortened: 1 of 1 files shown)


nvkelso:
I think this is the same rational as USA gov't sources: we include it as-is unless there is a request to take it down (and in that case we'd keep the source file but mark it as `skip: true`).

sergiyprotsiv:
Sounds good! 

In terms of data there is also a small offset related to OSM, not sure what he reason is. Many houses are also listed multiple times since the source has information on the number of floors, so the building that do not have a uniform floor count are listed twice...

sergiyprotsiv:
I am actually exploring other regions in the data and they all lack city names (it was fine for the first two since they are effectively single-city regions). I have found another source on the same website that seems to have country-wide consistent coverage, so I will probably use that and do a join with another table (or a spatial join with polygons).

slibby:
Hm, I think this needs some encoding help - the street names come out garbled for me:

image

But maybe it's just me. @migurski have we used encoding attributes in the transform for this before?

migurski:
It does look garbled. If the encoding is known, then it should be possible to specify it [with an `encoding` tag](https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#processing-tags). Here’s an example: [`sources/fr/guadeloupe.json`](https://github.com/openaddresses/openaddresses/blob/master/sources/fr/guadeloupe.json).

thatdatabaseguy:
Looks fine on my end. Might be that OpenOffice or whatever program that is uses an encoding other than UTF-8 by default, but the file's definitely UTF-8 encoded.

migurski:
Okay, sounds like this can be merged.

PR #2371, opened 2017-01-19
fixing house numbers in Volgograd oblast

thatdatabaseguy:
Number + letter should be concatenated. Also there was a spelling error for the field "construct" (the building number, more commonly used than letter) so wasn't getting picked up.


thatdatabaseguy:
Now we're cooking with gas. Merging.

PR #2375, opened 2017-01-19
Fix North Glengarry house numbers

thatdatabaseguy:
Regex didn't account for hyphenated numbers.

Also using "replace": "$1" will insert the full string if the field doesn't match the regex. One way to ensure that records match the regex is to remove the "replace" directive and only use one capture group as the default behavior in that case is to whitespace-join all captured groups as a string.

There's also probably a larger discussion to be had about the desired behavior of "replace", but this method will work in the meantime.

### DIFF
--- a/sources/ca/on/north_glengarry.json
+++ b/sources/ca/on/north_glengarry.json
@@ -12,8 +12,7 @@
         "number": {
             "function": "regexp",
             "field": "LOCATION",
-            "pattern": "^([0-9]+)( .*)",
-            "replace": "$1"
+            "pattern": "^([0-9]+(?:\-[0-9]+)?)(?: .*)"
         },
         "street": "STREETNAME",
         "type": "shapefile"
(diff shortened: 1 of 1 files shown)


iandees:
😢

thatdatabaseguy:
That's bizarre.

Does it have to do with [this commit](https://github.com/openaddresses/openaddresses/commit/ca840cc4538c1f8ef22ac1d1af2e412214288972)? That seems to have been pushed before [my change](https://github.com/openaddresses/openaddresses/commits/master/schema/source_schema.json?author=thatdatabaseguy) which fixed a similar error (JS validation of regexes that use a different flavor in Python), but if there was a complex merge then maybe that's not the actual history.

Also @trescube: is there a way to make the schema validator raise less cryptic exceptions?

thatdatabaseguy:
Solved. "type": "HTTP" needed to be lowercased. Still would be helpful if that were in the error message, and that doesn't explain the other cases.

PR #2380, opened 2017-01-19
adding Mogilev, Belarus

thatdatabaseguy:
Same format/regexes as Vitebsk


thatdatabaseguy:
Only 1154 addresses in there, which seems low, but otherwise looks ok. Maybe there's a better layer.

sergiyprotsiv:
There is another layer with 250K addresses here: http://mogilev.gismap.by/arcgis/rest/services/Poisk/MapServer/2

thatdatabaseguy:
Thanks @sergiyprotsiv! Trying that one now.

thatdatabaseguy:
Huzzah, looks like about 140k with house numbers!

I couldn't find a working, accessible layer for the other Belarusian oblasts, but if you do come across one, please feel free to copy these configs and add it in. The regexes should be working pretty well at this point (for parsing this style of address anyway).

sergiyprotsiv:
There are a bunch of sources for Gomel (one for each part of the region): http://gomel.gismap.by/arcgis/rest/services/poisk/MapServer
And that service also has data for Minsk: http://gomel.gismap.by/arcgis/rest/services/Poiskminsk/MapServer/2
They occasionally time out though have some character encoding issues. Pyersidump gets them via quad tree queries, but the text os all wrong.

thatdatabaseguy:
Ah, had tried that source but when trying to test a query in the browser it returned a 403. If the Gomel source is working programmatically I can add it.

PR #2385, opened 2017-01-20
Moscow, Russia

sergiyprotsiv:
This is an official dataset for Moscow. There are plenty of more convenient ESRI servers with data for Moscow, but this one is actually published by the city of Moscow as an official open data resource, so I think it is most in line with the priorities of OA.

I am attaching the script to get the data from their APIs (where they use a custom JSON format). Most addresses refer to Moscow, but some are for nearby towns, in which case the city name is prepended to the address. The original address line also contains the house number, the unit ("korpus") and the building number, which are all used to identify a specific building and thus I concatenated them as a the "number" property. It might be better to put them in a specific format: "д. 5а к. 10 с. 2" or something like that. Note that none of the three sub-parts are required (though "house" is usually there).  

Parsing the street and city names are down in clumsy Python, but can probably be just as well accomplished using regex (after I've seen what can be done in Belarus). Would be good to ensure that the default value for "city" is "Москва" though.

### DIFF
--- a/scripts/ru/msk_parse.py
+++ b/scripts/ru/msk_parse.py
@@ -1,0 +1,122 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+import re
+import sys
+import requests
+import json
+from shapely.geometry import Polygon, Point
+
+
+base_url = "http://api.data.mos.ru/v1/datasets/1927/"
+dump_filename = "/users/seriozha/Downloads/ru-msk.csv"
+
+s = requests.Session()
+outfile = open(dump_filename,'w')
+
+count = int(s.get(base_url + "count").content)
+
+print "Processing " + str(count) + " rows..."
+
+step = 500
+
+outfile.write("GUID,Address,Raion,Housenumber,Unit,Building,CityParsed,StreetParsed,lon,lat\n")
+
+
+for i in range(25000/step, count / step + 1):
+	step_url = base_url + "rows?$top=" + str(step) + "&$orderby=global_id&$skip=" + str(i * step)
+	print step_url
+	
+	for item in json.loads(s.get(step_url).content):
+		cells = item["Cells"]
+		
+		export_string = str(cells["global_id"]) + ","
+		
+		full_address = cells["ADRES"]
+		admArea = cells["AdmArea"]
+		house = cells["DMT"]
+		korp = cells["KRT"]
+		bldg = cells["STRT"]
+		
+		
+		if full_address is not None and len(full_address) >0 :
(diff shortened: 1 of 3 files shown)


scripts/ru/msk_parse.py
```diff
@@ -0,0 +1,122 @@
...
+		if house is not None and len(house) > 0:
+			elems.pop()
+		
+		if len(elems) > 0:
+			matchstring = elems[0]
+			m = re.search(u'(^(г. )|(г .)|(город )|(деревня )|(посёлок ))(.*)', matchstring)
+
+			# Check if the first field contains 
+			if m is not None:
+				city = matchstring[len(m.group(1)):len(matchstring)]
```
thatdatabaseguy: It's useful to keep the city name as it is in the source data. I'd keep the prefixes.

sources/ru/mow/statewide.json
```diff
@@ -0,0 +1,31 @@
...
+    "license": {
+        "url": "https://apidata.mos.ru/Home/Terms",
+        "attribution": true,
+        "share-alike": false
+        },
+    "type": "http",
+    "compression": "zip",
+    "language": "ru",
+    "conform": {
+        "number": ["Housenumber","Unit","Building"],
```
thatdatabaseguy: The Unit field should be moved to "unit" instead of "number".

For Building, one nice way to do this that stays true to the source might be with the format function like so:

```
"number": {
    "function": "format",
    "fields": ["Housenumber", "Building"],
    "format": "дом $1, корпус $2"
}
```

Note that if "Building" is null in this case, format will remove the intervening text so it should just be e.g. "дом 4"

thatdatabaseguy: Oh, I may have read that wrong re: how these addresses work. Thought "корпус" was building. Same idea could apply though.

```
"number": {
    "function": "format",
    "fields": ["Housenumber", "Unit", "Building"],
    "format": "дом $1, корпус $2, строение $3"
}
```

sergiyprotsiv:
Ok, did the changes, but will probably be unable to re-run the source until Sunday (to get the city names in).  If we use the Building as part of the number, it should be "дом $1, строение $2" instead of "дом $1, корпус $2" (корпус = unit)

thatdatabaseguy:
My mistake. If it's generally understood to be part of the house number, it should go under number. See the review comment for that format.

Assuming there's still sub-building information that could follow that like "кв.2", etc.

thatdatabaseguy:
Data looks great! 153k addresses with house numbers.

nvkelso:
WOW, great work!

migurski:
Nice. 👏

PR #2386, opened 2017-01-20
added Guarulhos, Brazil

astoff:


migurski:
Frustrating, maybe we should wait a little and then re-run.

astoff:
Re-run this, please.

astoff:
Since this server is unreachable from some places (maybe a firewall problem?), I have cached the data.

For the record, this dataset has no clear license information.

PR #2388, opened 2017-01-20
adding Santos, Brazil

astoff:


astoff:
i can fix this, but you may want to check why [this run](https://s3.amazonaws.com/data.openaddresses.io/runs/148350/output.txt), with [this version of the json](https://github.com/openaddresses/openaddresses/blob/1d0c336c1685df29a6b396108bbe3668b9d7473d/sources/br/sp/santos.json) failed. Seems to be a bug with GML processing.

migurski:
When I download the URL, I get a zip file containing a shapefile called `lotes.shp`, so the `type: xml` on that run shouldn't work. In [the last run](https://s3.amazonaws.com/data.openaddresses.io/runs/148356/output.txt), the error is a `KeyError: 'complemento'`. Maybe change the source JSON to `complement` to [match the existing fields](https://s3.amazonaws.com/data.openaddresses.io/runs/148356/sample.json)? This might be connected to shapefile’s 10-char limit for field names.

astoff:
@migurski Sorry, I had the wrong log link in my comment above, see the update. The relevant lines for the GML error seem to be
~~~~
017-01-20 17:47:38,145     INFO: Downloaded 204346637 bytes for file /tmp/worker-bmd8jc0b/work-0ne7ufu3/out/process_one-ymwb3sxr/cache-7jt_087o/http/br--sp--santos-81405639.xml
 ...
2017-01-20 17:47:40,521  WARNING: Could not find a .gml file
~~~~

And by the way, in the [first attempt](http://results.openaddresses.io/jobs/122074ee-d142-4075-9676-863a806a0bb3) for this source I requested a geojson but the machine didn't do the necessary CRS conversion, even though the end of the downloaded geojson says
~~~~
"crs":{"type":"name","properties":{"name":"urn:ogc:def:crs:EPSG::29193"}}
~~~~

migurski:
I think the GeoJSON CRS issue is because projections have been deprecated from the format, so only 4326 is now allowed there.

The XML/GML fix has been started in https://github.com/openaddresses/machine/pull/523, but I think some of the tests will need to be updated.

PR #2392, opened 2017-01-21
Brest, Belarus

sergiyprotsiv:
One of the few remaining missing places in Belarus. The data source seems similar to others, but has some encoding issues (the text on the server is all encoded as "CP1251"). I have put the "encoding" tag in the conform, but not sure if the machine can pick it up for an ESRI source.


sergiyprotsiv:
Ok, so the encoding did not quite work (maybe the ESRI servers are assumed to always give back UTF8, but in this case this is definitely Windows-1251). Any suggestions for a fix?

thatdatabaseguy:
It seems to process most of the file before failing, so maybe the file has mixed encoding.

One library I highly recommend for Python when dealing with Internet text is [ftfy](https://github.com/LuminosoInsight/python-ftfy). This will fix many common encoding problems including Mojibake, double-encoding, etc. in one function. Here's an example from this data set's sample JSON:

```python
import ftfy
input = u"Р‘СЂРµСЃС‚СЃРєР°СЏ РѕР±Р»., РЎС‚РѕР»РёРЅСЃРєРёР№ СЂ-РЅ, Р‘РµР»РѕСѓС€СЃРєРёР№ СЃ/СЃ, Рґ. Р С‹Р±РЅРёРєРё, СѓР». РўРѕР»СЃС‚РѕРіРѕ, Рґ. 17"
ftfy.fix_encoding(input)
# returns: "Брестская обл., Столинский р-н, Белоушский с/с, д. Рыбники, ул. Толстого, д. 17"
```

Whatever the input encoding, it gets re-encoded as UTF-8 and then converted to Unicode. I use this lib currently in post-processing OpenAddresses for libpostal because some of the data files still have incorrect encodings. Works like a charm.

@sergiyprotsiv this could probably be implemented as a quick preprocessing script for the time being, but I'd also like to make a PR sometime soon to include this as a dependency for machine and make it default behavior.

thatdatabaseguy:
LGTM

PR #2394, opened 2017-01-21
Kazakhstan countrywide

sergiyprotsiv:
This the countrywide layer for Kazakhstan. The coverage seems okay, though only really covers urban areas. A trivial script to get it working is attached.

I did not set the "city" column as I think there was no consensus (and the district column often actually says things like "XXX District", so it would be weird to have that as a city).

### DIFF
--- a/sources/kz/countrywide.json
+++ b/sources/kz/countrywide.json
@@ -1,0 +1,23 @@
+{
+    "coverage": {
+        "country": "kz",
+        "ISO 3166": {
+          "alpha2": "KZ", 
+          "country": "Kazakhstan"}
+    },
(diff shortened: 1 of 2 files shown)


sources/kz/countrywide.json
```diff
@@ -0,0 +1,23 @@
...
+    "type": "http",
+    "compression": "zip",
+    "language": "ru",
+    "conform": {
+        "type": "csv",
+        "number": "number",
+        "street": "street",
+        "district": "district",
+        "region": "region",
+        "srs": "EPSG:32642",
```
iandees: Huh, I'm surprised this worked. `esri2geojson` should have dumped to `EPSG:4326`.

PR #2396, opened 2017-01-21
Komi Republic, Russia

sergiyprotsiv:
Data for the northerly Komi Republic

### DIFF
--- a/sources/ru/ko/statewide.json
+++ b/sources/ru/ko/statewide.json
@@ -13,7 +13,7 @@
     "language": "ru",
     "conform": {
         "number": "NUM_H",
-        "street": "NAME_S",
+        "street":  ["TYPE_S", "NAME_S"],
         "city": ["TYPE_NP","NAME_NP"],
         "id": "OKATO_ID",
         "type": "geojson"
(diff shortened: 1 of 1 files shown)


sergiyprotsiv:
This server is a bit slow, but I got it working locally. Is it possible to set the default timeout to something like 300 seconds?

migurski:
Yes, I think I can do that! I'm unsure where it’s set now, but: https://github.com/openaddresses/machine/issues/528

@sergiyprotsiv, would you like to be added to the OA Github org so you can work on branches directly in this repository?

sergiyprotsiv:
Re-run this please.

sergiyprotsiv:
Re-run this please.

sergiyprotsiv:
Re-run this please.

sergiyprotsiv:
Nooo, it got 616K out of 640K and died. Try again, otherwise will upload a cached version.

thatdatabaseguy:
The "TYPE_S" field refers to the thoroughfare type e.g. улица, so maybe it should be:

```json
"street":  ["TYPE_S", "NAME_S"],
```

or perhaps the inverse order?

nvkelso:
Can this be merged?

sergiyprotsiv:
I think this one is fine too, though only a fraction of building s contain an address

thatdatabaseguy:
LGTM

PR #2397, opened 2017-01-22
Fix coverage geometry

astoff:
lat/lon order was wrong in geometry field


astoff:
This is ready to merge – one of the ESRI servers timed out, but the source files are correct now.

PR #2155, opened 2016-11-23
Added us/ks/sedgwick

davidchiles:

PR #2423, opened 2017-01-24
Added us/ks/city_of_wichita

### LINKED ISSUE #2002: Add addresses in Wichita, Kansas
Status:
- [ ] **Research sources** and propose best source (and it's data license)
- [ ] **Review proposed source**
- [ ] **Approve source** based on review comments
- [ ] **Add** to OpenAddresses

_NOTE: Proposed source notes should include basic column mapping to OpenAddress data model._

davidchiles:
Fixes #2002

### DIFF
--- a/sources/us/ks/city_of_wichita.json
+++ b/sources/us/ks/city_of_wichita.json
@@ -1,0 +1,34 @@
+{
+    "coverage": {
+        "geometry": {
+            "type": "Point",
+            "coordinates": [
+                -97.34,
+                37.68
+            ]
+        },
+        "US Census": {
+            "geoid": "2079000",
+            "name": "City of Witchita",
+            "state": "Kansas"
+        },
+        "country": "us",
+        "state": "ks",
+        "city": "Witchita"
+    },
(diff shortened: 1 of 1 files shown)


thatdatabaseguy:
There's a field called Prop_Addr_Unit in this data set that's appended to Prop_Addr and thus the street field in the output. This could be corrected with "remove_postfix", but I don't think the conforms allow functions to be chained.

davidchiles:
@thatdatabaseguy I think I understand what you mean. 

`street` needs both:
```js
{
    "function": "postfixed_street",
    "field": "Prop_Addr"
}
```
and 
```js
{
    "function": "remove_postfix",
    "field": "Prop_Addr",
    "field_to_remove": "Prop_Addr_Unit"
}
```

Which isn't possible?

davidchiles:
I also notice from the render that this is really all of Sedgwick County. Related #2155.

thatdatabaseguy:
@davidchiles  yep. Perhaps defining a "compound" function for machine would be worthwhile - that can probably save more than one nasty regex. Will try to make a PR for it over the next few days.

iandees:
Re-run this, please.

PR #2425, opened 2017-01-24
Updating license information in Switzerland

sergiyprotsiv:
After a review of the licenses using suggestions by @simonpoole, here are some of the changes to the cantons that I have contributed:
- Removed canton of Schwyz since the license poses restriction on use without prior confirmation with the source (if someone want to contact the source, please go ahead)
- Fixed the attribution in some cantons to be more specific
- Removed the license data to not imply CC0 in Uri
- This leaves Uri, Luzern and Aargau without the licenses since currency we are using the public ERSI/WMS servers and the copyright information in them is not readily present.

The status of Grisons is to be determined depending on whether I have interpreted the license correctly: https://github.com/openaddresses/openaddresses/commit/17e4e29fac52b032fc68d6fcdf1827df98a4a554

### DIFF
--- a/sources/ch/basel-land.json
+++ b/sources/ch/basel-land.json
@@ -11,7 +11,7 @@
     "type": "http",
     "compression": "zip",
     "license": {
-        "attribution": true,
+        "attribution": "Geodaten des Kantons Basel-Landschaft",
         "share-alike": false,
         "url": "https://www.geo.bl.ch/servlet/redirector/standard/geoshop/BLGeoShop.igs?user=public&password=public"
     },
--- a/sources/ch/basel-land.json
+++ b/sources/ch/basel-land.json
@@ -11,7 +11,8 @@
     "type": "http",
     "compression": "zip",
     "license": {
-        "attribution": "Geodaten des Kantons Basel-Landschaft",
+        "attribution": true,
+        "attribution name": "Geodaten des Kantons Basel-Landschaft",
         "share-alike": false,
         "url": "https://www.geo.bl.ch/servlet/redirector/standard/geoshop/BLGeoShop.igs?user=public&password=public"
     },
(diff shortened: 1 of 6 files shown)


simonpoole:
Copyright doe not enter the picture at this level. Access to Government geo-data is regulated in the federal and cantonal geo-information laws. To determine on what terms a dataset can be distributed you need to check the relevant meta-data for special conditions and if none, the relevant terms of use or similar published by the data owner. Assuming that just because you can access a server or a dataset and can't immediately find T&Cs for is naturally does not mean it is a free for all (and wouldn't with copyright either).

sergiyprotsiv:
Re-test this please

sergiyprotsiv:
Let me know if you want more of the Swiss sources hidden due to licensing. Perhaps better be on the safer side.
In any case, would be great to merge this PR once it passes tests (it seems to fail on strings in "attribution" tags after they were changed from bools).

sergiyprotsiv:
I'm going to merge this unless anyone has objections. It will make one of the cantons invisible and fix the licensing/attribution for others.

nvkelso:
The "attribution name" syntax looks weird to me, let's get a review from @iandees or @migurski

iandees:
I agree that "attribution name" sounds odd. Do we have a different precedent for the required attribution language? I might be thinking of the OSM layer index, which uses "attribution text" or similar.

migurski:
We can change it to whatever. "Name" sounded right to me in the context of an owner who had to be attributed, but if it's objectionable we can modify the scheme.

nvkelso:
If that's the property name now then seems mergeable.

Where should I file issue for changing property name / add documentation? In this repo?

icekalt89:
l1111

0000))66)))
)
)
))
)'-1

migurski:
Ops might be best, since coordinated changes will need to happen between a few repos. Here‘s where the previous changes to license tags were done: https://github.com/openaddresses/openaddresses-ops/issues/7

nvkelso:
Filed https://github.com/openaddresses/openaddresses-ops/issues/18 for the `"attribution name"` followup.

Issue #2411, opened 2017-01-24
Missing data in Winnipeg, Canada


nvkelso:
Big city, no data. @justinelliotmeyers you want to take a look?

justinelliotmeyers:
https://data.winnipeg.ca/City-Planning/Addresses/cam2-ii3u

justinelliotmeyers:
although we don't need parcels, here they are: https://data.winnipeg.ca/City-Planning/Survey-Parcel/emk4-cdaw

justinelliotmeyers:
![winnipeg_ca](https://cloud.githubusercontent.com/assets/5324316/22254250/094599b2-e222-11e6-9d2d-aee2ce84c14a.jpg)

trescube:
@justinelliotmeyers how did you get this into QGIS?  The CSV download has the lat/lon values in a concatenated field that the machine can't split apart.  The JSON download isn't GeoJSON.

justinelliotmeyers:
I honestly don't remember. I may have just manually ripped apart the csv and made the geometry (lat long) fields work (maybe a sed with "(" and a few other things)

or I did it in arcmap and not qgis - but I really don't remember!? Do you need help with it?

trescube:
Oh, nope, I can extract what I need, just didn't know if you were able to easily import it.

justinelliotmeyers:
did we ever implement this?

nvkelso:
Not yet, it's in the `ready for PR` backlog.

nvkelso:
This data is on the map now, closing.

<img width="317" alt="screen shot 2017-07-25 at 20 42 26" src="https://user-images.githubusercontent.com/853051/28603557-d05fa074-7179-11e7-9281-e960e4f96536.png">

PR #2426, opened 2017-01-24
Add City of Winnipeg, Manitoba

iandees:
Fixes https://github.com/openaddresses/openaddresses/issues/2411

Currently stuck because the lat/lon is specified as a single column named `Geometry` what the format `(lat, lon)` and I'm not sure what the best way to split those apart is.

### DIFF
--- a/sources/ca/mb/city_of_winnipeg.json
+++ b/sources/ca/mb/city_of_winnipeg.json
@@ -29,6 +29,18 @@
         "unit": [
             "Unit Type",
             "Unit Number"
-        ]
+        ],
+        "lat": {
+            "function": "regexp",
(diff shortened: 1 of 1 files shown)


migurski:
In theory, a regexp should work here?

migurski:
Hm, maybe that doesn’t work. It would be nice if it worked. I might see if there's an easy way to fix Machine for this.

nvkelso:
Bump

nvkelso:
@migurski bump

migurski:
Thanks for the bump, I'll look at this today or tomorrow.

iandees:
Closing this because @ingalls' stopgap is fine.

PR #2427, opened 2017-01-25
adding Wuhan, China

thatdatabaseguy:
This just looked too good, couldn't pass up.

### DIFF
--- a/sources/cn/42/wuhan.json
+++ b/sources/cn/42/wuhan.json
@@ -1,6 +1,6 @@
 {
     "coverage": {
-        "country": "cn"
+        "country": "cn",
         "state": "42",
         "city": "Wuhan"
     },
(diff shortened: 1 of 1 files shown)


thatdatabaseguy:
Hm, only 1001 addresses, which seems like a count issue.

thatdatabaseguy:
Couldn't find anything wrong with the source, might just be 1001 addresses.

trescube:
Looks like it goes up to 3427:  http://www.digitalwuhan.gov.cn:8399/arcgis/rest/services/HHLJ/hhljdy/MapServer/0/3427

I dumped with pyesridump but it only got 999, I'm not sure how to get the rest.  It looks like pyesridump just grabs the first page when the `Source does not support feature count` in this case.

PR #2430, opened 2017-01-26
City of Kassel, Germany

sergiyprotsiv:
City of Kassel, Germany using their ESRI server.


migurski:
Experimental, but per-run dotmaps seem to be holding up: http://ec2-54-172-233-64.compute-1.amazonaws.com/150398/

nvkelso:
Loads snappy on my home wifi :)

PR #2433, opened 2017-01-26
Venice, Italy

sergiyprotsiv:
A data source for Venice, where we do not have data other than the relatively poor nationwide dataset.

### DIFF
--- a/sources/it/34/venice.json
+++ b/sources/it/34/venice.json
@@ -7,6 +7,7 @@
     "type": "ESRI",
     "language": "it",
     "conform": {
+        "type": "geojson",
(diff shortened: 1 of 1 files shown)


thatdatabaseguy:
Hm, looks like "_" is used as a NULL value for CIVICO_SUB (the "esponente"), but since the full address field looks to be well formatted with commas, maybe try regexing the house number out of the string?

```json
"number": {
    "function": "regexp",
    "field": "INDIRIZZO",
    "pattern": "^(?:.+?)(?:,\\s*(.*?)\\s*)$"
},
```

PR #2436, opened 2017-01-26
Added us/sc/orangeburg

davidchiles:


trescube:
Could you add `postcode` support using the `ADD_ZIP` field?  Otherwise it looks great!

trescube:
👍

PR #2437, opened 2017-01-27
Chechen Republic, Russia 

sergiyprotsiv:
This covers the majority of Chechen Republic (though not completely). From their official GIS site.
Contains five nearly identical source files for separate cities (Grozny has a regex for city since in that file if the address was in the main city the field was left blank).

### DIFF
--- a/sources/ru/ce/argun.json
+++ b/sources/ru/ce/argun.json
@@ -3,7 +3,7 @@
         "country": "ru",
         "city": "Argun"
     },
-    "data": "http://geoportal95.ru:6080/arcgis/rest/services/%D0%90%D0%B4%D1%80%D0%B5%D1%81%D0%BD%D1%8B%D0%B9_%D0%BF%D0%BB%D0%B0%D0%BD_zd/MapServer/5",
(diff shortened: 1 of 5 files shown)


nvkelso:
Is this ready to merge?

sergiyprotsiv:
I think this one is fine

migurski:
Whoa:

migurski:
Seeing some curious offset:

<a href="http://ec2-54-172-233-64.compute-1.amazonaws.com/150428/#lat=43.2025&lng=46.0894&z=14">screen shot 2017-01-31 at 9 48 43 am</a>

sergiyprotsiv:
From what I saw, here the offsets appear for only one file (Kurchaloy) and it appears to be an OSM issue - comparing the data with Google Maps provides almost no offset.

migurski:
So OSM is in the wrong place, but GMaps aerial is in the right place?

sergiyprotsiv:
Yes, exactly, it seems that the OSM has misaligned streets there. Here is the same area vs Google Maps.
screen shot 2017-02-08 at 21 44 24

migurski:
I'm curious if it's possible to get input from someone on the ground. OSM often warns that aerial imagery might be misaligned, and the only thing to do is show up with a GPS unit locally.

PR #2438, opened 2017-01-27
Nacka, Sweden

sergiyprotsiv:
Municipality of Nacka (a large suburb of Stockholm). From their WFS server, which probably means it will fail when it sees an xml file (https://github.com/openaddresses/machine/pull/523)

### DIFF
--- a/sources/se/municipality_of_nacka.json
+++ b/sources/se/municipality_of_nacka.json
@@ -1,0 +1,13 @@
+{
+    "coverage": {
+        "country": "se",
+        "city": "Nacka"
+    },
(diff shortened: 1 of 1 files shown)


astoff:
You can request a geojson file from the WFS server, as in [br/sp/santos.json](https://github.com/openaddresses/openaddresses/blob/2bd622f147d0e9134dfe1e0d60f6e101a219e68e/sources/br/sp/santos.json).

sergiyprotsiv:
Thanks a lot @astoff, this worked great!

PR #2480, opened 2017-02-02
adding postcode field to Kharkiv, Ukraine

albarrentine:


migurski:
Hm, preview has that in the wrong place. Might be misprojected.

thatdatabaseguy:
Didn't change the source projection, but these lat/lons are in Kenya. Any idea what it should be?

migurski:
Try omitting the SRS entirely, since it's an ESRI source that can reproject.

migurski:
Try omitting the SRS entirely, since it's an ESRI source that can reproject.

thatdatabaseguy:
No luck. The [server](http://cdr.citynet.kharkov.ua/arcgis/rest/services/gis_ort_stat_general/MapServer/1) seems to say it's using EPSG:28467, which was the original SRS (meaning these lat/lons were never right), but, at least if I'm reading the [definition](http://spatialreference.org/ref/epsg/28467/) correctly, the range of coordinate values should be two orders of magnitude larger than the source's bounds: 

```
Extent:
XMin: 8203.805388036104
YMin: 951.5518080910333
XMax: 31482.935049609117
YMax: 25401.013972328936
Spatial Reference: 28467  (28467) 
```

Does that make any sense?

migurski:
This is pretty weird. I wonder if the server is misconfigured, and maybe we don't have a way of overriding this?

thatdatabaseguy:
As suggested by @trescube earlier today, looked up one of the addresses in Google Maps and then ran projections with every known coordinate system in GDAL to see which was closest to the real lat/lon. None of the systems got within 500km, so can be pretty sure this is more than a mislabeled coordinate system.

migurski:
They might have one with custom parameters that's not represented by an EPSG code — I've seen this before in civil engineering projects, but not in a public dataset.

migurski:
[EPSG 28467](http://spatialreference.org/ref/epsg/28467/) is a transverse Mercator centered on Eastern Ukraine, so it’s a logical fit for Kharkiv. The small numbers in the bounds make me think that it’s using a custom x/y offset. This may require some work on Machine to allow for SRS values outside of known EPSG codes.

migurski:
I derived this offset for 28467 through trial and error, but it’s not a good fit: `+proj=tmerc +lat_0=0 +lon_0=39 +k=1 +x_0=214234 +y_0=-5530474 +ellps=krass +units=m +no_defs`

stefanb:
http://spatialreference.org/ref/epsg/pulkovo-1942-gauss-kruger-7n/

iandees:
Rerun this please.

ingalls:
Tracking on new infra here: https://github.com/openaddresses/openaddresses/pull/5569

PR #2481, opened 2017-02-02
switched to ESRI source, added coverage info

trescube:


trescube:
Re-run this please

trescube:
Re-run this please

trescube:
This change clears up the issues from the original source where there were 4600 `0` and 12361 `99999` house number addresses that were invalid.  The original source appears to no longer be updated, too.

PR #2482, opened 2017-02-02
adding postcodes for the state of New York

thatdatabaseguy:


trescube:
The note says we're not supposed to use the `ZipCode` field.

iandees:
I had a phone call with Cheryl Benjamin from New York State on October 31st, 2016 where she told me that they reached agreement with HERE to distribute the zipcodes in this dataset:

> i just got off the phone with Cheryl Benjamin and her colleagues at New York State. she wanted us to know that the zipcodes in their data are open now (they reached an agreement with HERE to distribute them), the data files on their clearinghouse will be updated twice a year and the Esri service will be updated every other week

thatdatabaseguy:
Ha, it does say that. Was looking at a few sources where postcode was omitted. In any case, will add it in now that it's kosher to do so.

thatdatabaseguy:
Hm, not the most helpful log.

Re-run this please.

thatdatabaseguy:
Ok, so got this one to run locally using ```openaddr-process-one``` (the extent of what I have working locally on my research machine). @migurski do the webhooks have a time limit or something?

migurski:
The webhooks don't have a time limit. New York State has always been very unreliable, with timeouts and other failures that make me think they've got a firewall throttling us: https://results.openaddresses.io/sources/us/ny/statewide

migurski:
(Actually, that's not entirely true: the webhooks do give up on a source after 9 hours)

thatdatabaseguy:
Ah, ok. The local version I ran took just a little over 9 hours. Is there a way to increase that limit?

migurski:
Yeah — I can raise the limit for the whole service, maybe to 12 hours? I won't be able to deploy that until the current batch is complete in the next day.

thatdatabaseguy:
Sounds good, no worries, will re-run this then.

iandees:
Closing in favor of #2808.

PR #2483, opened 2017-02-03
Thüringen, Germany

sergiyprotsiv:
Could not pass by this statewide source for Thüringen. No specific license, but is is under the "open data" section of the official geoportal.
The source contains only street and address, city and other information is provided via foreign keys (which I skip to be able to use the source directly). 
No field names, so this is how I am guessing they should work. No specific SRID either, but the source mentions UTM Zone 32, so the first guess is that the SRID is 25832.

### DIFF
--- a/scripts/de/th/process.py
+++ b/scripts/de/th/process.py
@@ -1,0 +1,21 @@
+# coding: utf-8
+
+# Download the file from http://geoportal.geoportal-th.de/hausko_umr/HK-TH.zip
+# Unzip and run from within the unzipped folder
+
+import csv
+import six
+reader = csv.reader(open('schluessel-TH.txt'), delimiter=';')
+cities = {}
+for line in reader:
+    if line[0] == 'G':
+        cities[tuple(line[1:5])] = line[5]
+
+reader = csv.reader(open('adressen-TH.txt'), delimiter=';')
+writer = csv.writer(open('thuringen.txt', 'w'), delimiter=';')
+for line in reader:
+    city_id = line[3:7]
+    city = cities.get(tuple(city_id), '')
+    line.append(city)
+    writer.writerow(line)
+    
--- a/sources/de/th/statewide.json
+++ b/sources/de/th/statewide.json
@@ -1,0 +1,20 @@
+{
+	"coverage": {
+        "country": "de",
+        "state": "th",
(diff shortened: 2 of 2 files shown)


sergiyprotsiv:
Parses well, but the projection is probably way off since the resulting coordinates are nonsensical (or it could be a problem that the original source has commas as decimal separators).

nvkelso:
Good find!

thatdatabaseguy:
Thanks @sergiyprotsiv, huge addition with > 600k addresses!

It would be useful to add city names as well, though they are stored as a foreign key to a separate "table". There's a file included in the ZIP archive called schluessel-TH.txt with rows like:

```
G;16;0;66;065;Schwarza
```

The meaning of the fields depends on the value of the first field. Any record that starts with "G" is "gemeinde" or community/city. The other prefixes are for more coarse entities like "L" for "Land" (state), etc. In the case of rows starting with "G", the next 4 fields form a unique ID which can be joined to adressen-TH.txt. There the records look like the following:

```
N;DETHL57P0000FVGJ;A;16;0;66;065;0000;00018;13;;32608487,445;5609511,790;Bei St. Bartholomäi;;;;
```

The combination of the 4th, 5th, 6th, and 7th fields can be joined to the 2nd, 3rd, 4th, and 5th fields of schluessel-TH.txt to get the cities. If you're able to upload a denormalized CSV to e.g. Dropbox, we can copy it over to the cache and use that as the source.

thatdatabaseguy:
If this PR were on the main repo I'd just push to this branch, but here's a quick Python script that can be checked in to obtain the cities. Creates a new semicolon-delimited CSV called "thuringen.txt" with city name in the last column (column 19).

```python
# coding: utf-8
import csv
import six
reader = csv.reader(open('schluessel-TH.txt'), delimiter=';')
cities = {}
for line in reader:
    if line[0] == 'G':
        cities[tuple(line[1:5])] = line[5]

reader = csv.reader(open('adressen-TH.txt'), delimiter=';')
writer = csv.writer(open('thuringen.txt', 'w'), delimiter=';')
for line in reader:
    city_id = line[3:7]
    city = cities.get(tuple(city_id), '')
    line.append(city)
    writer.writerow(line)
```

sergiyprotsiv:
Thanks @thatdatabaseguy, I have updated the source accordingly.

PR #2452, opened 2017-01-31
Add churchill county nv

trescube:

PR #2484, opened 2017-02-05
added experimental test support in source + test driver

trescube:
## Embedding Tests Within Sources

While adding sources recently, I've noticed that a lot of the more complicated regexs being introduced are hard to maintain unless the person modifying them is the original author who has combed through the data.  In openaddresses/openaddresses#2452, I started adding examples to illustrate some of the inputs I came across for future reference.  This is useful but, while I'm constructing the regexs, I use a primitive groovy script to test some inputs which is helpful at that moment but doesn't preserve the history in a reproducible way. This PR proposes to embed tests with inputs and expected values directly within the source and provides a script written in node to exercise these tests.

### Tests

Source tests are listed in a `test` object that's a child of the root and can have 3 elements:

* `enabled`: boolean, defaults to `true`, skips all tests if set to `false`
* `description`: overall theme of the tests
* `tests`: array of test cases

Each test case in `tests` has 3 elements:

* `description`: what specific behavior this test is capturing
* `inputs`: the fields that match the source fields that the `conform` properties reference
* `expected`: the expected values for the `conform` fields

### Example

I'll use the following source test as an example:

```
{
    "description": "address with /-delimited number",
    "inputs": {
        "alternativ": "Jiřská 9/10, Hradčany, 11900 Praha 1"
    },
    "expected": {
        "number": "9/10",
        "street": "Jiřská",
        "city": "Praha 1",
        "postcode": "11900"
    }
}
```

In the Czech source, number, street, postcode, and city are all in a single field, so there's a single input value which matches the field name that the `conform` functions are looking for.  The `expected` block operates how one would expect tests to operate:  the extracted `number` value should be `9/10`, the extracted `street` value should be `Jiřská`, and so on.

### Test Driver

Since this is just a proof of concept, I wrote the driver in node which matches the behavior of the machine when dealing with the regex matcher. Currently the driver only looks for `conform` fields that employ the `regexp` function since that's where the majority of complexity lies. I don't know python beyond the basics of modifying the machine but I'm open to writing this testing framework in python to match the machine regex flavor.

Where the test driver lives is up for debate.  For my personal development style, it's very useful to be able to get instant feedback locally on changes to regexs as I find interesting inputs in the source data.  That is, it's cumbersome to have to commit the code to github, wait for travis to finish processing the entire source, the comb thru the results looking for what worked and what didn't.  However, I can also see the value of adding this as a pre-processing step in the machine (like the current node tests) to do some quick validations before embarking on processing the
entire source.

I used [tape](https://www.npmjs.com/package/tape) for testing in the driver.  Each test in `tests` prints a header with fairly descriptive messages:

```
# testing 'address with /-delimited number'
ok 1 number field: expected '9/10' from input 'Jiřská 9/10, Hradčany, 11900 Praha 1'
ok 2 street field: expected 'Jiřská' from input 'Jiřská 9/10, Hradčany, 11900 Praha 1'
ok 3 city field: expected 'Praha 1' from input 'Jiřská 9/10, Hradčany, 11900 Praha 1'
ok 4 postcode field: expected '11900' from input 'Jiřská 9/10, Hradčany, 11900 Praha 1'
```

If a test fails, a helpful information block is printed:

```
# testing 'address with special-case house-only identifier 'č.ev.''
not ok 17 number field: expected 'č.ev. 7/3a' from input 'Albertov č.ev. 7/3a, Nové Město, 12800 Praha 2'
  ---
    operator: equal
    expected: 'č.ev. 7/3a'
    actual:   'č.ev. 7/3'
    at: Test.<anonymous> (/Users/stephenhess/git/trescube/openaddresses/test/regexp_tests.js:52:39)
  ...
ok 18 street field: expected 'Albertov' from input 'Albertov č.ev. 7/3a, Nové Město, 12800 Praha 2'
ok 19 city field: expected 'Praha 2' from input 'Albertov č.ev. 7/3a, Nové Město, 12800 Praha 2'
ok 20 postcode field: expected '12800' from input 'Albertov č.ev. 7/3a, Nové Město, 12800 Praha 2'
```

By running each test as a separate tape test, it's easy to navigate to any failing tests.

### Running It

To run the driver, simply type `node test/regexp_tests.js sources/cz/countrywide.json`.


migurski:
Nice! 👏

migurski:
Thinking a bit more about where this could live. Since it doesn't require running the whole source, it seems possible to run these tests quickly and outside of the full webhook process, but maybe using the same Machine code that does the actual work so that the test tests what we expect it to test? This is really cool.

migurski:
@trescube and I chatted by phone today. He’s going to make some terminology tweaks, but we think that this approach will be really useful to include in the Machine code. I'll add support for it once this source is in master, and then follow up in https://github.com/openaddresses/machine/issues/544.

PR #2487, opened 2017-02-07
Add us coverages

trescube:


trescube:
Re-run this please

migurski:
I think there’s a bunch of copies of Milwaukee running. It can take [6 hours to cache](http://results.openaddresses.io/sources/us/wi/city_of_milwaukee), so this might take a while.

trescube:
Re-run this please

PR #2492, opened 2017-02-07
Novosibirsk, Russia

sergiyprotsiv:
Data from their official ESRI server, without license information. The address data is in a "table" and is linked to buildings with geometry, so it was a clunky custom pyesridump script to download it. Let me know if you want that uploaded too.

Probably will not work straight away since csv requires lat/lon and this has WKT building geometry. Can get centroids instead of buildings, or maybe there is a way to provide WKT to the machine?


sergiyprotsiv:
I have added the centroid coordinates to the file and kept the WKT polygons of the houses in case they are needed in the future.

sergiyprotsiv:
Re-run this please.

thatdatabaseguy:
This looks like it might be the city of Novosibirsk, not the oblast, but the addresses look good. The build failure looks like something we fixed a while back in master with JS parsing Python regexes. Maybe try merging in master?

sergiyprotsiv:
It's a bit in the middle: it's much larger than the city itself, but only covers the Eastern third of the region (though the vast majority of the population): http://loi.sscc.ru/bdm/nso/big/dorogi/road_oblast.jpg

thatdatabaseguy:
@trescube any ideas?

sergiyprotsiv:
Re-run this please.

Issue #2440, opened 2017-01-27
Bangkok, Thailand


sergiyprotsiv:
There is a source of house numbers in Bangkok: http://cmc.bangkok.go.th/arcgis/rest/services/FGDS4000T/MapServer/5

It has a solid 2 million houses, most with house number. It does not have the street name or an identifier to link with a street table. The geometry for streets is there, however: http://cmc.bangkok.go.th/arcgis/rest/services/FGDS4000T/MapServer/7

What's the best way?
- Import only the house numbers and let downstream users link them to streets
- Create a (likely error-prone) script to match houses to streets by proximity
- Skip the source and wait for a better one

nvkelso:
The related Pelias project has street number - street name interpolation. I'm in favor of adding as a simple import now and if better data comes along later we can switch over.

@iandees and @trescube what's your opinion?

nvkelso:
@iandees Are you okay adding street number (missing street name) address to the project? Are there examples of this today?

iandees:
I think we require both street name and street number right now. There aren't any examples of a source missing street name (at least on purpose) right now.

If there's no way we can automatically give these points an attribute that's close to street name I'd prefer if we skip this data for now.

PR #2493, opened 2017-02-07
Bangkok, Thailand

sergiyprotsiv:
House number only source from the city of Bangkok's ESRI server (https://github.com/openaddresses/openaddresses/issues/2440). No license info.


nvkelso:
Dense!

slibby:
I _think_ that the Travis build is failing from a missing `street` in the conform object, which I believe is a requirement. @migurski can you validate that? The Travis failure log is pretty obscure

migurski:
Yuck, yeah, that is super obscure. A missing "street" is a problem though!

sergiyprotsiv:
Yeah, the missing street discussion was started here: https://github.com/openaddresses/openaddresses/issues/2440
Basically, it could be fine for Pelias, but not necessarily for other users.

iandees:
I'm going to close this for now. I think the next step here if someone wants to take it is to build a script that joins with the centerlines layer in the same map service.

PR #2522, opened 2017-02-12
updated Stearns, MN, USA to points

trescube:


bigbadboss:
Wat is this and how do I use this

trescube:
The message was you forget `compression: zip` is very confusing.

PR #2526, opened 2017-02-15
Update & fixes for us/ky/kenton

davidchiles:
Changes:
1. Fixed Kenton `geoid`.
2. Switched to address endpoint.


trescube:
Thanks!  And good catch on the geoid!

PR #2529, opened 2017-02-16
Fix street field in Pickens, SC

thatdatabaseguy:
The STREET field is a fully qualified street name with thoroughfare type and direction.


trescube:
D'oh!  Good catch, and with no name-only field for street, that's some strange data needs they have.

PR #2539, opened 2017-02-17
added city of Sioux Falls, SD, USA

trescube:


thatdatabaseguy:
There's a HOUSENUM field, regex shouldn't be necessary.

PR #2541, opened 2017-02-18
added coverages to Ukranian cities

trescube:


sergiyprotsiv:
Kharkiv remains in Kenya due to a "special" SRID

PR #2542, opened 2017-02-18
Bogotá

astoff:
I am not familiar with Bogotá's addressing scheme, which seems a bit peculiar, so it might be good to double check this. In particular, I don't know if the street name abbreviations are in common use, but ideally they should be translated according to this table:

CL Calle
DG Diagonal
AC Avenida calle
KR Carrera
TV Transversal
AK Avenida carrera
CRT Carretera

Also, the terms of use demand:
- [...] brindar el apropiado reconocimiento o crédito de titularidad al LICENCIANTE, contemplando como mínimo el nombre del titular y la fecha de la versión de los DATOS ABIERTOS utilizados. Todo ello con sujeción a las normas que regulan el derecho de autor.
- Conceder los DATOS ABIERTOS en los mismos términos y condiciones de uso en que le fueron otorgados originalmente en la presente LICENCIA ABIERTA.

Does the second point implies share-alike? It's also not clear to me whether we fulfill their request to acknowledge the download date. Perhaps the fact that one _can_ trace back the download date is enough? (There are also "static" shapefiles to download in the website, we could use those instead.)

### DIFF
--- a/scripts/co/bogota.py
+++ b/scripts/co/bogota.py
@@ -20,6 +20,8 @@ try:
     feature = next(features)
 
     while True:
+        if not feature.get('geometry') or feature['geometry']['coordinates'][0] == 'Nan':
+            continue
         props = feature['properties']
         interior = props['PDoCInteri']
         if interior and interior.strip():
--- a/scripts/co/bogota.py
+++ b/scripts/co/bogota.py
@@ -25,9 +25,9 @@ try:
             continue
         props = feature['properties']
         interior = props['PDoCInteri']
-        if interior and interior.strip() and not interior.strip()[0].isdigit():
+        if interior and interior.strip():
             street = props['PDoTexto']
-            if street and street.strip():
+            if street and street.strip() and not street.strip()[0].isdigit():
                 props['PDoCInteri'] = street
                 props['PDoTexto'] = interior
         outfile.write(json.dumps(feature))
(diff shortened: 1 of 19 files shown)


thatdatabaseguy:
Great find! Have been looking for a Colombian data set.

License does appear to have a share-alike clause.

In general OpenAddresses does not try to expand abbreviations, and from what I've seen addresses in Colombia are often written using these standard abbreviations anyway, so fine to leave them.

There are many numbers in Bogotá addresses, so something like "KR 57 # 23-15" means "Carrera 57 cross street Calle 23, the building 15 meters from the intersection", though in practice the "23-15" is used as the house number.

astoff:
@thatdatabaseguy do you agree with mapping the field "codigo del interior" to unit? It's defined as "texto que identifica la placa domiciliaria asociada a los Interiores, Bodegas, Mejoras, Locales, Casas"

thatdatabaseguy:
In theory yes, the "numero interior" should be the apartment number, but here for whatever reason when "PDoCInteri" is not null it appears to be the house number and "PDoTexto" is the unit. When it is null, "PDoTexto" is the house number.

Tested this by looking up a few of the coordinates in OSM and Google Maps, and seems to hold true (although streets in Colombia are not always entirely up-to-date).

The machine conforms don't support that kind of conditional, so uploading a script and (after it finishes) a cached file which should resolve this.

thatdatabaseguy:
Re-run this please.

thatdatabaseguy:
This happened with the Kuwait data set as well. The build finished and was successful but the Github status checks still say pending.

Re-run this, please.

PR #2552, opened 2017-02-20
Medellín, Colombia

astoff:
This source contains a field called `INTERIOR` which I believe should be mappedto unit, but I am not sure.  Also, the house number field (`PLACA`) includes the contents of the `INTERIOR` field at the end, inside parentheses, so it should probably be regexed away.

The website mentioned in the source has an "open data" section, but it does not seem to specify the terms of use.  Moreover, I am not sure this source can actually be directly accessed from that section of the website.

### DIFF
--- a/scripts/co/medellin.py
+++ b/scripts/co/medellin.py
@@ -1,0 +1,34 @@
+import json
+import math
+import sys
+import logging
+
+from esridump.dumper import EsriDumper
+
+logging.basicConfig(level=logging.DEBUG)
+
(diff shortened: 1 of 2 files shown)


thatdatabaseguy:
Re-run this, please

thatdatabaseguy:
Re-run this, please.

thatdatabaseguy:
Re-run this, please

astoff:
Re-run this, please

thatdatabaseguy:
Ugh, this server is really flaky.

Re-run this please.

PR #2567, opened 2017-02-23
add Kauai, HI

thatdatabaseguy:


trescube:
You beat me to it!  Do you think `UNIT_NO` is useful?  Where available it's also part of the house number and I can't tell from google maps if it's legitimately part of the house number or it's actually a true unit.

trescube:
I spot-checked a few and they're standalone houses.  There don't appear to be any duplicate house numbers (just the digits) among those with units.

thatdatabaseguy:
Ok, looked like they were actual unit numbers except for the lettered units, but can remove them.

trescube:
Oh, given the sheer number of alpha-only `UNIT_NO` values, I didn't see the digits-only ones.  Those appear to be legit units and backed up by [street view](https://www.google.com/maps/place/4643+Hauaala+Rd,+Kapaa,+HI+96746/@22.0830797,-159.3165792,3a,60y,208.72h,65.93t/data=!3m6!1e1!3m4!1s2Frb1rrBStQHIkatjTkprQ!2e0!7i13312!8i6656!4m5!3m4!1s0x7c06e0edce8b4911:0x67dd278345458656!8m2!3d22.082798!4d-159.3166029!6m1!1e1).

trescube:
Sorry, should've been more explicit, the digits-only units appear to be legit units.  Given the number of records with 'CONDOMINIUM' in the name, it's probably safe to assume that most are legit units so this is probably good as-is.

PR #2568, opened 2017-02-24
Add Monroe County, Michigan

oini:
Added Monroe County data found in https://cityofmonroegis.monroemi.gov/Apps/GISViewer/ over this ESRI server -> https://cityofmonroegis.monroemi.gov/arcgis/rest/services/Production/Parcels/Mapserver/0 to `sources/us/mi/monroe.json`

cc @ramyaragupathy @msirenko @ingalls @migurski

### DIFF
--- a/sources/us/mi/city_of_monroe.json
+++ b/sources/us/mi/city_of_monroe.json
@@ -1,13 +1,20 @@
 {
     "coverage": {
+        "geometry": {
+            "type": "Point",
+            "coordinates": [
+                -83.39,
+                41.92
+            ]
+        },
         "US Census": {
-            "geoid": "26115",
-            "name": "Monroe County",
+            "geoid": "2655020",
+            "name": "City of Monroe",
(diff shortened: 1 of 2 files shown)


iandees:
This looks like the City of Monroe, Michigan, not the county. Take a look at [this example](https://github.com/openaddresses/openaddresses/blob/master/sources/us/tn/city_of_nashville.json) of a city source. We seem to have settled on these differences for city-based sources:

- The filename for cities begin with `city_of_` (whereas counties are just the name of the county)
- We add a `geometry` key to the coverage object. In most cases its just a single point centered over the city, but in some cases we add a polygon
- For US cities we use the US Census's "place" geoid. I look for that on Census Reporter: e.g. https://censusreporter.org/profiles/16000US2655020-monroe-mi/. The geoid (`2655020`) is the part in the URL after the `US` but before the dash.

iandees:
Great! Thanks for the contribution.

PR #2591, opened 2017-02-26
Baha California Sur - Full addresses from INEGI

msirenko:
First test of an entire mexico import


sergiyprotsiv:
Solved by https://github.com/openaddresses/openaddresses/pull/2743

PR #2626, opened 2017-02-28
adding Gwinnett County, GA

thatdatabaseguy:


thatdatabaseguy:
Gwinnett County contains Lawrenceville, home of Migos.

332k addresses (both the bad and the boujee). Looks good.

PR #2629, opened 2017-03-01
adding Henry County, GA

thatdatabaseguy:
Taking a guess on the coordinate system.


iandees:
Looks like the SRS should be `epsg:2240`.

iandees:
The .prj file doesn't seem to result in a correct projection in QGIS at least. I had to set the layer SRS manually by searching based on the strings in the .prj file.

iandees:
Or maybe I'm just doing something wrong in QGIS :).

thatdatabaseguy:
It seems like for shapefiles there's some metadata that specifies the projection. That map looks more or less correct though a spot check of the addresses seems to have them a block or two off of OSM/Google. Will try epsg:2240 to see if that's closer.

thatdatabaseguy:
Hm same result, but map looks good.

PR #2630, opened 2017-03-01
Add Kalamaria, Thessaloniki, Greece

thatdatabaseguy:
Found this on the OSM sources list. It's CC-BY, first source in Greece, basically a borough of Thessaloniki. ~10k address points but a start.


nvkelso:
Nice!

PR #2649, opened 2017-03-05
Inconsistencies in Italy

sabas:
Moved some files in the correct directory per standard, and updated the json fields.
Some files need to be renamed based on their municipality coverage (instead of county)


sabas:
The biella file need to be renamed as well

PR #2665, opened 2017-03-09
switch South Tyrol source to csv format

tyrasd:
Shapefile generation seems to [time out constantly](http://results.openaddresses.io/sources/it/32/south-tyrol), while csv output appears to works fine.

### DIFF
--- a/sources/it/32/south-tyrol.json
+++ b/sources/it/32/south-tyrol.json
@@ -19,6 +19,7 @@
     "conform": {
         "accuracy": 1,
         "type": "csv",
+        "encoding": "ISO-8859-15",
         "srs": "EPSG:25832",
         "lat": "GEHA_COORD_Y",
         "lon": "GEHA_COORD_X",
--- a/sources/it/32/south-tyrol.json
+++ b/sources/it/32/south-tyrol.json
@@ -20,6 +20,8 @@
         "accuracy": 1,
         "type": "csv",
         "srs": "EPSG:25832",
+        "lat": "GEHA_COORD_Y",
+        "lon": "GEHA_COORD_X",
         "number": {
             "function": "join",
             "fields": ["GEHA_NUMMER", "GEHA_NENNER"],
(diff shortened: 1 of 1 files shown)


trescube:
@migurski I can't tell what's going wrong with this source.  It looks like it may be trying to decompress an uncompressed file.

migurski:
https://s3.amazonaws.com/data.openaddresses.io/runs/164351/output.txt just indicates a unicode encoding problem.

tyrasd:
> unicode encoding problem

thanks, didn't think about the encoding being different in the csv output.

trescube:
Thanks, looks good!

migurski:
There are a lot of doubled-up streetnames here, slash-delimited in Italian and German. Maybe this could be two sources with different regexps?

screen shot 2017-03-12 at 11 17 38 am

tyrasd:
Hmm. This region is bilingual (German/Italian, see [wikipedia](https://en.wikipedia.org/wiki/South_Tyrol#Languages)). That's why I've originally put both names into the street/city fields using [joins](https://github.com/tyrasd/openaddresses/blob/264fb97474542a2a526bea4ebb9b0914485dd22d/sources/it/32/south-tyrol.json#L31-L46).

But I think you're right. This should better be mapped as two distinct data sets. That then give a 2x too high housenumber count for the region, but that shouldn't be an issue, right? Only when one does reverse geocoding it can then be difficult to find out which of the variants to choose… there isn't a metadata-field where one can put the language of the source, is there?

migurski:
It’s not an issue, no — we do something similar for data in Finland which also comes in two language variants. [Language is an optional tag you can add](https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#optional-tags).

trescube:
Pelias (mapzen search - a user of OA data) could certainly take advantage of multiple languages when available.

tyrasd:
:+1: (I've totally overlooked the language field :see_no_evil:)

I've got one more language related question, though: in this data set, some (but not all) addresses even have translations into a third language, [Ladin](https://en.wikipedia.org/wiki/Ladin_language). Would openaddresses be able to skip the addresses where the street and city fields are empty? If yes, I could add the Ladin variant also. Would using the *ISO 639-3* language code `lld` be OK since this language doesn't have a *ISO 639-1* or *ISO 639-2* code?

PR #2668, opened 2017-03-13
create additional source for Ladin in South Tyrol

tyrasd:
follows #2667 and https://github.com/openaddresses/openaddresses/pull/2665#issuecomment-285979481


trescube:
Hey, could you rebase this on master and push?  The error messages are much cleaner now.

tyrasd:
Done.

I think the error is most likely related to this source not having every address translated into this specific language.

tyrasd:
Oh, [this one](https://travis-ci.org/openaddresses/openaddresses/jobs/222796367) looks different. Unfortunately, the [Ladin language](https://en.wikipedia.org/wiki/Ladin_language) doesn't have a two digit ISO 639-1 code, but only a 3 digit ISO 639-3 one. :thinking:

trescube:
Oh, interesting.  Hmm...  @iandees or @migurski is there any reason why `language` should be 2 characters?  According to the [contribution guide](https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md), `language` must be ISO 639-1 so that's how the schema is written but this source would violate that.  Thoughts?

iandees:
@trescube I don't think there's a particular reason we limited it to 2 letters. I adjusted the schema and tests to allow three letter language codes.

Issue #578, opened 2014-10-19
us-ny-saratoga


jalbertbowden:
Saratoga County 
http://www.maphost.com/saratoga/

PR #2676, opened 2017-03-14
Highway addresses in Kelowna, BC add an extra HWY

thatdatabaseguy:
Implementing with regex


migurski:
Testing to see if #578 worked:

Re-run this please.

Issue #2687, opened 2017-03-23
South Korea Address Data


justinelliotmeyers:
after mapping south korea for the past several years, I came across their address data. They are in the process of recreating the entire system I believe. For some reason I thought we had a nationwide dataset, but I was wrong.

The updated data is somewhere in here: http://www.juso.go.kr/addrlink/addrlinkJusoDBMain.do

this stuff gets updated monthly for sure, but may be more frequent.

I am not sure the terms either, but their nation is moving towards open data every day ;)

PR #2688, opened 2017-03-25
Korea countrywide

sergiyprotsiv:
I took a stab at the Korean data suggested by Justin: https://github.com/openaddresses/openaddresses/issues/2687. It's from the Ministry of the Interior. I did not see a clear license, but the Ministry is clearly trying to get everyone to use the addresses - lots of datasets and APIs.

The data seems quite complete, though positional accuracy is hard to judge relative to OSM data, since OSM rarely contains houses. 

There is a description of the address formats here: https://www.juso.go.kr/dn.do?fileName=GuideBook_eng.pdf&realFileName=GuideBook_eng.pdf&regYmd=2016 If someone has better familiarity with address formatting, let me know. There is also official Romanization available, but it has no coordinates, so it will need to be processed offline.

### DIFF
--- a/scripts/kr/korea.py
+++ b/scripts/kr/korea.py
@@ -1,0 +1,23 @@
+import csv
+import sys
+import subprocess
+
+if (len(sys.argv) < 2):
+    sys.exit('Usage: python korea.py file1 [file2 ...]')
+
+filenames = sys.argv[1:]
+
+for filename in filenames:
+    infile = open(filename)
+    outfile = open('{}.out'.format(filename), 'w')
+    reader = csv.reader(infile, delimiter='|')
+    writer = csv.writer(outfile, delimiter='|')
+    for line in reader:
+        building_number = line[9]
+        building_sub = line[10]
+        if building_number and not all((c == '0' for c in building_sub)):
+            building_number = '-'.join((building_number, building_sub))
+        writer.writerow(list(line) + [building_number])
+    subprocess.check_call(['mv', '{}.out'.format(filename), filename])
+    infile.close()
+    outfile.close()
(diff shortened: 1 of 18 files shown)


nvkelso:
Dense!

thatdatabaseguy:
Great data set!

One thing that looks a bit odd is the house number. From [Wikipedia](https://en.wikipedia.org/wiki/Addresses_in_South_Korea) it seems like they should be hyphenated, but in this data set "0" is a NULL value, so they'd have to hyphenated if the number in COLUMN11 is not equal to zero, which machine does not support at present. Best way to handle this is probably with a script. This will do the trick:

```python
import csv
import sys
import subprocess

if (len(sys.argv) < 2):
    sys.exit('Usage: python korea.py file1 [file2 ...]')

filenames = sys.argv[1:]

for filename in filenames:
    infile = open(filename)
    outfile = open('{}.out'.format(filename), 'w')
    reader = csv.reader(infile, delimiter='|')
    writer = csv.writer(outfile, delimiter='|')
    for line in reader:
        building_number = line[9]
        building_sub = line[10]
        if building_number and not all((c == '0' for c in building_sub)):
            building_number = '-'.join((building_number, building_sub))
        writer.writerow(list(line) + [building_number])
    subprocess.check_call(['mv', '{}.out'.format(filename), filename])
    infile.close()
    outfile.close()
```

House number is stored in COLUMN19.

thatdatabaseguy:
LGTM, > 6M addresses in :kr:!

migurski:
👏👏👏

iandees:
Wow this is great! Thanks for working on the sources.

msirenko:
@sergiyprotsiv amazing!

justinelliotmeyers:
Thanks everyone for your help with this!!!

PR #2689, opened 2017-03-26
Create sibley.json

mmdolbow:
This is a parcel service but it appears to have true situs address columns to use, with almost 15K records for the county. Hope it works well.


nvkelso:
LGTM! This area is south-west of Minneapolis–St.-Paul.

iandees:
Thanks Mike!

Issue #2593, opened 2017-02-27
Full (much better) Mexico Coverage from INEGI


msirenko:
Posting this here for visibility. In short, we found a version of the Mexico INEGI address dataset with ~20x more address points - enough that this looks like the comprehensive coverage list. It seems like the old ~1.5M address shapefiles were business addresses or similar

Blockers
- [ ] Machine doesn't look like it handles zipped directories 
- [ ] There are 10,000 files we're sourcing this data from, so we should do this in parts so we don't DDOS the INEGI server

Current thread: https://github.com/openaddresses/openaddresses/pull/2591

cc @ingalls @iandees @justinelliotmeyers @migurski @sbma44 

Special thanks to freelancers @blacktopol & @TedYav for helping us find + JSONify this

[**Mexico_JSONs.zip**](https://github.com/openaddresses/openaddresses/files/802896/Mexico_JSONs.zip)

iandees:
To be clearer about the few levels of zip-ception here:
- The linked .zip file above points to a zipped directory of source JSON objects.
- These source objects point to HTTP URLs for zipped shapefiles
- Some of the URLs point to zipfiles that contain zipfiles (which contain shapefiles)

My suggestion would be to build an external script (to be put in [`/scripts`](https://github.com/openaddresses/openaddresses/tree/master/scripts)) that downloads all these shapefiles and generates a CSV for the whole country. The resulting CSV would then be zipped and uploaded to our S3 bucket and there would have a single source in this repo referencing that object.

My opinion might change based on two things:
- How often are these shapefiles updated? If they're individually updated on a periodic basis (weekly? monthly?) it might make sense to try and split this up into smaller pieces so the updated data can be more pro-actively loaded. If it's not updated (this is a one-time dump) or the whole dataset is updated on a e.g. quarterly basis, then running this whole thing at once makes sense to me.
- Do the shapefiles all contain different metadata? If they're the same (my spot-checking indicates they are) then doing the whole thing at once makes sense to me. If not, we'll need to re-think.

Finally: whatever script we create to generate the output should simply convert from shapefile to CSV and not attempt to do any mapping between shapefile attribute names to OpenAddresses attribute names. The source object's conform attribute will do that heavy lifting so we can adjust it later.

migurski:
I think I agree with Ian. This source sounds similar to Australia's GNAF, which we also handle quarterly with a custom download process in `scripts/au`.

sbma44:
[...]

PR #2696, opened 2017-03-28
Mexico Countrywide

sergiyprotsiv:
A first attempt to load the countrywide Mexican data. Almost 30 million addresses - lets see if the machine can handle this.

I am attaching the script that relies on the JSON file linked to in https://github.com/openaddresses/openaddresses/issues/2593 to get the file names necessary for downloading. When an updated dataset is released, the INEGI website will need to be scrapped again for the file links. The updates are supposed to happen every five years (they are related to the census), but so far this is the only one I found.

The script recursively imports the shapefiles from zips and then from zips within zips, etc. The internal structure is not uniform (sometimes there is a shapefile in a folder, sometimes directly in a zip, sometimes in folder in a zip within a zip, etc). The resulting combined shapefile is 32 GB, but the csv is more manageable (1 GB compressed). I have put it on Dropbox for now, since the S3 upload limit is 256 MB. Can be split up into statewide sources if necessary.

A couple of things to note:
- The datasource as it is now only merges the data that is actually within the shapefiles. This means that the metadata on the municipality is lost (as this is only known when browsing the web). It can definitely be added as an extra column at the expense of a more processed source (and the need to process the shape data in Python instead of relying on command-line OGR). Should be straightforward, let me know if this is better.

- Houses in rural towns rarely have the house number. In the files this is denoted as "SN". My understanding is that the relevant information then is some identifier for the block the house is in, particularly the two intersecting streets (one before and one after). The data on these intersecting streets are actually in the file as well - the question is what is the best way to model it (if necessary).

### DIFF
--- a/scripts/mx/process_inegi_census_files.py
+++ b/scripts/mx/process_inegi_census_files.py
@@ -1,0 +1,122 @@
+import json
+import requests
+import zipfile
+import os.path
+import time
+
+
+json_dir = "./output"
+cache_dir = "./cache"
+tmp_dir = "./tmp"
+
+
+# Returns the path where a file should be cached
+def cache_file_path(url):
+	return os.path.join(cache_dir, os.path.basename(url))
+
+# Returns True if we have a valid (i.e. completely downloaded) file already cached. If a file is damaged it is removed.
+def file_already_cached(url):
+	local_path = cache_file_path(url)
+	
+	if not os.path.exists(local_path):
+		return False
+	
+	file_ok = True
+	try:
+		with zipfile.ZipFile(local_path, 'r') as zip_file:
+			file_ok = zip_file.testzip() is None
+	except:
+		file_ok = False
+	
+	if not file_ok:
+		os.remove(local_path)
+		
+	return file_ok
+	
+# Downloads a file to cache
+def download_file_to_cache(url, chunk_size):
+	response = requests.get(url, stream=True)
+	with open(cache_file_path(url), "wb") as cached_file:
+		for chunk in response.iter_content(chunk_size = chunk_size):
+			cached_file.write(chunk)
+
+
+# First we get the URLs of necessary files to be downloaded
+# Currently using the JSON files pre-populated by Mapbox, would normally need to scrap the INEGI website for links
+
+
+relevant_urls = set()
+
+for dp, dn, fn in os.walk(json_dir):
+	for file_name in [ fi for fi in fn if fi.endswith(".json") ]:
+		file_path = os.path.join(dp, file_name)
+		
+		if os.stat(file_path).st_size == 0:
+			continue
+		
+		with open(os.path.join(dp, file_name)) as json_file:
+			data = json.load(json_file)
+    		relevant_urls.add(data["data"])
+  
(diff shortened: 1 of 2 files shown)


sergiyprotsiv:
Ok, the import succeeded, but the projection is wrong and is in fact non-standard EPSG, so the best bet would be to convert to EPSG:4326 when processing the files.

Will launch another run overnight. Any preferences on splitting by state and adding a municipality column?

migurski:
Splitting by state might be unnecessary — Australia is another country that we have in one big pre-processed file, and it’s been fine.

sergiyprotsiv:
Thanks! Any idea why the Travis build fails? I merged the current changes from master, and the build still does not work.

migurski:
I’m not really sure — the error message is pretty useless. I do see that the projection is in the "src" tag, but it should actually be "srs": https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#processing-tags

nvkelso:
Awesome!

The municipality and cross street are interesting to me.

@migurski and @iandees what's the best way to indicate the cross street (intersection) addresses?

migurski:
I'll try to get to the bottom of the atravis check. The data looks great!

sergiyprotsiv:
All of my recent PRs have the same issue, so it could be related to a somehow outdated fork on my side. I tried pulling everything from upstream master and pushing it into my branches, but it does not seem to work.

migurski:
Your stuff looks up-to-date to me! I am fairly sure this is not your fault, but I don’t have an up-to-date Node environment locally that would allow me to test this. Sorry, it might take a little while.

iandees:
The tests pass locally for me. Not sure what Travis's deal is.

iandees:
Oh, they fail now that I merged. I must not have been on the right branch when I tested locally.

iandees:
I am very displeased with the output of our node-based testing framework. I'm going to let this stay broken for now and rewrite our tests in Python with a test framework that gives meaningful output.

sergiyprotsiv:
This was not quite ready to be merged though, will upload a new file with municipalities later on (and we need to figure out what to do with neighboring streets).

iandees:
Sorry, I thought I saw you were ready. You have merge rights anyway, right? I should just let you merge when ready :-)

sergiyprotsiv:
No worries! I also found that there were a few sources missing in the JSON, so will do a quick scrapper to download the missing file

trescube:
Hey, now you know how I feel trying to decipher machine output when something goes wrong!  :)  I appreciate @migurski's efforts to make error messages more meaningful.  

The [schema](https://github.com/openaddresses/openaddresses/blob/master/schema/source_schema.json#L76-L78) is fairly loose with it's definition of coverage (I based it off of https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#coverage-object) and the only required field is `country`.  That error message is not helpful, that's for sure.  I fixed this in another [PR](#2725) or the error will happen anytime the machine runs a source since `npm test` validates every single source.

trescube:
@sergiyprotsiv since you're still adding things, please note the change I made in #2725

PR #2697, opened 2017-03-29
added more fields

trescube:
This source was missing the post-directional on `street`.  Additionally, the `PHYSICAL_STREET_SUFFIX` field appears to contain unit information.  I used the `join` function for `unit` to hopefully identify correct since in the data I've seen only one of `PHYSICAL_SUITE` and `PHYSICAL_STREET_SUFFIX` are populated.

The source times out quite a bit so the data may have to be cached.

### DIFF
--- a/sources/us/az/maricopa.json
+++ b/sources/us/az/maricopa.json
@@ -9,17 +9,23 @@
         "state": "az",
         "county": "Maricopa"
     },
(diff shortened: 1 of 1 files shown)


trescube:
Re-run this please

trescube:
Re-run this please

trescube:
Holy crap it actually worked.

PR #2716, opened 2017-03-29
fixing street name field capitalization in Brevard County, FL

thatdatabaseguy:
street name was missing, changing /SiteStreetName/SiteStreetname/


trescube:
Good catch!

PR #2726, opened 2017-03-30
Cleanup node test error conditions/messages

trescube:
The [recent kerfuffle](https://github.com/openaddresses/openaddresses/pull/2696) over node tests revolve around how `tape` was being used to validate that no errors occurred.  The error stems from this line:

```
t.notOk(validate.errors, `${source}`);
```

Normally this is okay when `errors` is undefined and is even okay for tape when not undefined but produces unparseable input to tap-dot.  I switched this to:

```
t.ok(valid, `${source}: ${JSON.stringify(validate.errors)}`);
```

and all is fine now.  Here's the more meaningful error message that missing country in coverage would've produced:

```
  4053 tests
  2122 passed
  1 failed

  Failed Tests:   There was 1 failure

    x sources/mx/countrywide.json: [{"keyword":"required","dataPath":".coverage","schemaPath":"#/required","params":{"missingProperty":"country"},"message":"should have required property 'country'"}]
```

This should work and create more meaningful error messages until the testing framework gets converted to python.


iandees:
Thanks for looking into this, @trescube!

PR #2743, opened 2017-03-31
Mexico update

sergiyprotsiv:
This has two main changes:
- The list of files to download is now (laboriously) scrapped from the server, so we do not rely on the list provided before. This adds flexibility for future updates and 19 files that were missing in the previous version (so the total number of addresses in this source is > 30M).
- The conform also contains data on state/municipality/locality that was merged in based on another data source (details in the scripts file).

If this passes tests, should be free to merge. The only remaining issue would be to decide on whether to do anything about the intersecting streets.


nvkelso:
I would split intersections off into its own followup issue.

PR #2766, opened 2017-04-01
Update and rename sources/uy/mo/montevideo.json to sources/uy/montevi…

astoff:
…deo.json

Adding ISO 3166

### DIFF
--- a/sources/uy/mo/regionwide.json
+++ b/sources/uy/mo/regionwide.json
@@ -1,5 +1,12 @@
 {
     "coverage": {
+        "geometry": {
+            "type": "Point",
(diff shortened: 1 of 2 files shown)


astoff:
Sure, I can do that. I changed this because I noticed the source covers the entire province.

The shapefile is corrupted right now, but it usually works. Anything against merging even if the test fails?

astoff:
@iandees do you mean to keep geometry and ISO 3166 objects in parallel, as in the current version?  I renamed the file to uy/mo/regionwide.json because the Montevideo department is actually slightly larger than the city itself.

Issue #179, opened 2014-05-02
Belgium


ingalls:
http://data.belgium.be/

ingalls:
This got us the flanders dataset

katossky:
There is also a data set for Wallonie, to arrive soon. Maybe you can have an early access by mailing to icar@spw.wallonie.be. I can assist for the French if needed.

The documentation is [here](http://geoportail.wallonie.be/georeferentiel/icar) (albeit in French) and the soon soon-to-come data will be available [here](http://geoportail.wallonie.be/catalogue/2998bccd-dae4-49fb-b6a5-867e6c37680f.html). But [the licence](http://geoportail.wallonie.be/files/documents/ConditionsSPW/DataSPW-CGU.pdf) does not seem open-data friendly at all (my translation):

> The user ... can not distribute the data to a third party ; can not publish the data on the Internet via a web service ; can publish it as an static image.

nvkelso:
Is this a generic open data license that they use, or something draft for
the draft dataset that they are taking public comment on?

katossky:
It's difficult for me to understand. The licence is spread across 3 documents:

- 1 general "access" licence
- 1 specific "access" licence, restricting application of the previous one
- 1 "user" licence (the one that I mention in my previous comment)

katossky:
I sent a mail to ask for more information, with copy to OpenAddress.

sergiyprotsiv:
@katossky, I have preliminary data here: https://github.com/openaddresses/openaddresses/pull/2768
Let me know if there are any news on the license.

katossky:
I am in discussion with them. They ask who is behind OpenAdresses. I do not know what to answer.

Since the licence authorizes a special agreement with the data producer SPW, I guess they want to know with who they are going to agree anything. Fore sure they can't sign anything with a Github account :)

Please help!

iandees:
I could probably sign something. What are their terms?

katossky:
They don't have terms yet. They ask "Who are we talking with?".

katossky:
That's what I was gonna say, but I am afraid they would freak out :)

Do you read French? You're the one behind openaddresses@gmail.com, right? I always copy e-mails to that address.

katossky:
I replied, with copy to @ingalls and @iandees, in both French and English. I'll keep you updated.

iandees:
[...]

PR #2768, opened 2017-04-02
Wallonia, Belgium

sergiyprotsiv:
Here is the link to the data on the region of Wallonia in Belgium (which would complete the Belgium coverage) (referenced in https://github.com/openaddresses/openaddresses/issues/179).

Unfortunately the license as is remains unclear. Or rahter the license is pretty clear that it cannot be published online (http://geoportail.wallonie.be/files/documents/ConditionsSPW/DataSPW-CGU.pdf), but unclear that it refers to this data since there are three files described here: http://geoportail.wallonie.be/catalogue/b795de68-726c-4bdf-a62a-a42686aa5b6f.html

I added skip:true for now, can always switch it back later on.


katossky:
I am in discussion with them right now, see #179 .

iandees:
I'm not comfortable with the license situation here, so I'm going to close.

PR #2808, opened 2017-04-04
Update New York State file

iandees:
Two changes here:

- The source data switched to a zipped GDB that the New York State GIS folks sent me. It explicitly has New York City carved out of the dataset. They will be sending it to me quarterly.
- A few months ago the New York State GIS team negotiated an agreement with Here to allow the zipcodes that are part of this dataset to be free and open, so I removed the note and added back in the zipcodes.

### DIFF
--- a/sources/us/ny/statewide.json
+++ b/sources/us/ny/statewide.json
@@ -5,7 +5,7 @@
         "state": "ny"
     },
     "website": "http://gis.ny.gov/gisdata/inventories/details.cfm?DSID=921",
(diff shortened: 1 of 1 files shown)


trescube:
oh, that's pretty

iandees:
:drooling_face:

migurski:
Never realized how empty that northeastern bit really is.

trescube:
Does this supersede #2482?

iandees:
Yep, thanks @trescube.

PR #2821, opened 2017-04-07
updated data url

trescube:


trescube:
Re-run this please

trescube:
Seems to fail randomly on different chunks, the data may have to be cached.  The layer name implies that it's 3 years old and not updated so caching may be the best approach.

iandees:
Woah.

trescube:
There are < 5 million points in this source but there are > 10 million people in the state so there's probably still value in sourcing as many counties as possible.

migurski:
👏🏼

nvkelso:
Hello outer banks! Nice job!

PR #2830, opened 2017-04-08
added Lumpkin County, GA, USA

trescube:


trescube:
Records are missing geometry for some reason.

PR #2896, opened 2017-04-23
updated data url and conform

trescube:
previous layer was points but records have no parseable geometry


trescube:
Neither address nor parcel layers have geometry.

PR #2898, opened 2017-04-23
updated srs, source was erroring on previous value

trescube:


trescube:
@migurski this one is similar to #2899 where there's an error in the logs, the processing was successful, but the lat/lon in the output seems suspicious:  

```
1607946.30958,433234.291462,120,ROCK VALLEY RD,,PEACHLAND,,,,,7f930af6d15043a9
```

migurski:
If the original SRS was a projection and we’re assigning it the value EPSG:4326, i'm not surprised to see that the numbers look odd here. I would expect a North Carolina state plane projection such as [EPSG:3359](http://spatialreference.org/ref/epsg/3359/), whose [proj4 representation](http://spatialreference.org/ref/epsg/2264/proj4/) looks similar to the value in the diff.

trescube:
K, thx, I'll give that a try!

trescube:
That did it, thx @migurski !

PR #2899, opened 2017-04-23
added number field

trescube:


trescube:
@migurski this is one that was successful but has an error in the logs.  There's data in the output but the lon/lat seem suspicious:

```
1524481.77876,511963.186367,1119,CHURCH HILL RD,,HICKORY TWP,,,,,2cc5518385073885
```

Is this a case where `srs` should be specified?

migurski:
Probably! [EPSG:2271](http://spatialreference.org/ref/epsg/2271/) for (1524481, 511963) yields [a point near Hickory PA](http://www.openstreetmap.org/#map=18/41.560389/-79.372545) on Church Hill Road so I bet it’s the one.

trescube:
That did it, thx again @migurski !

PR #2903, opened 2017-04-23
Taiwan - Tainan

sergiyprotsiv:
Tainan county of Taiwan. An ESRI source from the government website.

I will shortly be uploading a few more Taiwan sources that cover most of the populous parts of the country.

Appreciate comments on the Chinese street naming and numbering system. The combined "address" field lists "street" then "lane" then "house number". I have combined "street" and "lane" to form the OA "street field. Perhaps the space should be skipped.

### DIFF
--- a/sources/tw/tnn/regionwide.json
+++ b/sources/tw/tnn/regionwide.json
@@ -11,7 +11,11 @@
     "conform": {
         "type": "geojson",
         "number": "NO",
-        "street": ["ROADNAME","LANE"],
+        "street": {
+            "function": "join",
+            "fields": ["ROADNAME","LANE"],
+            "separator": "-"
+        },
         "city": "VILLAGE",
         "district": "DISTRICT",
         "region": "COUNTY"
--- a/sources/tw/tnn/regionwide.json
+++ b/sources/tw/tnn/regionwide.json
@@ -10,11 +10,20 @@
     "type": "ESRI",
     "conform": {
         "type": "geojson",
-        "number": "NO",
+        "number":{
+            "function": "regexp",
+            "field": "NO",
+            "pattern": "^([0-9\uff10-\uff19]+號?)"
+        },
+        "unit":{
+            "function": "regexp",
+            "field": "NO",
+            "pattern": "^(?:[0-9\uff10-\uff19]+號?)(.*)"
+        },
         "street": {
             "function": "join",
             "fields": ["ROADNAME","LANE"],
-            "separator": "-"
+            "separator": ""
         },
         "city": "VILLAGE",
         "district": "DISTRICT",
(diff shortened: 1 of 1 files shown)


sources/tw/tnn/regionwide.json
```diff
@@ -0,0 +1,23 @@
...
+    },
+    "data": "http://water.tainan.gov.tw/tnwrbarcgis/rest/services/TNRW/Address/MapServer/0",
+    "type": "ESRI",
+    "conform": {
+        "type": "geojson",
+        "number": "NO",
+        "street": {
+            "function": "join",
+            "fields": ["ROADNAME","LANE"],
+            "separator": "-"
```
thatdatabaseguy: separator should just be the empty string: ""

sources/tw/tnn/regionwide.json
```diff
@@ -0,0 +1,23 @@
...
+            "alpha2": "TW-TNN",
+            "country": "Taiwan"
+        },
+        "country": "tw"
+    },
+    "data": "http://water.tainan.gov.tw/tnwrbarcgis/rest/services/TNRW/Address/MapServer/0",
+    "type": "ESRI",
+    "conform": {
+        "type": "geojson",
+        "number": "NO",
```
thatdatabaseguy: The number field in these data sets contains the house number, floor and apartment. The following regex should capture only the house number: "^([0-9\uff10-\uff19]+號?)", and "^(?:[0-9\uff10-\uff19]+號?)(.*)" will capture unit.

sergiyprotsiv:
@thatdatabaseguy fantastic, thanks a lot. Will change the rest of the sources accordingly.

sergiyprotsiv:
Re-run this, please.

thatdatabaseguy:
Maybe try changing to "號" to "'\u865f".

thatdatabaseguy:
LGTM, those regexes should work for the other Taiwan data sets as well.

sergiyprotsiv:
Thanks, @thatdatabaseguy ! I looked at the output and in rare cases here (and more prevalently in other Taiwanese sources) there is a structure like ３９４之５號 (i.e. 號 comes in the end, 之 is in the middle). I do not think this should this be parsed as "house = 394, unit  = 之５號", but not sure what does this string imply. Should it be "house = 394號, unit  = ５之" or  "house =５號 , unit  = 394之" or something else entirely?

thatdatabaseguy:
That's fine. "號" just means "number", so an example like "之５號" is just like saying "Unit No. 5" instead of "之５" which would be more like "Unit 5". That symbol always comes after a number in Chinese, so when it's seen after the house number, it's referring to the house number.

sergiyprotsiv:
Great, thanks!

PR #2906, opened 2017-04-23
Taiwan - Taoyuan

sergiyprotsiv:
Another large region in Taiwan.

### DIFF
--- a/sources/tw/tao/regionwide.json
+++ b/sources/tw/tao/regionwide.json
@@ -10,7 +10,16 @@
     "type": "ESRI",
     "conform": {
         "type": "geojson",
-        "number": "HOUSE_NO",
+        "number":{
+            "function": "regexp",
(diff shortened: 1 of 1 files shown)


sources/tw/tao/regionwide.json
```diff
@@ -0,0 +1,23 @@
...
+    "conform": {
+        "type": "geojson",
+        "number": "HOUSE_NO",
+        "street": {
+            "function": "join",
+            "fields": ["STREET","SECT","LANE","ALLEY"],
+            "separator": ""
+        },
+        "city": "LI",
+        "district": "TOWNNAME",
```
thatdatabaseguy: TOWNNAME is the city, LI is district (technically it's a city district, but OpenAddresses currently doesn't have a field for that, so it's usually stored in district)

PR #2909, opened 2017-04-24
Taiwan - Yilan

sergiyprotsiv:
This source contains one file per township (covering the whole province), since the countywide address layer has a weird crop.

What's the best way to do coverage objects? So far I have put the whole county into each file, but I think the correct way would be to put in the polygons for the township, right?


sergiyprotsiv:
Is there a simple way to combine "remove_prefix" and "regex" functions? Need to remove the road name from the beginning of the address before doing the parsing.

migurski:
Testing a new version of Machine:

Re-run this, please.

sergiyprotsiv:
@thatdatabaseguy When you have time, can you take a look at the regex for this source? Basically, I want to do the same splitting of house number and unit as in other Taiwanese sources, but before that to remove the street name from the address. If there a simple way to combine "remove_prefix" and "regex" functions, would be simple. From what I understand, we can't use functions of functions, right?

andrewharvey:
Closes #3503

PR #2911, opened 2017-04-24
Taiwan - Hsinchu City

sergiyprotsiv:
This one has a strange offset in my local cache, let's see if it was just a fluke and the machine can sort it out...

### DIFF
--- a/sources/hsz/regionwide.json
+++ b/sources/hsz/regionwide.json
@@ -10,7 +10,16 @@
     "type": "ESRI",
     "conform": {
         "type": "geojson",
-        "number": "ADDR_NO",
+        "number":{
+            "function": "regexp",
+            "field": "ADDR_NO",
+            "pattern": "^([0-9\uff10-\uff19]+\u865f?)"
+        },
+        "unit":{
+            "function": "regexp",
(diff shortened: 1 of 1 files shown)


nvkelso:
projections :\

On Sun, Apr 23, 2017 at 9:52 PM, OpenAddresses Bot <notifications@github.com

sergiyprotsiv:
Only fixed the number parsing, the projection will still be off. My simple attempts to fix it did not bear fruit, will need to modify somehow.
Do we have a way to provide a PROJ string just in case? Or if it is a non-standard EPSG projection this source will need to be processed offline?

andrewharvey:
Since this offset looks about the same as [GCJ-02 or BD-09](https://en.wikipedia.org/wiki/Restrictions_on_geographic_data_in_China#Coordinate_systems), I thought I'd test it out just in case for some strange reason they were using China's coordinate system, no idea why they'd do that but worth testing, anyway testing with https://github.com/googollee/eviltransform for all 6 transforms was negative, nothing made it shift into the right location.

iandees:
I'm going to close this as I don't think it's worth spending much more time trying to find a good projection for this data.

PR #2918, opened 2017-04-25
added countrywide Uruguay

### LINKED ISSUE #2917: Uruguay Nation-wide Addresses
uy_oa_yeeeeehawwww_sample
uy_oa_yeeeeehawwww

After talking it over with @iandees we will add the data - I see no mention of any terms (Source: http://correo.com.uy/ ) - so we need to maintain that in the metadata if possible.

Dropbox link coming soon ;)

trescube:
Fixes #2917

### DIFF
--- a/sources/uy/countrywide.json
+++ b/sources/uy/countrywide.json
@@ -13,6 +13,7 @@
         "type": "shapefile",
         "id": "id",
         "number": "numero",
-        "street": "nombre"
+        "street": "nombre",
+        "postcode": "cp"
     }
 }
--- a/sources/uy/countrywide.json
+++ b/sources/uy/countrywide.json
@@ -1,0 +1,18 @@
+{
+    "coverage": {
+      "country": "uy",
+      "ISO 3166": {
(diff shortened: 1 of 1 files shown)


trescube:
The data isn't fantastic but it's reasonable.

astoff:
The field "letra" is probably a suffix to the number, and should be concatenated.

nvkelso:
"Wrote 929117 points to /tmp/points-86eu3t99.bin”!

migurski:
Yup! When drawing the preview images that get posted here in comments, we stream all the points to a file in binary form to make it quick to saw through them:

https://github.com/openaddresses/machine/blob/d2d0dc4ae/openaddr/preview.py#L210-L220

muralito:
Where does the source file https://s3.amazonaws.com/data.openaddresses.io/cache/uploads/trescube/bf749a/uy.zip does come from?

nvkelso:
It’s listed in the original issue:
https://github.com/openaddresses/openaddresses/issues/2917

PR #2923, opened 2017-04-26
added Bucharest city and metro area

trescube:


trescube:
Postal codes in Bucharest are slightly off of what google maps has but not dramatically.  

Thx @justinelliotmeyers!

nvkelso:
Epic ring road

PR #2932, opened 2017-04-27
switching city and district in Taoyuan

thatdatabaseguy:
"LI" is a city district, "TOWNNAME" is the city


thatdatabaseguy:
Re-run this, please.

PR #2934, opened 2017-04-27
Portugal - Lisboa

sergiyprotsiv:
This is a minimal source for Lisbon, Portugal. Covers just the municipality itself. There are two more address datasets in the source, but the others are designed for lower zoom levels.

Coverage does not contain geometry so far, will add it to all Portuguese sources eventually.

### DIFF
--- a/sources/pt/11/lisboa.json
+++ b/sources/pt/11/lisboa.json
@@ -10,7 +10,7 @@
         "number": {
             "function": "regexp",
             "field": "NP",
-            "pattern": "(.*)(\\(.*\\))$"
+            "pattern": "(.*)(\\(.*\\))?$"
         },        
         "street": "MORADA"
     }
--- a/sources/pt/11/lisboa.json
+++ b/sources/pt/11/lisboa.json
@@ -11,7 +11,11 @@
             "function": "regexp",
             "field": "NP",
             "pattern": "([^\\(]+)"
-        },        
-        "street": "MORADA"
+        },
+        "street": {
+            "function": "regexp",
+            "field": "MORADA",
+            "pattern": "([^\\(]+)"
+        }
     }
 }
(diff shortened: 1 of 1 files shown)


sergiyprotsiv:
Most of the addresses seem okay, but there are quite a few that obviously contain wrongly formatted data or text, but without an obvious pattern. Shall I just filter out the addresses with regex (e.g. ^[0-9]+[A-Za-z]) and perhaps risk removing some of the valid ones, or keep things as they are?

astoff:
This is a bit of nitpicking, but could you replace `"city": "Lisbon"` with `"county": "Lisboa"` in the coverage? (Also, should it include `"state": "11"` or `"state": "Lisboa"` or not?).

Here's a quick explanation of Portugal's organization (from what I gather from Wikipedia). The country's administrative subdivisions are distrito > câmara municipal (aka município or concelho) > freguesia. They are analogous to state > county > civil township in the US. Cities and towns are an independent, and apparently somewhat informal, kind of entity. However, for added confusion, distritos and câmaras municipais are often named after their main city.

Now, it seems that most address data in Portugal will come from the câmaras municipais (as indicated by the "cm-" prefix in the domain names). It looks like the city and municipio of Lisboa coincide, but it will not be the case for all other regions, so let's be consistent from the beginning.

As to the regex, it seems to me that the only thing to fix is to remove all text between parenthesis at the end of the street names and numbers. Things like "Bloco 64B" are valid house numbers.

astoff:
The numbers look good to me. The same regexp should do the job for street names.

sergiyprotsiv:
Re-run this, please.

sergiyprotsiv:
Re-run this, please.

astoff:
Looks good to me!

migurski:
Testing a new version of Machine:

Re-run this, please.

PR #2936, opened 2017-04-27
add Matosinhos, Portugal

astoff:
Portuguese "freguesias" are like US civil townships, therefore that field in the original data is not relevant here.

### DIFF
--- a/sources/pt/13/matosinhos.json
+++ b/sources/pt/13/matosinhos.json
@@ -4,8 +4,10 @@
         "county": "Matosinhos",
         "geometry": { "type": "Point", "coordinates": [-8.7, 41.18] }
     },
-    "data": "http://web2.cm-matosinhos.pt/arcgis/rest/services/Limite_Planos/MapServer/3",
-    "type": "ESRI",
+    "website": "http://web2.cm-matosinhos.pt/portal/",
+    "note": "Cached from http://web2.cm-matosinhos.pt/arcgis/rest/services/Limite_Planos/MapServer/3",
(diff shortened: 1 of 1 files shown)


astoff:
Re-run this, please.

migurski:
This is a new error for us in Esri dumper:

```
2017-04-27 09:03:30,583  WARNING: Error in process_one.process()
  File "/usr/local/lib/python3.4/dist-packages/esridump/dumper.py", line 366, in __iter__
    data = self._handle_esri_errors(response, "Could not retrieve this chunk of objects")
  File "/usr/local/lib/python3.4/dist-packages/esridump/dumper.py", line 76, in _handle_esri_errors
    ', '.join(error['details']),
esridump.errors.EsriDownloadError: Could not retrieve this chunk of objects: Pagination request requires either orderBy field or the layer/table needs to have OID Field to facilitate the request. 
```

@iandees, have you seen this before? Is this a situation where we should fall back to geographic requests?

iandees:
Huh, I haven't seen that before. The "Pagination request requires either orderBy field or the layer/table needs to have OID Field to facilitate the request." message is an error coming from the Esri server. At the point that pyesridump is erroring I would expect it to have found an OID column already, so I'm not sure why the Esri server is complaining about it.

The simplest approach would be to detect this sort of error and fall back to geographic queries. A little bit more complicated would be to detect this and pick a column to order by.

astoff:
@sergiyprotsiv does the modified esridump you used for the countrywide Portugal source work with this server?

astoff:
I forced a geo query (by commenting a line on esridump) and managed to dump the layer, so we can cache the data for now.

PR #2940, opened 2017-04-30
Portugal - Porto

sergiyprotsiv:
Municipality of Porto. 
It is a cache of the (somewhat unstable) source here: http://mipweb.cm-porto.pt/arcgis/rest/services/Cache/CACHE_ENQUADRAMENTO/MapServer/0
Since the source did not have the name of the street, but it was in a related layer (http://mipweb.cm-porto.pt/arcgis/rest/services/Cache/CACHE_ENQUADRAMENTO/MapServer/1), this is a join of the two.

### DIFF
--- a/sources/pt/13/porto.json
+++ b/sources/pt/13/porto.json
@@ -13,6 +13,11 @@
         "lat": "Y",
         "lon": "X",
         "number": "N_POLICIA",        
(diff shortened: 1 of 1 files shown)


astoff:
Postal codes are formatted as `{CP_4}-{CP_3}`.

PR #2948, opened 2017-04-30
Portugal Countrywide

sergiyprotsiv:
A countrywide Portugal dataset, more than 3.5 million points, though it seems that it only covers residential buildings. It is hosted by INE (Portuguese statistical office).

It was created as part of the 2011 Census, and the database itself is called "Base geográfica de edifícios", but there is little specific info on the terms. Most census-related data in Europe are free to use (e.g. see the terms of the related Base Geográfica de Referenciação da Informação 2011, http://mapas.ine.pt/download/metadados/bgri11.html).

There are multiple similar sources on INE's webpage, but many of them return 0 count for the datapoints after a delay. I used a custom version of pyesridump with forced OID-based queries and it worked fine. Since the main dataset (for continental Portugal) needs a bit of pre-processing, I used that step to also merge in the data for the Azores and Madeira archipelagoes that were separate sources with separate projections. See the script file for more details.

### DIFF
--- a/scripts/pt/portugal_geoedif.py
+++ b/scripts/pt/portugal_geoedif.py
@@ -1,0 +1,66 @@
+import json
+import os.path
+
+# The source data can be obtained by running
+# esri2geojson https://inenetw05.ine.pt:6443/arcgis/rest/services/GeoEdif/GEOEDIF_AC25/MapServer/0 pt-azores25.json
+# esri2geojson https://inenetw05.ine.pt:6443/arcgis/rest/services/GeoEdif/GEOEDIF_AC26/MapServer/0 pt-azores26.json
+# esri2geojson https://inenetw05.ine.pt:6443/arcgis/rest/services/GeoEdif/GEOEDIF_MAD/MapServer/0 pt-madeira.json
+# esri2geojson https://inenetw05.ine.pt:6443/arcgis/rest/services/GeoEdif/GEOEDIF_CONT/MapServer/0 pt-continente.json
+# In practice the continental dataset is too large (>3M points), so the server fails before returning the count. Forcing esri2geojson to use OIDs does the trick
+
+
+tmp_dir = "./tmp"
+common_db_path = os.path.abspath("./portugal_geoedif.db")
+common_csv_path = os.path.abspath("./portugal_geoedif.csv")
+
+small_files = ['pt-azores25','pt-azores26','pt-madeira']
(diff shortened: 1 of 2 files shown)


sergiyprotsiv:
According to Census statistics, there are just 25K non-residential buildings in Portugal (http://censos.ine.pt/xportal/xmain?xpid=CENSOS&xpgid=ine_censos_indicador&contexto=ind&indOcorrCod=0005969&selTab=tab10). It could be that the census only covers residential buildings and the 25K figure refers to partially used residential.

thatdatabaseguy:
With cities and postal codes to boot, looks great @sergiyprotsiv!

nvkelso:
Wow! Great work :)

migurski:
Awesome!

scripts/pt/portugal_geoedif.py
```diff
@@ -0,0 +1,66 @@
+import json
+import os.path
+
+# The source data can be obtained by running
+# esri2geojson https://inenetw05.ine.pt:6443/arcgis/rest/services/GeoEdif/GEOEDIF_AC25/MapServer/0 pt-azores25.json
+# esri2geojson https://inenetw05.ine.pt:6443/arcgis/rest/services/GeoEdif/GEOEDIF_AC26/MapServer/0 pt-azores26.json
+# esri2geojson https://inenetw05.ine.pt:6443/arcgis/rest/services/GeoEdif/GEOEDIF_MAD/MapServer/0 pt-madeira.json
+# esri2geojson https://inenetw05.ine.pt:6443/arcgis/rest/services/GeoEdif/GEOEDIF_CONT/MapServer/0 pt-continente.json
+# In practice the continental dataset is too large (>3M points), so the server fails before returning the count. Forcing esri2geojson to use OIDs does the trick
```
migurski: I’d be curious to hear more about this. /cc @iandees

sergiyprotsiv: Do you want the raw jsons uploaded too for inspection?
I figured that since this is a large download (taxes their server substantially) and that this particular source is unlikely to get updated (a version of data for new census is likely to be published as a different layer), the processing is fine.

sergiyprotsiv: The server failing to return count is often an issue even with the web interface, not just with pyesridump. Solved it by just commenting out the test for pagination support in the dumper and thus forcing OIDs.

PR #2969, opened 2017-05-03
Create garrett.json

jbchurchill:
I attempted to set up a .json file modeled after the Allegany.json file. Ours is based on our County's "Address Points" point file and the fields are different.

### DIFF
--- a/sources/us/md/garrett.json
+++ b/sources/us/md/garrett.json
@@ -1,7 +1,7 @@
 {
     "coverage": {
         "US Census": {
-            "geoid": "24012",
+            "geoid": "24023",
             "name": "Garrett",
             "state": "Maryland"
         },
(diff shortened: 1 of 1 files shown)


trescube:
Sorry, the values for `district` and `region` have to match the case in the source (`COUNTY` and `STATE`).  Otherwise it looks great!

jbchurchill:
I think I have this done as requested. Let me know if there is anything else that wasn't correct.

jbchurchill:
Ok I changed the case of COUNTY and STATE. I guess I didn't realize they were pointing to the source.

trescube:
Oh, I see!  Yeah, the attribute values should be taken directly from the ESRI source verbatim.

trescube:
Looks great, thanks again for the contribution!  

If you're willing to help out on more sources, you can look for holes in the coverage at any level (city, county, state, or country) or check out [Lost sources](https://results.openaddresses.io/dashboard) for sources that used to work but don't anymore!

PR #2971, opened 2017-05-04
Addition meta

ingalls:
Often ESRI sources need some additional meta data to be able to run.

This defines an object to pass options into pyesridump

Ref: https://github.com/openaddresses/pyesridump/pull/38

cc/ @migurski @iandees


iandees:
Would it make sense to generalize this to "extra args" that just get passed in to the cacher? 🤔

migurski:
What do each of the parts do? The documentation is slightly thin, and I’m wondering about the boolean flag. Also, would it be possible to rename the `type` property? The word is used a few other ways in our sources and in https://github.com/openaddresses/openaddresses-ops/issues/16 I’ve proposed that we try to avoid it since it’s so overloaded in meaning.

ingalls:
Closing as stale

PR #2979, opened 2017-05-08
Russia - Yamalo-Nenetskiy avtonomnyy okrug

sergiyprotsiv:
Another Northerly sparsely populated place in Russia. Not big w.r.t. number of addresses, but the coverage region would look quite impressive.

### DIFF
--- a/sources/ru/yan/regionwide.json
+++ b/sources/ru/yan/regionwide.json
@@ -1,0 +1,20 @@
+{
+    "coverage": {
+        "ISO 3166": {
+            "alpha2": "RU-YAN",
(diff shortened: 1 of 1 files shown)


sergiyprotsiv:
This one has a lot of houses labelled "0", but I cannot simply filter them out via regex since I need to combine it with a field that stories the letter part of house number. Any suggestions?

trescube:
I'd leave them as zeroes.  Sometimes that's valid and the consumer can filter out that clearly aren't (as we do in Pelias).

sergiyprotsiv:
Sounds good, thanks.

PR #2980, opened 2017-05-08
Added Slovenia countrywide addresses

stefanb:
Most current data can be checked at:
http://raba.openstreetmap.si/openaddresses/si-addresses-2017-05-07.zip

fixes #2926

Issue #2926, opened 2017-04-26
Slovenia Addresses


justinelliotmeyers:
![slov](https://cloud.githubusercontent.com/assets/5324316/25448039/4113dd30-2a85-11e7-8500-cac1a666174a.png)

justinelliotmeyers:
can @nvkelso or @iandees look at the terms?
http://egp.gu.gov.si/egp/

right under blue box (where it says Splošni pogoji uporabe geodetskih podatkov )

I think it is fine, but want another set of eyes on it

iandees:
It looks like a CC-BY license, which is fine.

justinelliotmeyers:
downloading a few supplemental datasets to hopefully add in street names as these only contain number/ apartment. they should join based on a street id field. stay tuned

nvkelso:
Wow!

justinelliotmeyers:
if you want you guys can sign up and dl directly!

justinelliotmeyers:
there is a whole bunch of other goodies in there ;) admins, settlements, etc

migurski:
😮 

That’s a whole country.

justinelliotmeyers:
yeah - we aren't messing around here... this is a serious global dataset!!!!!!!!!

sbma44:
awesome! [this was rumored to be coming](https://github.com/openaddresses/openaddresses/issues/1107) but it's amazing to see it actually happened

justinelliotmeyers:
here is the formatting: http://www.e-prostor.gov.si/fileadmin/struktura/sifrant_rpe_enot.htm

justinelliotmeyers:
uploading raw gis data to dropbox

justinelliotmeyers:
https://www.dropbox.com/sh/8pred67i3zycstl/AABhJOfrXRyqHi0s8J0ZbJWVa?dl=0

stefanb:
The dataset contains plenty of other information, should some of that be included as well (besides minimal mandatory street name and numbers)?
Eg city names, post codes (with names), commune & region names...

Some smaller places don't have street system, so the city name + house number is normally used instead. Is it ok to repeat the city name in both fields (street and city) in such cases?

There are 500000+ addresses, is it ok to have them all in a single csv or should it be split into smaller chunks (eg regionally)? "Address counts per source" chart on  http://results.openaddresses.io/dashboard tops at 200000 addresses, but there seem to be larger sources mentioned in the Lost sources section.

[...]

Issue #1701, opened 2016-05-26
au-countrywide quarterly update needs reprocessing


andrewharvey:
As per https://github.com/openaddresses/openaddresses/blob/master/sources/au/countrywide.json the source must be regenerated now that there is a May release of http://data.gov.au/dataset/geocoded-national-address-file-g-naf

`http://s3.amazonaws.com/data.openaddresses.io/cache/au.zip` still shows `Last-Modified: Sat, 05 Mar 2016`

ingalls:
cc/ @sbma44 did you have a script you used for this?

andrewharvey:
@ingalls I believe [this is the script](https://github.com/openaddresses/openaddresses/blob/master/scripts/au/gnaf.sh). I've made [pull request](https://github.com/openaddresses/openaddresses/pull/1703) which makes some changes to this script for the May16 release and some other fixes.

I've run the updated script which produced the [updated au.zip file](https://www.alantgeo.com.au/share/au.zip).

simonpoole:
While on this topic, I note that 
* OA claims that the data does not require attribution (it is licenced under a modified CC-by 4.0 licence)
* OA claims that it is freely shareable data, which it however, due to the above mentioned modification, isn't
In summary I'm not quite sure what this dataset is doing in OA except to one up other players.

iandees:
@simonpoole that would be a great topic for a different ticket.

andrewharvey:
>OA claims that the data does not require attribution (it is licenced under a modified CC-by 4.0 licence that clearly requires attribution)

Within https://github.com/openaddresses/openaddresses/blob/master/sources/au/countrywide.json license.attribution is true, where does OA claim it doesn't require attribution?

>OA claims that it is freely shareable data, which it however, due to the above mentioned modification, isn't.

The license.url contains the CC BY 4.0 license modified with a restriction on use for sending physical mail. The license doesn't prevent people sharing and using the data for other purposes does it?

simonpoole:
From licence.txt in the OQ -zip archive:

au/countrywide
Website: http://data.gov.au/dataset/geocoded-national-address-file-g-naf
License: https://data.gov.au/dataset/geocoded-national-address-file-g-naf/resource/09f74802-08b1-4214-a6ea-3591b2753d30
Required attribution: None

The modification to CC 4.0 needs to be passed on downstream.

migurski:
[...]

PR #2982, opened 2017-05-08
Slovenia countrywide - updated data and improved scripts

stefanb:
...to be compliant with the CC-BY license of the source dataset

Improves #2926, continuation of  #2980

### DIFF
--- a/scripts/si/.gitignore
+++ b/scripts/si/.gitignore
@@ -4,3 +4,4 @@ si-addresses*
 timestamp.txt
 CREDENTIALS*
 cookies.txt
+LICENSE-data.txt
--- a/scripts/si/LICENSE-data-template.txt
+++ b/scripts/si/LICENSE-data-template.txt
@@ -1,0 +1,13 @@
+Register prostorskih enot 
+Vir podatkov: Geodetska uprava Republike Slovenije, http://www.gu.gov.si
+Datum podatkov: %YYYY-MM-DD%, najnovejše podatke je možno dobiti na http://egp.gu.gov.si
+Podatki so objavljeni pod pogoji Creative Commons Priznanje avtorstva 2.5 Slovenija
+Polno besedilo dovoljenja je na naslovu: https://creativecommons.org/licenses/by/2.5/si/legalcode
+
+  --- English translation below: ---
+
+Registry of spatial units
+Source of data: The Surveying and Mapping Authority of the Republic of Slovenia, http://www.gu.gov.si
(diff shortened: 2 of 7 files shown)


nvkelso:
Ideally OpenAddresses gets the source data from the source, or we host on our own S3 bucket. But it looks like https://github.com/openaddresses/openaddresses/pull/2982/commits/909901f5dba13b7ad6e1ae12e4e2aa7426405f18 switches to OpenStreetMap local chapter hosting. Please advise.

stefanb:
Slovenian data requires some processing to download (it is behind form login) and de-normalize the data.
There is a `Makefile` for GNU `make`  script in scripts/si that does it all, but i have no access to your amazon s3 bucket to publish the results there.

If you can have 
* `make` executed periodically (eg once a month), 
* someone register at the data source (they just require email and they send a password, registrant has to agree to use the data in accordance with the CC-BY 2.5 license) and
* publish to openaddress' s3 bucket.

Then i will gladly change the scripts accordingly.

nvkelso:
@stefanb thanks for the background! @iandees what do you want to do here?

iandees:
Thanks for doing all the work on the download script here, @stefanb. I'm ok with using externally-hosted data for this because the script to generate the data is so well documented. 👍

stefanb:
Tnx!
I'll add it to crontab. What is the current schedule, when should the data
be ready?

12. maj 2017 01:31 je oseba "Ian Dees" <notifications@github.com> napisala:

Merged #2982 <https://github.com/openaddresses/openaddresses/pull/2982>.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/openaddresses/openaddresses/pull/2982#event-1079138194>,
or mute the thread

migurski:
Hi, we don’t currently run upstream source scripts like this on any kind of schedule. The closest thing is Australia’s ~quarterly updates, which we [process with a docker script](https://github.com/openaddresses/openaddresses/tree/master/scripts/au) when reminded in https://github.com/openaddresses/openaddresses/issues/1701.

stefanb:
I meant global batch import schedule (twice per week?), so that I can
adjust mine (running make in scripts/si) accordingly.

13. maj 2017 01:29 je oseba "Michal Migurski" <notifications@github.com>
napisala:

migurski:
Ah okay. Our global batch does happen ~twice weekly, typically on Mondays and Fridays.

PR #3009, opened 2017-05-26
update au/tasmania source

### LINKED ISSUE #1457: Update Tasmania info

andrewharvey:
Closes #1457

AU/Tasmania addresses are provided in a single state wide file.

Although we already have au/countrywide, it's still useful to have both in OA I think.

### DIFF
--- a/sources/au/tasmania.json
+++ b/sources/au/tasmania.json
@@ -1,20 +1,35 @@
 {
     "type": "http",
-    "note": "data URL expires; must use web interface & email workflow to download fresh data. cached archive acquired 2014-11-17. License provisions in archive are outdated as per conversation with LIST staff available at https://soundcloud.com/sbma444/confirmation-of-tasmanias-list-dataset-being-cc-by",
-    "data": "http://s3.amazonaws.com/data.openaddresses.io/cache/au-tasmania.zip",
+    "data": "http://listdata.thelist.tas.gov.au/opendata/data/LIST_ADDRESS_POINTS_STATEWIDE.zip",
     "license": {
-        "text": "CCBY",
+        "text": "CC BY 3.0 AU",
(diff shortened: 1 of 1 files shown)


andrewharvey:
address ranges appear as `84.0-86` since they are stored as real's not integers in the source data. Is there any way to fix this?

trescube:
Not without modifications to the OA machine.  It appears that when a single value is available, the conversion from real to integer is supported but not when the `format` function is used.  I'll take a look.

trescube:
When `number` consists entirely of, say, `349.0`, [this line](https://github.com/openaddresses/machine/blob/master/openaddr/conform.py#L957) removes the `.0` suffix when performing final normalization.  Unfortunately, when the `format` function has a field value ending with `.0` but other field values follow, no normalization is performed.  This shouldn't be a big deal to fix.

trescube:
Submitted a PR to the machine.  Hopefully this will be live in the next day or two.

trescube:
Fixed in the machine, should be available within a day or two for source processing.

trescube:
DO NOT MERGE, processing was accidental, waiting on openaddresses/machine#649 to be deployed.

trescube:
Re-run this please

andrewharvey:
Thanks @trescube for your effort in changing machine to fix the issue here! Looks good to me, merge away.

Hopefully in the future there will be tools/reporting from OA to see a diff of regions doubled up (eg. au/tasmania and au/countrywide), and to get a combined dataset without the duplicates.

PR #3010, opened 2017-05-26
New data source for New Zealand

thatdatabaseguy:
There's a new street addresses layer for New Zealand here: https://data.linz.govt.nz/layer/3353-nz-street-address/. The layer that was being used previously (street addresses electoral) has been deprecated and is going to be replaced with this layer.  Uploaded a new cached source for it and updated the script/download instructions.

The new layer has a slightly cleaner separation of cities and suburbs. Suburbs are not really a separate component in OpenAddresses, so storing them in the district column. Spot checked it with OSM and it looks like in cases where suburb_locality is present but town_city is not, the suburb_locality field is a standalone remote village which could be considered the city. Otherwise it's a suburb/neighborhood. Could also invert the fields (suburb_locality=city, town_city=district) if anyone has a preference.


thatdatabaseguy:
Added a new cached Docker image to the instructions and decided to indeed switch locality_suburb to be the city, with the understanding that when there's a district field present and the name differs, it's a suburb.

migurski:
Thanks Al, I appreciate you taking the time to bring the scripts up to date as well.

andrewharvey:
oh man, I've just been working on this over at https://github.com/openaddresses/openaddresses/pull/2161#issuecomment-302059696

andrewharvey:
hey @thatdatabaseguy @migurski a lot of the work here was already at https://github.com/openaddresses/openaddresses/pull/2161 and now there are merge conflicts there.

I've been working on updating that PR based on the advice at https://github.com/openaddresses/openaddresses/pull/2161#issuecomment-302059696. Any chance we could revert this PR and focus on #2161?

thatdatabaseguy:
Didn't see that. Happy to add the region field if needed, but for my purposes (training data for [libpostal](https://github.com/openvenues/libpostal)), "town_city" is important. It contextualizes whether suburb_locality is a suburb of a major city or a standalone village, which are two different classes for us. I generally don't think a source field should be ignored in favor of a derived one when possible.

andrewharvey:
@thatdatabaseguy So back in Nov 2016 I updated it per https://github.com/openaddresses/openaddresses/pull/2161/commits/bd9ed11dbe95f5485fd98fe0ca468e067cdee234 and it was just as complete as your PR but no one merged it it sat there until now, so I figured that's because it wasn't good enough and needed the spatial join, so I've been slaving away trying to fix that so it could be merged, then it turns out that your PR which implements what I had in Nov 2016 gets merged straight away and now I need to go to additional effort to rebase so I'm not too happy about that. I'd prefer this PR get's reverted (to remove the merge conflicts) and we all focus efforts on #2161.

The folks from the data provider suggest to derive the other fields, https://github.com/openaddresses/openaddresses/pull/2161#issuecomment-302059696. I think we should follow their advice.

> but for my purposes (training data for libpostal), "town_city" is important. It contextualizes whether suburb_locality is a suburb of a major city or a standalone village, which are two different classes for us. I generally don't think a source field should be ignored in favor of a derived one when possible.

OA doesn't have the capacity within the current schema to differentiate a suburb within a major city and and standalone village https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#attribute-tags. Maybe you could propose a change in the schema model to support your use case here?

I can't comment on NZ but in Australia the city an address falls in doesn't form part of the address so shouldn't be in OA (which is about street addresses after all, not administrative regions). eg. `1 Main Street, Balmain, NSW, Australia`, although the city this is in is Sydney, it doesn't form part of the address so we use the `suburb` in the OA `city` field as it's the best fit and leave `Sydney` out.

Until we can clarify what it should be for NZ, we should revert this and instead focus on merging #2161.

/cc @migurski

thatdatabaseguy:
Sorry I missed your PR. The spatial join was definitely not a requirement for merging. It probably just didn't get merged because people were out for Thanksgiving in the US in late November, it required maintainer action, and the build was failing/it seemed like there were some issues still being worked out, so it got lost in the fray of red PRs. Mine was merged quickly simply because, as a frequent contributor to this repo, I have push access to master as well as the cache, and therefore I was able to upload the new file and merge it myself. In any popular project with many contributors worldwide, there's bound to be some duplication of effort, but it's usually more effective to focus on preventing current/future wasted effort (i.e. asking @migurski to add you to the OpenAddresses org so this doesn't happen again) than on sunk costs. If it's a matter of credit attribution, happy to use your changes instead. The merge conflicts can be resolved with "git checkout --ours" on the various files in your PR, which would just overwrite my changes in favor of yours where applicable.

In any case, I would still advocate for keeping the town_city field from the source in place of at least one of the spatial joins. In reading over #2161, two things jump out:

1. In that [same comment](https://github.com/openaddresses/openaddresses/pull/2161#issuecomment-302059696) from Bill Nelson at LINZ, he says "Stats NZ boundaries are administrative / political boundaries so not always the same as common usage in postal addresses." That reads less like advice on what should be done and more like: here's how to do it if the city/district/region schema is a requirement (which it's not).

2. In https://github.com/openaddresses/openaddresses/pull/2161#issuecomment-304435852 you mentioned that the spatial join, which is the only real difference between the PRs, doesn't complete in a reasonable amount of time (could still likely be implemented with SQLite but would need to build an R-tree index for that to be efficient - see [here](http://www.gdal.org/ogr_sql_sqlite.html) under "Spatialite spatial index")

So I'm left asking: if district/region are not required or regularly used in NZ addresses, are not part of the original source, and adding them requires a more complicated script and a longer runtime (even with a best-case R-tree that narrows every point query down to exactly one bounding box, that's still ~4 million point-in-polygon tests on relatively large polygons, which is not cheap), why not just include what's in the primary source?

For OpenAddresses the only required fields in practice are lat/lon, number, and street. City and postcode are nice to have, and everything else is optional. I usually try to ensure the presence of city when it's available and does not conflict with OSM's definitions, as it can save hours of compute time in point-in-polygon tests for libpostal's use case (we import virtually all 450M+ addresses from OA, so it adds up) without really disrupting anyone else's workflow.

Re: a suburb field in OA, it's possible but considerably more involved as it would require a backward-incompatible schema change to every source, as well as migrating everything in a way that wouldn't break downstream consumers. That seems a relatively large effort for a handful of sources when there are adequate workarounds using the existing fields.

andrewharvey:
> Sorry I missed your PR. The spatial join was definitely not a requirement for merging. It probably just didn't get merged because people were out for Thanksgiving in the US in late November, it required maintainer action, and the build was failing/it seemed like there were some issues still being worked out, so it got lost in the fray of red PRs. Mine was merged quickly simply because, as a frequent contributor to this repo, I have push access to master as well as the cache, and therefore I was able to upload the new file and merge it myself. In any popular project with many contributors worldwide, there's bound to be some duplication of effort, but it's usually more effective to focus on preventing current/future wasted effort (i.e. asking @migurski to add you to the OpenAddresses org so this doesn't happen again) than on sunk costs. If it's a matter of credit attribution, happy to use your changes instead. The merge conflicts can be resolved with "git checkout --ours" on the various files in your PR, which would just overwrite my changes in favor of yours where applicable.

It was more disappointing that it could have been merged long ago, but good that it's finally merged now. No worry about credit. Agreed it will be better to focus on future work now, then dwell on this. Thanks for the git tip, I didn't know about that option.

> In #2161 (comment) you mentioned that the spatial join, which is the only real difference between the PRs, doesn't complete in a reasonable amount of time (could still likely be implemented with SQLite but would need to build an R-tree index for that to be efficient - see here under "Spatialite spatial index")

Yeah I tried that, but it's still way too slow.

> So I'm left asking: if district/region are not required or regularly used in NZ addresses,

It would be great for some locals to help out here. https://www.nzpost.co.nz/personal/sending-within-nz/how-to-address-mail/correct-address-formats-envelope-layouts says "Don’t use a province, region, district or territory name instead of (or in addition to) the town or city name.". which seems to backup your approach.

Issue #3021, opened 2017-06-01
Dubai Address Data


mmb436:
I reached out to the GIS unit at Dubai Municipality for address data, and they said that their Makani address data is open and free to use, and sent me this link:
http://login.dm.gov.ae/wps/portal/opendata?urile=wcm:path:/DMContentEnglish/Home/OpenData/Makani&mapping=opendata

I need help figuring out how to download the actual address data. There are some instructions here:
http://login.dm.gov.ae/wps/wcm/connect/b1c96c7e-98d3-447a-b4b4-6c84f15e027a/Makani+Public+Web+Service+Access.pdf?MOD=AJPERES

Can someone please help figure this out?

thatdatabaseguy:
Great find! Have been looking forward to data from Dubai.

Poked around a bit and came across this ESRI layer: http://www.makani.ae/arcgis/rest/services/DMap_EL_2015/MapServer/14. It has ~127k records with Makani codes, which are essentially unique IDs for each building entrance (each Makani has a QR code and is the only piece of information needed to get to practically any location in Dubai, very interesting system). Of those, ~57k have a street name listed. That is, suffice to say: :fire::fire::fire:

For adding into OpenAddresses, we usually separate multilingual sources by language, so there'd be one file for Arabic and one for English (see [Saudi Arabia](https://github.com/openaddresses/openaddresses/tree/master/sources/sa/al-madinah) for an example). I'd recommend mapping Makani to house number, the road name in each language to street, and the community name in each language to district (city would be understood to be Dubai).

If you'd prefer to send the pull request, feel free. Otherwise I can do it.

mmb436:
Thanks. I actually don't know how to convert the files, so if you're OK to do the pull request, that would be great. Thanks again.

thatdatabaseguy:
No problem, just added them in #3026.

PR #3026, opened 2017-06-02
adding Dubai addresses 🇦🇪

thatdatabaseguy:
From #3021, split into Arabic and English files similar to Saudi Arabia.

### DIFF
--- a/sources/ae/du/dubai-ar.json
+++ b/sources/ae/du/dubai-ar.json
@@ -1,0 +1,15 @@
+{
+    "coverage": {
+        "country": "ae",
+        "city": "Dubai"
+    },
+    "data": "http://www.makani.ae/arcgis/rest/services/DMap_EL_2015/MapServer/14",
+    "type": "ESRI",
+    "language": "ar",
+    "conform": {
+        "number": "MAKANI",
+        "street": "ROADNAMEAR",
+        "district": "COMM_NAM_1",
+        "type": "geojson"
+    }
+}
--- a/sources/ae/du/dubai-en.json
+++ b/sources/ae/du/dubai-en.json
@@ -1,0 +1,15 @@
+{
+    "coverage": {
+        "country": "ae",
+        "city": "Dubai"
+    },
+    "data": "http://www.makani.ae/arcgis/rest/services/DMap_EL_2015/MapServer/14",
+    "type": "ESRI",
+    "language": "en",
+    "conform": {
+        "number": "MAKANI",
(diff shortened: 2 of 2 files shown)


thatdatabaseguy:
Re-run this please.

nvkelso:
Those islands!

On Thu, Jun 1, 2017 at 10:58 PM, OpenAddresses Bot <notifications@github.com

CChelmis:
@thatdatabaseguy, is it possible to map Long/Lat/Makani info to building names?

thatdatabaseguy:
@CChelmis there's not a field in OpenAddresses for building name, so wouldn't be able to list it here. If you have building polygons, it's possible to reverse geocode to the building to get the name. We do this for planet-wide OSM buildings with [libpostal's R-tree polygon indices](https://github.com/openvenues/libpostal/blob/master/scripts/geodata/polygons/reverse_geocode.py).

It seems like there's a building layer here with names: http://www.makani.ae/arcgis/rest/services/DMap_EL_2015/MapServer/7, but not an obvious and correct way to join it to entrances/Makani.

CChelmis:
@thatdatabaseguy tbh just getting started with OpenAddresses and geocoding. I was curious though, from the file(sources/ae/du/dubai-en.json) the Makani for the entrance corresponding to the Index Tower for example is 26484 89015. However, if you search for Index in http://www.makani.ae the same entrance appears to have Makani 26486 89012. So was not clear to me where this discrepancy is coming from. Moreover, searching for 26484 89015 in http://www.makani.ae yields no results.

nvkelso:
Building name would be useful to add to data model. Not very common, but useful when it is.

migurski:
This looks to be the same as “Building” in the NAD Schema: 

screen shot 2017-06-05 at 5 54 39 pm

thatdatabaseguy:
@CChelmis hm, makani.ae may use this layer instead: http://www.makani.ae/arcgis/rest/services/SmartApp/MapServer/3, which contains "26484 89015" but not "26486 89012". That was the first layer I looked into but it didn't have street names, so opted for the layer that did (both street and house number are [required](https://github.com/openaddresses/openaddresses/blob/79f38428cf38bf02b70c3b88800de3c2a15c6941/schema/source_schema.json#L136) for OpenAddresses). The response from makani.ae does appear to return street names, but it's unclear which layers are included as the ESRI requests are made server-side.

Possible that (some of?) the Makani numbers in the layer used here are out of date.

PR #3074, opened 2017-06-15
updated data url, added unit and city

trescube:


migurski:
Hm, I don't see any street names in the output for this one.

trescube:
Might have fat-fingered something... lemme check.

trescube:
Yep, fixed in #3076

PR #3139, opened 2017-06-22
Add New Mexico Statewide

ingalls:
Got a tipoff that the state address file from 2016 lives here: http://rgis-data.unm.edu/rgis/ApolloPro.aspx

### DIFF
--- a/sources/us/nm/statewide.json
+++ b/sources/us/nm/statewide.json
@@ -1,0 +1,19 @@
+{
+    "coverage": {
+        "country": "us",
+        "state": "nm"
+    },
+    "data": "http://data.openaddresses.io/cache/uploads/ingalls/us-nm.zip",
+    "type": "http",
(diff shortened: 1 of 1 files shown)


trescube:
That is a fine looking map (lived in New Mexico for 2 years).

iandees:
@ingalls how did you download and cache these?

iandees:
For posterity:
1. Go to the website http://rgis-data.unm.edu/rgis/ApolloPro.aspx
1. On the left, under "Search RGIS Data Catalog", search for `address points`
1. First result in the lower right should be "New Mexico Address Points"
1. Click it so the row highlights red, then click the Download link above it

nvkelso:
Last big hole in the western USA, great work!

migurski:
A fine-looking state in a fine-looking aspect ratio.

Issue #3169, opened 2017-06-28
Fingal County, Ireland Addresses


justinelliotmeyers:
http://gis.fingalcoco.ie/ArcGIS/rest/services/Permeability/Geodirectory/MapServer/0

Not 100% sure these are addresses - but believe they are

Did not see anything mentioning copyright in Disclaimer - or pop-up in map

trescube:
Whatever it is, there are 1.1 million records.

nvkelso:
Nice find!

trescube:
Only ended up being 597k records.  Still a ton of data, though!

![Uploading Screen Shot 2017-06-28 at 12.00.57 PM.png…]()

trescube:
<img width="582" alt="screen shot 2017-06-28 at 12 11 42 pm" src="https://user-images.githubusercontent.com/658991/27647932-fc47b724-5bfa-11e7-9534-846cfa67244e.png">

trescube:
And whatever this is to the west. 

<img width="1245" alt="screen shot 2017-06-28 at 12 12 28 pm" src="https://user-images.githubusercontent.com/658991/27647984-1ad7a244-5bfb-11e7-8b04-13e2f743d1da.png">

trescube:
Ah, the top image appears to be the region of Dublin (which contains Counties Dublin and Fingal).  Still not sure what the bottom image is.

trescube:
The bottom image is for Galway, though there don't appear to be any actual addresses for that city.  County Dublin looks great, though.

iandees:
Have we done any work to confirm license on this?

PR #3172, opened 2017-06-28
Revert "added County Dublin, IE"

trescube:
Reverts openaddresses/openaddresses#3170

Waiting on license confirmation.


iandees:
For #3169.

PR #3173, opened 2017-07-07
freshen au-victoria data

andrewharvey:
I'll need someone to push the zip to the OA S3 cache then I'll force push this commit to update the data url, before merging this PR.

### DIFF
--- a/sources/au/victoria.json
+++ b/sources/au/victoria.json
@@ -1,7 +1,7 @@
 {
     "type": "http",
-    "data": "http://s3.amazonaws.com/data.openaddresses.io/cache/au-victoria.zip",
-    "website": "http://services.land.vic.gov.au/landchannel/content/vicmapdata?productID=1",
+    "data": "https://www.alantgeo.com.au/share/au-victoria.zip",
+    "website": "https://www.data.vic.gov.au/data/dataset/address-vicmap-address",
     "compression": "zip",
     "coverage": {
         "ISO 3166": {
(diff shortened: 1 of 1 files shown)


sources/au/victoria.json
```diff
@@ -1,7 +1,7 @@
 {
     "type": "http",
-    "data": "http://s3.amazonaws.com/data.openaddresses.io/cache/au-victoria.zip",
-    "website": "http://services.land.vic.gov.au/landchannel/content/vicmapdata?productID=1",
+    "data": "https://www.alantgeo.com.au/share/au-victoria.zip",
```
iandees: This URL is 404'ing. Can you throw it in Dropbox or similar and let us know what the URL is so we can add it to our S3 bucket?

andrewharvey: Thanks for pointing that out, I forgot that file was still needed for this PR!

I'm re-uploading it now, will post back when it's done.

It's the same as downloading via the official website (except the official website requires you to go through a shopping cart process and the download URL isn't static).

iandees: Try uploading with this form:

https://results.openaddresses.io/upload-cache

It will generate a URL in our S3 that you can use, if you'd like.

andrewharvey: The original URL is working again now.

That upload-cache URL would be handy, I'm just not sure about authorising it to "write all public repository data" and to "write all user data"...

iandees: Yea, that's a curiosity with GitHub's OAuth system. We have to use these permissions to determine if you're a contributor to openaddresses' GitHub team.

If you're not comfortable, I will reupload it when I get back home.

andrewharvey: I see, until that's fix by GH then, if you don't mind uploading it when you get a chance that would be great. Thank you!

PR #3174, opened 2017-07-07
GeoNB

ingalls:
GeoNB provincewide source is currently missing a lot of data as the shapefile doesn't have all the aux tables.

This adds a docker file to process the raw data into OA format as well as updating the source to point at the new file.

### DIFF
--- a/scripts/ca/nb/geonb.sh
+++ b/scripts/ca/nb/geonb.sh
@@ -101,6 +101,15 @@ echo "
             END AS street,
             unit_id AS unit_id,
             floor AS floor,
+            counties.name AS district,
+            CASE
+                WHEN place_names.name_e = place_names.name_f THEN
+                    place_names.name_e
+                WHEN street_types.lang_cd = 1 THEN
+                    place_names.name_e||','||place_names.name_f
+                WHEN  street_types.lang_cd = 0 THEN
+                    place_names.name_f||','||place_names.name_e
+            END AS place,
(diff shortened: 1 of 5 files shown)


scripts/ca/nb/geonb.sh
```diff
@@ -0,0 +1,142 @@
...
+    CREATE TABLE final AS
+        SELECT
+            l.id AS id,
+            l.pi_par_pid AS  pid,
+            l.civic_num||COALESCE(l.civic_num_suff,'') AS number,
+            CASE
+                WHEN street_types.lang_cd = 1 THEN
+                    street_name||COALESCE(' '||street_types.name, '')||COALESCE(' '||street_directions.name_e, '')
+                ELSE
+                    COALESCE(street_types.name,'')||street_name||COALESCE(street_directions.name_f, '')
```
thatdatabaseguy: Noticed some of the French street names in the new file are missing a space e.g. "RueVeniot".

One idea that might make life easier in these sorts of situations is using the concat_ws function instead of COALESCE. Does the same thing (including ignoring NULL arguments) but with nicer syntax. Example:

```
concat_ws(' ', street_types.name, street_name, street_directions.name_f)
```

ingalls: Oh that's awesome! I wasn't familiar with that function. I'm going to be away from the computer for the next week or so, want to create a PR if you can and tag @migurski or @iandees to just double check it for merge?

thatdatabaseguy: cool, done with #3254

PR #3200, opened 2017-07-11
Documentation for the "chain" conform function

thatdatabaseguy:


trescube:
Nice!  I was trying to use `chain` for [Teton, ID](https://github.com/openaddresses/openaddresses/blob/master/sources/us/id/teton.json) and made a mental note to ask you to add some docs for chain.  Great timing!

PR #3202, opened 2017-07-11
adding chain function to schema validator

thatdatabaseguy:


migurski:
Nice! Let’s set up a source to use this successfully, and then [get it documented in the contribution guide](https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#attribute-functions).

thatdatabaseguy:
Hm, might need to make another PR to machine because the way it decides which fields to request is also dependent on conform structure.

PR #3234, opened 2017-07-14
updated conform

trescube:


thatdatabaseguy:
Weirdly these lat/lons are in Naperville, IL. Data url is right, but that's what the server returns.

PR #3253, opened 2017-07-16
removing house number prefix from street name in WV

thatdatabaseguy:
A fraction of the street names contain house numbers, but luckily ADDRNUM is populated in almost all cases.


thatdatabaseguy:
Re-run this please.

PR #3288, opened 2017-07-24
updated data url and conform (City of Cambridge, MA)

trescube:
I switched this from GeoJSON to CSV due to errors in the machine processing like this:

`ValueError: dict contains fields not in fieldnames: 'EditDate'`

Some, but not all, records from the source GeoJSON have a field named `EditDate`.  This was causing an issue because the machine converts the GeoJSON to CSV internally and the dictionary of keys is established from the first record (which doesn't contain `EditDate`) so when a record containing an unknown key comes along, the machine fails on the source.  

The CSV source is consistent in fields throughout the file.

### DIFF
--- a/sources/us/ma/city_of_cambridge.json
+++ b/sources/us/ma/city_of_cambridge.json
@@ -16,12 +16,24 @@
         "state": "ma",
         "city": "Cambridge"
     },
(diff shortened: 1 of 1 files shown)


trescube:
Aw, dangit, the source doesn't contain lat/lon.  Grr...

trescube:
Superseded by #4154

PR #3295, opened 2017-07-25
Adding supermanzana/manzana/lote patterns to house numbers in Mexico countrywide

thatdatabaseguy:
Updated the script, uploaded the resulting file to S3, and switched the source.


thatdatabaseguy:
"No space left on device" (sorry if that crashed anything). Peculiar though since it's definitely been run before several times before and the file size hasn't increased. Trying with hopefully a different machine.

Re-run this, please.

migurski:
If it fails again, could you file a Machine issue? I can increase volume size for our workers.

PR #3311, opened 2017-07-27
updated data url and conform (Dodge County, WI)

trescube:


trescube:
Re-run this please

sbma44:
Good times: looks like changing the URL scheme to `http` from `https` results in the server responding with HTML rather than JSON, which breaks this source. That's a shame, since the https endpoint's cert appears to be invalid, but the warnings generated don't appear to break `machine`.

Issue #3341, opened 2017-08-15
License violation for French addresses


jdesboeufs:
Hello,

I'm the actual product owner of adresse.data.gouv.fr (e.g. Base Adresse Nationale), whose datasets are referenced here.

Sadly our addresses are not currently available under CC0.

They are available under 2 licenses:
- ODbL 1.0
- _Licence gratuite de repartage_,  a country-specific license which is __more restrictive__ than ODbL 1.0, and clearly not compatible with other countries data.

In the future we will certainly provide alternative datasets with only Attribution constraint.

PR is coming.

nvkelso:
Thanks for the heads up @jdesboeufs! 

With your PR they will be flagged as such :)

migurski:
Fixed in https://github.com/openaddresses/openaddresses/pull/3342

migurski:
@jdesboeufs, different parts of the OpenAddresses service update at different frequencies. You can expect these changes to show up everywhere over the course of the next few days.

jdesboeufs:
No problem ;)

I will try to help you to maintain the `sources/fr` directory from today!
Many cities have attribution-only addresses published on their Open Data portal. Are you interested?

migurski:
Thank you, Jérôme! We are definitely interested in additional data. It’s helpful to have new sources even when they're redundant.

frodrigo:
This fix is only a partial fix.
The data pointed is not the ODbL version, is actually the BAN_licence_gratuite_repartage. The links must point to the ODbL version (provided by OSM-FR). (unless @jdesboeufs have now a better dataset to provide ;-) )

jdesboeufs:
The situation is evolving quickly.
The current specific license is not legal anymore, so ODbL is at the moment a good fallback.

PR #3342, opened 2017-08-15
Fix license for French addresses

jdesboeufs:
See #3341


iandees:
Thanks for updating these, @jdesboeufs!

migurski:
Yes, thank you!

PR #3354, opened 2017-08-23
add missing au/victoria metadata

missinglink:
[edit] can we please review/merge https://github.com/openaddresses/openaddresses/pull/3173 first, this PR would be a good candidate for a rebase against that PR

as reported in https://github.com/pelias/pelias/issues/628, the `au/victoria` conform is missing some fields.

see: https://gist.github.com/missinglink/3b9d0e2e7e65ca7e8e638ebc1a0c58f9 for example

note: I wasn't sure how to handle the `1A/1-35 KILFEERA ROAD BENALLA 3672` entry in that example, please advise.

PR #3360, opened 2017-08-25
[WIP] au/victoria split unit and number [need help]

andrewharvey:
I've sketched out splitting up the unit and number for au/victoria, which is something also proposed at https://github.com/openaddresses/openaddresses/pull/3354 but I don't think that it's right.

Currently we have `"number": "NUM_ADD"`. `NUM_ADD` is basically the UNIT and NUMBER pre combined.

This PR is actually a step back as it currently stands as I'm not sure how to deal with `hsaunitid`.

- [ ] The OA unit should be `hsaunitid` if defined, else use the format function in this PR, but I'm not sure how to express that. Any suggestions?

* I think for now we can ignore all the floor data. It's not right including it in `unit`, so best we omit it until OA has a structure for floors.
* same goes for building name

Good addresses for testing are:

> A7/6A-8C EVERGREEN MEWS
> 4-6/35D-35F BUCKLEY STREET

The documentation for the data is at  http://www.depi.vic.gov.au/__data/assets/word_doc/0016/340333/Vicmap-Address-Product-Description.docx

### DIFF
--- a/sources/au/vic/statewide.json
+++ b/sources/au/vic/statewide.json
@@ -30,7 +30,16 @@
                     "format": "gdb",
                     "layer": "ADDRESS",
                     "id": "PFI",
-                    "number": "NUM_ADDRESS",
+                    "unit": {
+                        "function": "format",
+                        "fields": ["BLG_UNIT_PREFIX_1", "BLG_UNIT_ID_1", "BLG_UNIT_SUFFIX_1", "BLG_UNIT_PREFIX_2", "BLG_UNIT_ID_2", "BLG_UNIT_SUFFIX_2"],
+                        "format": "$1$2$3-$4$5$6"
+                    },
+                    "number": {
+                        "function": "format",
+                        "fields": ["HOUSE_PREFIX_1", "HOUSE_NUMBER_1", "HOUSE_SUFFIX_1", "HOUSE_PREFIX_2", "HOUSE_NUMBER_2", "HOUSE_SUFFIX_2"],
+                        "format": "$1$2$3-$4$5$6"
+                    },
                     "street": [
                         "ROAD_NAME",
                         "ROAD_TYPE",


sources/au/victoria.json
```diff
@@ -23,10 +23,21 @@
     "conform": {
         "file": "ll_gda94/sde_shape/whole/VIC/VMADD/layer/address.shp",
         "type": "shapefile",
-        "number": "NUM_ADD",
+        "id": "PFI",
```
andrewharvey: > Uniquely identifies each address record. Persists through either  attribute or spatial representation changes. i.e. Remains for the life of the address object it identifies.

sources/au/victoria.json
```diff
@@ -23,10 +23,21 @@
...
+        "number": {
+            "function": "format",
+            "fields": ["hse_pref1", "hse_num1", "hse_suf1", "hse_pref2", "hse_num2", "hse_suf2"],
+            "format": "$1$2$3-$4$5$6"
+        },
         "street": [
             "ROAD_NAME",
-            "ROAD_TYPE"
+            "ROAD_TYPE",
+            "RD_SUF"
```
andrewharvey: This was overlooked before, but should be there.

andrewharvey:
PFI 426215601 which is this address `A7/6A-8C EVERGREEN MEWS ARMADALE 3143` shows the problem.

number is `6A8C`
unit is `A70`

So the `-` is being dropped, and `0` are appearing where they shouldn't.

The source data is:

 bunit_pre1 | bunit_id1 | bunit_suf1 | bunit_pre2 | bunit_id2 | bunit_suf2
------------|-----------|------------|------------|-----------|------------
 A          |         7 |            |            |         0 |            

The presence of the 0 there might be a problem...

sources/au/victoria.json
ranjan782: no I think not

andrewharvey: Could you please explain why not? There is a field in the source data ezi_address which is the complete formatted address which includes the RD_SUF, so if the ezi_address has it, then I think the OA address should.

iandees:
Closing as stale and conflicted with existing changes.

andrewharvey:
@iandees I'd rather leave this open for others to find easier. I got as far as I could but needed some help with

> The OA unit should be hsaunitid if defined, else use the format function in this PR, but I'm not sure how to express that. Any suggestions?

I'll rebase against master.

andrewharvey:
just rebased against master though the underlying issues remain.

iandees:
Looks ok. The build failure here is because you're on a fork.

Thanks!

andrewharvey:
@iandees I don't think this is ready yet, it looks like it can split unit out in simple cases, but it also gets many of the ranges wrong (whereas before the unit is mixed in with the number with a `/` and ranges are correct). I think it's safer to revert this one, until it's fully working and better than before.

iandees:
Ah, ok. Sorry for merging too early.

PR #3380, opened 2017-09-05
update address point URL to map service 

pmacMaps:
The map service currently used in this repo has a definition query set, so only some records are being accessed.  The updated map service contains all records from the address points dataset.

I've also added the post-directional field from the service (POD).


trescube:
Great catch, thanks!

Issue #3390, opened 2017-09-14
Sicily Italy


justinelliotmeyers:
https://dati.regione.sicilia.it/dataset/carta-tecnica-numerica-scala-1-2-000-numeri-civici

sergiyprotsiv:
Fixed in https://github.com/openaddresses/openaddresses/pull/3392

Issue #3339, opened 2017-08-10
Sicilia, Italy


sabas:
Source from the new open data portal
https://dati.regione.sicilia.it/dataset/carta-tecnica-numerica-scala-1-2-000-numeri-civici
The geojson is splitted by municipality

sergiyprotsiv:
Thanks a lot, @sabas! I think that the data in an official packaged form is preferable to getting it from the ESRI servers.
That said, the actual content appears to be identical to what is currently in OA database and the current method (ESRI scrapping) is fully automated, while using the GeoJSON would requite either custom preprocessing or introducing 350 separate sources for each municipality.
Do you know whether there are any plans for regular updates of this source? I think that would be the best reason to switch to the new format.

sabas:
Pinging @aborruso
Andrea, in your opinion the dataset from the open data portal will be kept updated?

aborruso:
Hi @sabas and @sergiyprotsiv there is no plan for regular update. 

The value of the static file is that the local public administration could switch off the ESRI rest server.
We could merge the 350 json and insert a municipality field.

sergiyprotsiv:
@aborruso thanks, that would make the move much simpler!
Can you confirm that the data in the GeoJSOn is identical to the ESRI server?

aborruso:
Hi @sergiyprotsiv, starting from 350 geojson file I have created a merged one and zipped it.

This is a temporary URL to download it http://dev.ondata.it/test/addresses.zip.

I have used gdal/ogr to create it.

```
for f in *.geojson; do ogr2ogr -update -append merge.sqlite $f -f "SQLite" -dsco SPATIALITE=YES -nln "addresses"; done;
ogr2ogr -f geojson addresses.geojson merge.sqlite addresses
```

The output contains the info about municipality in the fields `istat`, `comune` and `dataset`.

```json
{
  "type": "Feature",
  "properties": {
    "foglio": "5931607",
    "ident": 1,
    "civico": 0,
    "esponente": " ",
    "tipo_civico": "Residenziale",
    "progr_via": 10,
    "toponimo_strada": "CONTRADA ALCAMO MARINA",
    "orig_fid": 513645,
    "istat": "81001",
    "comune": "Alcamo",
    "provincia": "TP",
    "dataset": "81001_Alcamo"
  },
  "geometry": {
    "type": "Point",
    "coordinates": [
      12.9392456,
      38.0289006
    ]
  }
}
```

[...]

PR #3392, opened 2017-09-18
Update Sicily to use GeoJSON files

sergiyprotsiv:
Since the original ESRI server is unreliable, I follow the suggestions in https://github.com/openaddresses/openaddresses/issues/3339 and https://github.com/openaddresses/openaddresses/issues/3390 and use the new official GeoJSON source.
Since the original source contains 376 separate GeoJSON files, there is a preprocessing step involved to merge them. This should not be an issue, since the source is not updated regularly.


aborruso:
Thank you and thank you @sabas

PR #3398, opened 2017-09-18
added a bevy of Alabama counties

### LINKED ISSUE #3384: Alabama State Wide Parcels
not sure if this is every county, but it is a lot ;)

http://146.148.59.120/arcgis/rest/services/Parcels

trescube:
Fixes #3384


migurski:
This is way more than a bevy. 🎉

PR #3401, opened 2017-09-20
Adding contact and a little bit more detail in license object

iandees:
Adds stuff to the `CONTRIBUTING.md` document for two things:

* A `contact` section with stuff like phone, email, and mailing address.
* A couple extra pieces in `license` to give us somewhere capture more detailed license information.


iandees:
I also need to make changes to the schema JSON to match these, but I want to get feedback from others first.

migurski:
These are great changes. I’m curious why `remarks` would be a URL in addition to the other URL?

iandees:
Yea, that's confusing. I think the `license.url` points to the page on the provider's site that shows the license. It's pretty rare for the provider to have a page specifically for the license, so `license.remarks` would point to e.g. a document in another GitHub repo with a PDF dump of the license. This is based on the [whosonfirst-sources](https://github.com/whosonfirst/whosonfirst-sources) repo, which has a `remarks` field. [Here's an example for Ordnance Survey](https://github.com/whosonfirst/whosonfirst-sources/blob/master/sources/os.json#L12).

iandees:
@migurski can you think of a better name than "remarks" for that field? Otherwise I'm going to merge this in tonight.

migurski:
I can't, sorry. It still seems weird (the WoF example is odd: whose remarks are on that page?) but I’m having a hard time coming up with an alternative.

Issue #3404, opened 2017-09-23
Belarus


justinelliotmeyers:
Source:http://map.nca.by/

Can someone review terms before we green light uploading this ? @iandees or @nvkelso ?

I think we want to pay attention to section 3

Thanks!

![byuntitled](https://user-images.githubusercontent.com/5324316/30772921-082cd92a-a034-11e7-8d4f-4f3729a7cf54.jpg)

sergiyprotsiv:
My reading of the license is that most of it refers to the "Cadastral Map" as a piece of software (see section 1.2). I think that this is what the "noncommercial" clause and the "do not put on the open web" refer to (i.e. do not put a copy of their website elsewhere).
On the other hand section 4 states:
"4.1 The information contained in the Public Cadastral Map is freely accessible and is contained in the Appendix"
The Appendix then contains the list of datasources, which includes "Address Points" among other things.

justinelliotmeyers:
Thank you @sergiyprotsiv  !!!

justinelliotmeyers:
Uploading to Dropbox...

nvkelso:
I’m okay with that interpretation – I see the same language when using
Google Translate.

migurski:
Uploaded to https://s3.amazonaws.com/data.openaddresses.io/cache/uploads/migurski/e29094/belarus.zip

Here are some sample rows for building a conform:

```
OBJECTID,nameregion,namedistr,obj_name,elementtyp,elementnam,num_house,num_corp,ind_house,im_type,POINT_X,POINT_Y
1,Минская обл.,Вилейский р-н,д. Роличи,ул.,Зелёная,25.000000000000000,0.000000000000000, ,КС,26.640353226820025,55.063889450537772
2,Витебская обл.,Глубокский р-н,г. Глубокое,ул.,Молодёжная,26.000000000000000,0.000000000000000, ,ЗУ,27.702774862270132,55.148415073996546
3,Витебская обл.,Глубокский р-н,г. Глубокое,ул.,Московская,141.000000000000000,0.000000000000000, ,ЗУ,27.674873328341505,55.121986558532910
4,Витебская обл.,Глубокский р-н,г. Глубокое,ул.,Змитрока Бядули,13.000000000000000,0.000000000000000, ,ЗУ,27.692278842294797,55.154571624081257
[...]

PR #3409, opened 2017-09-26
Belarus countrywide

sergiyprotsiv:
A conform to import the source for Belarus discussed in https://github.com/openaddresses/openaddresses/issues/3404.

@thatdatabaseguy, could you please check the preferred format for addresses. Specifically, whether we should have "korpus" as part of the house number or as a unit.

Some fields in the input csv were obviously created as numeric, might need a regex on them to remove decimals and replace zeroes with nulls.

### DIFF
--- a/sources/by/countrywide.json
+++ b/sources/by/countrywide.json
@@ -1,10 +1,10 @@
 {
 	"coverage": {
         "country": "by",
-      "ISO 3166": {
-          "alpha2": "BY",
-          "country": "Belarus"
-      }
+	"ISO 3166": {
+		"alpha2": "BY",
+		"country": "Belarus"
+	}
     },
 	"data": "https://s3.amazonaws.com/data.openaddresses.io/cache/uploads/migurski/e29094/belarus.zip",
 	"website": "http://map.nca.by/",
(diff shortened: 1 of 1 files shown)


thatdatabaseguy:
It covers Minsk! Thanks @sergiyprotsiv

Format looks fine. Both fields being floats/decimals in the source means either some preprocessing or some postprocessing - I'm fine with the latter. Since there's no other unit field, it should be OK to stick the building field under unit.

nvkelso:
Looks like nearly 100% coverage!

sergiyprotsiv:
Thanks to @justinelliotmeyers ! The only issue is that we now have houses like 1.00000000000Б, i.e. the letter is appended to the string representing the floating point number. Is there any simple way to deal with it in the conform? Otherwise I'll just reprocess the input csv to remove the floats.

thatdatabaseguy:
I added a pull request https://github.com/openaddresses/machine/pull/670 for stripping multiple trailing zeroes, but for some reason my PRs to machine are not triggering a build.

@sergiyprotsiv might want to preprocess it for the moment if it's not too much trouble.

PR #3416, opened 2017-09-29
move au sources into region directories

andrewharvey:
This cleans up the au directory into regions in line with https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#naming-files


andrewharvey:
act/statewide is broken and has been since 2015 http://results.openaddresses.io/sources/au/city_of_canberra.

I'd like to delete it since the ACT Government doesn't seem to provide their address data as open data currently, but I'm not sure if deleting the source would also clear out the 2015 data which I guess is better than nothing.

migurski:
We could use the cached data we have from late 2015 as a data source, perhaps? That would keep those ~176k addresses around.

andrewharvey:
That would be ideal, that's what I'm not sure of, if changing the directory/filename will loose the history that we have at http://results.openaddresses.io/sources/au/city_of_canberra

migurski:
It’ll lose that specific history, but the address points will still exist someplace which may be a fair trade. We don't really have a smart way of handling file movements.

andrewharvey:
> It’ll lose that specific history, but the address points will still exist someplace which may be a fair trade. We don't really have a smart way of handling file movements.

I'm happy with that then. If the historical address points will still exist somewhere.

PR #3417, opened 2017-09-30
Better source for Wilkin

mmdolbow:
This source has more records and is likely to be better and more permanent, via their new open data website: http://share-wilkinco.opendata.arcgis.com/

### DIFF
--- a/sources/us/mn/wilkin.json
+++ b/sources/us/mn/wilkin.json
@@ -9,7 +9,8 @@
         "state": "mn",
         "county": "Wilkin"
     },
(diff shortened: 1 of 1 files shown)


mmdolbow:
Well, yeah, I'll do that if I can figure out the githubbery. ;)   That's what I get for just looking at the open data page, where they call it "Zip Code"....but of course it's just a single string with no spaces in the REST endpoint, as you discovered. Gonna give it a shot...

trescube:
Looks great, thanks for updating this!

PR #3419, opened 2017-10-02
us/fl/sarasota: switched source, updated fields

trescube:


trescube:
The provider seems to be performing some reorganization so I'll wait it out.

trescube:
Re-run this please

PR #3426, opened 2017-10-05
add city of melbourne, victoria, australia

### LINKED ISSUE #3669: Melbourne, Australia
https://data.melbourne.vic.gov.au/Property-Planning/Street-addresses/a7rp-xtya
m_st_addr

http://creativecommons.org/licenses/by/4.0/legalcode

andrewharvey:
Closes #3669

- [x] add city of melbourne
- [ ] decide what should be in `coverage`. ie. use ISO codes or full name? Use ISO field even if ISO fields only apply to country/state and not down to LGA level?


iandees:
image

Woah these are definitely address points made out of address ranges. Crazy!

PR #3468, opened 2017-10-12
Change number flag, Mason County WA

lizziegooding:
The `number` flag currently uses the `ADDRESS` field, which contains the full street address, rather than the `STREETNUMB` field, which contains only the address number.

cc @ingalls


trescube:
Could you add `STRNUMSUFF` as the `unit` field in the conform?

PR #3470, opened 2017-10-12
Slovenia: Delete old reprojected shapefile before reprojection using gk-shp, updated year in attribution, using https, updated README

stefanb:
There was a problem when new .shp files were downloaded daily, but were not processed as the old reprojected shapefile already existed and were used instead for the output.

http://results.openaddresses.io/sources/si/countrywide shows no addresses being added between July and October 2017 (staying at 554,306 addresses instead of gradually climbing to 555,224)


stefanb:
http://results.openaddresses.io/sources/si/countrywide shows that after October 14th 2017 the number of addresses in the prepared .csv file is increasing nicely using this fix.

stefanb:
Also updated the attribution year to 2018 (source definition `sources/si/countrywide.json` is being generated from the `scripts/si/countrywide-template.json` temple by the `Makefile`, but not committed to git to be respected during official machine run)

PR #3520, opened 2017-10-23
WIP: Washtenaw County, MI, USA: updated data url and conform

trescube:
Appears to only be Ann Arbor, needs investigation.


nkorth:
URL for this server has changed to "webmapsecure" and it requires a token now, see #4382

ingalls:
This seems blocked for the time being as we have no way of obtaining a token. Going to close as stale for now.

PR #3536, opened 2017-10-26
Update des_moines.json

iojjjjiiii:
URL changes every month


iojjjjiiii:
Guess they haven't fixed it yet... I'll re-try in November...

PR #3556, opened 2017-10-27
Alamance County, NC, USA: updated data url, added id

trescube:


trescube:
Reports a failure even though everything appears to have been successful.  I'm merging anyway since the dotmap and output data are complete.

PR #3648, opened 2017-11-29
update au/countrywide for NOV17 release

andrewharvey:
I couldn't find the corresponding NOV17 Admin Boundaries release, but I think that's okay.


migurski:
Thank you for updating this, and publishing a fresh copy of the whole dataset!

Issue #3635, opened 2017-11-17
WA statewide parcels


justinelliotmeyers:
http://geoservices.wa.gov/arcgis/rest/services/Parcels/ECY_Washington_State_Parcel_Database_2017/MapServer/0

justinelliotmeyers:
@ingalls dont know where this is at?
Source refreshed I believe:
http://geo.wa.gov/datasets/washington-state-parcels-2017
https://services.arcgis.com/jsIt88o09Q0r1j8h/arcgis/rest/services/Parcels2017gdb/FeatureServer/0

Just wanted to check, i dont see it in the runs:
http://results.openaddresses.io/?runs=all#runs

Thanks!

justinelliotmeyers:
Current Parcels:
https://services.arcgis.com/jsIt88o09Q0r1j8h/arcgis/rest/services/Current_Parcels/FeatureServer/0

Source:
http://geo.wa.gov/datasets/current-parcels

migurski:
🤘

PR #3652, opened 2017-11-30
Washington Statewide Source

ingalls:
Closes: https://github.com/openaddresses/openaddresses/issues/3635


ingalls:
@migurski did this timeout?

PR #3674, opened 2017-12-07
NC Alamance

ingalls:
Thank goodness file is no longer timestamped


ingalls:
http://results.openaddresses.io/jobs/ff83d62c-b5da-4f33-bfe7-a5707f05ca45

PR #3679, opened 2017-12-08
include link to upload-cache as an option for the data URL if upstream doesn't provide one

andrewharvey:
I couldn't find a link to https://results.openaddresses.io/upload-cache in the CONTRIBUTING file so I've added it here, is that okay?


andrewharvey:
> Good idea! Can you reword this a little bit? “Members of the OpenAddresses Github Organization” may help clarify who’s allowed to do this.

Thanks for the feedback, I agree this wording is clearer. Change commited.

PR #3741, opened 2017-12-13
Create statewide.json

fanenji:
Addresses from Regione Liguria


ingalls:
Hey @fanenji Thanks for your contribution! 

Our testing suite is failing as you used `accuracy: 0`. We don't have the ability to edit sources once they are uploaded to us. Judging from the data `accuracy: 1` or just omitting it would probably be better.

fanenji:
Thanks @ingalls

As suggested I set the Accuracy to 1.

ingalls:
🎉

Issue #3696, opened 2017-12-12
Okmulgee, OK Parcels


justinelliotmeyers:
https://maps.meshekgis.com/arcgis/rest/services/City_of_Okmulgee/Okmulgee_GIS_WebViewer/MapServer/1

PR #3768, opened 2017-12-21
City of Okmulgee

ingalls:
Closes: https://github.com/openaddresses/openaddresses/issues/3696


ingalls:
These are mailing addresses - not situs

PR #3789, opened 2018-01-09
Use OPA Data from Philly's data portal

mjumbewu:
Use the most up-to-date Office of Property Assessment data from Philly's data
portal. Use a regular expression for the street number, as it's a little easier
than trying to construct the full street number from component parts, especially
in the case of ranged addresses where the high range value is not included as a
component in the data.


iandees:
Great, thanks for adding this!

PR #3861, opened 2018-02-08
Add US census info to sources

trescube:


ingalls:
@trescube did you add the Census Data by hand? Or via a script? 

If by script we should add City data to https://github.com/openaddresses/openaddresses/blob/master/apply-us-data.py

PR #3867, opened 2018-02-15
new source for Polk County, FL

albarrentine:


albarrentine:
Hrm, this source definitely has > 380k rows. I think the server's returning the wrong min/max, though it supports the statistics feature, so esridump tries to make queries using those IDs and returns no results.

albarrentine:
Just committed some changes in https://github.com/openaddresses/pyesridump/pull/45 to handle this case (and possibly some other broken PRs), but would need to update the OA servers with the new version.

Can build locally for now. @migurski how often does pyesridump get upgraded on the build servers?

PR #3908, opened 2018-03-01
OpenAddresses V2 [Experimental]

ingalls:
## Context

Ref: https://github.com/openaddresses/openaddresses-ops/issues/21

Per above
- [x] Define new schema to use that supports multiple layers and multiple sources per layer
- [x] Provide scripts for converting current sources into valid v2 sources
- [x] Update all sources to V2

**DO NOT MERGE** A lot of work has to be done before this is even close to being ready


andrewharvey:
This is awesome, I've sketched out some parcel and building sources for au/nz in [v2-au-nz-parcels-buildings](https://github.com/openaddresses/openaddresses/compare/master...v2-au-nz-parcels-buildings). The proposed schema works really well.

andrewharvey:
@ingalls Does this mean that the OA v2 to expand scope to things like buildings and parcels is abandoned?

iandees:
This particular approach might be, but we're still very interested in expanding scope to other data sources.

ingalls:
@andrewharvey This same approach is still something I'm working on, just going to go with the incremental approach instead of switching all the sources at once.

https://github.com/openaddresses/openaddresses/pull/3923

PR #3909, opened 2018-03-02
update au/countrywide for feb2018

andrewharvey:
- [ ] The count went down, I need to check if this is an issue on our side or not.
- [x] @andrewharvey-openaddr was removed from the openaddresses github organisation on 25th Jan so I can no longer push files to the OA S3.

Anyone trying to reproduce you may run into https://github.com/minus34/gnaf-loader/issues/29

### DIFF
--- a/scripts/au/Dockerfile
+++ b/scripts/au/Dockerfile
@@ -6,6 +6,6 @@ RUN apt-get update -y && \
                     python python-psycopg2
 
 RUN git clone https://github.com/minus34/gnaf-loader.git /usr/local/gnaf-loader && \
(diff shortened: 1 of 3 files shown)


ingalls:
Oh shoot that was my fault @andrewharvey I was cleaning out and simplifying permissions and assumed that was an old account. I've sent an invite to re-add that account to the org.

ingalls:
@andrewharvey Just give me a ping when your comfortable with the changes, ie count and I'll get this merged! thanks a ton for your help!

andrewharvey:
@ingalls I think the count going down is from upstream so I've merged this. Cheers.

PR #3913, opened 2018-03-06
[wip] update au/qld/statewide based on DCDB source

andrewharvey:
- [ ] move cache to oa infrastructure. blocked by https://github.com/openaddresses/openaddresses-ops/issues/22
- [ ] drop empty records (have a coordinate but not address fields
- [x] drop LGA name from locality

### DIFF
--- a/sources/au/qld/README.statewide.md
+++ b/sources/au/qld/README.statewide.md
@@ -1,0 +1,11 @@
+# au/qld/statewide
+
+## Parcels/Addresses
+[Cadastral data weekly - whole of State Queensland](http://qldspatial.information.qld.gov.au/catalogue/custom/detail.page?fid={4091CAF1-50E6-4BC3-B3D4-229AA318231A}) data is updated weekly on Sunday.
+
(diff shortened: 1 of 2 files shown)


andrewharvey:
There is a separate address dataset at http://qldspatial.information.qld.gov.au/catalogue/custom/detail.page?fid={F878C43D-3087-4102-8F28-1CFEA49B34F1}. I'll see if it's better to use this instead of the full DCDB here...

andrewharvey:
> There is a separate address dataset at http://qldspatial.information.qld.gov.au/catalogue/custom/detail.page?fid={F878C43D-3087-4102-8F28-1CFEA49B34F1}. I'll see if it's better to use this instead of the full DCDB here...

- Formatting the address range is more difficult
- lacks postcode

andrewharvey:
Merging this as is. Ideally the cached file would be hosted on OA infrastructure but it's much easier when I can just scp, rsync, s3cmd new files to the cache, so I'll stick with the setup in this PR.

PR #3923, opened 2018-03-11
OpenAddresses V2 Incremental

ingalls:
- Convert US/DC/Statewide to use the OA V2 Schema
- Add V2 Schema
- Add tests for V2 Schema
- Source tests will validate against schema v1 or v2 depending on the format


ingalls:
This is about to be merged here: https://github.com/openaddresses/openaddresses/pull/5127

PR #3926, opened 2018-03-14
add au-qld-brisbane

andrewharvey:


andrewharvey:
Some rows contain multiple coordinates separated by a pipe.

Due to https://github.com/openaddresses/machine/issues/534#issuecomment-373216329 we can't use a regexp to pull out the first one, so these addresses are output with no coordinates.

I'm merging this now as most addresses are good, we can fix this up post https://github.com/openaddresses/machine/issues/534

jharpster:
Have you compared these addresses to the G-NAF and found much difference between the two sets?

andrewharvey:
> Have you compared these addresses to the G-NAF and found much difference between the two sets?

Extracting the region from au/countrywide using `ogr2ogr -f CSV -spat 152.6831 -27.8117 153.4961 -27.2131 -oo X_POSSIBLE_NAMES=LON -oo Y_POSSIBLE_NAMES=LAT -oo KEEP_GEOM_COLUMNS=YES au-countrywide-brisbane.csv au/countrywide.csv`

then comparing to au/qld/brisbane in QGIS (click to open a larger image)

osm-bcc-gnaf

It's quite revealing actually and shows that in this area the OSM is very good. 335 Boundary Street shows how the GNAF (in this area at least) and BCC data are cadastre based (likely lot centroids) with OSM being building based.

au/qld/brisbane is missing a few addresses but I believe this is due to the way they store them in their source data as I mentioned at https://github.com/openaddresses/openaddresses/pull/3926#issuecomment-373217364 so it's hard to do a fair comparison without fixing that first.

PR #3931, opened 2018-03-20
Fix build errors and improve script execution time

mcliment:
There were some build errors in the script and missing fields.

As well, the speed was optimized by removing calls to `subprocess`.


nigelhorne:
Wouldn't it also remove a subprocess to instead of

cat xml/* | grep -a zip | sed -e "s/^.*http/http/" -e "s/\" type=.*$$//" | grep -a -v "</id" | grep -a -v "<inspire" >> spain_catastre/gml_urls.latin1.txt

have

grep -a zip xml/* | sed -e "s/^.*http/http/" -e "s/\" type=.*$$//" | grep -a -v "</id" | grep -a -v "<inspire" >> spain_catastre/gml_urls.latin1.txt

In otherwords, I don't see what the initial calling of 'cat' adds.

PR #3940, opened 2018-03-26
[wip] add oa v2 au/nz parels/buildings

andrewharvey:


nvkelso:
Thanks for tackling this, a huge chunk of work!

andrewharvey:
https://catalogue.data.wa.gov.au/dataset/building-outlines/resource/fcfadd8a-f069-312a-86ff-a9a2d7c94339

migurski:

migurski:
Great! I think Nick had been working on it, but I don't believe it’s being actively advanced at the moment. Thanks for keeping this branch around.

andrewharvey:
I've added all the parcels from here in new PRs. This PR will still be useful later if buildings are supported, but for now I'll close it.

PR #3968, opened 2018-04-14
Fixing Kreis Viersen - URL to updated KML data

stefanb:
Fixing http://results.openaddresses.io/sources/de/nw/kreis_viersen


migurski:
Thanks!

PR #3875, opened 2018-02-16
WIP: Broward County, FL, USA: updated data url and conform

trescube:
Don't merge, source was broken but Arcgis source contains far fewer points than the shapefile.

PR #3969, opened 2018-04-14
Fixed US FL Broward county data URL

stefanb:
Fixing http://results.openaddresses.io/sources/us/fl/broward


stefanb:
Possibly a duplicate of #3875

stefanb:
Bump - this is now updated and checked with circle ci as well.

PR #3970, opened 2018-04-14
Fixed Italy (IT), South Tyrol

stefanb:
Fixing:
* http://results.openaddresses.io/sources/it/32/south-tyrol-de
* http://results.openaddresses.io/sources/it/32/south-tyrol-it
* http://results.openaddresses.io/sources/it/32/south-tyrol-lld - no such  "LLD" language in CSV


stefanb:
Interesting, `south-tyrol-lld` should be failing, because there are no `*_LD` fields in the [downloaded  CSV](http://geoservices.buergernetz.bz.it/geoserver/p_bz-gnames/ows?SERVICE=WFS&VERSION=1.0.0&REQUEST=GetFeature&TYPENAME=p_bz-gnames:Addresses&SRSNAME=EPSG:25832&outputFormat=CSV).

stefanb:
The new source does not contain the LLD language anymore. Should we set that source to skip, or leave it to fail?

stefanb:
Bump - this is now updated and checked with circle ci as well.

PR #3977, opened 2018-04-17
WIP: au/nsw/statewide from NSW_Property web service

### LINKED ISSUE #1152: investigate NSW APIs
CC-BY licensed, let's see if there's anything here

http://spatialservices.finance.nsw.gov.au/mapping_and_imagery/lpi_web_services

cc @ingalls

andrewharvey:
closes #1152

I'm prototyping this and it's proving difficult to run it through machine locally so I'm opening this PR to get some test runs happening.

- [ ] ~~do we need to combine 4 ESRI services, if so how should we do that~~ yes, geojson-merge.
- [ ] ~~should we let machine hammer the upstream server, or should we cache it first~~ cache it
- [ ] ~~will a complete esridump even work~~ no, since it seems they restart their server around 5am, but manually restarting at the offset it stopped and seems to work.
- [x] ~~should we use a regexp to extract the locality/street via a lookup of localities or via street type as the separator~~ for now using locality regexp
- [ ] ~~~should we add a new function to extract street names~~~
- [ ] split unit/number


andrewharvey:
I've got about ~1.3M records so far out of the ~3.8M, expect it to take a few days. It looks like they restart their server around 5am each morning as it times out at the same time each day. Given this, the TOS, and the the fact the data is split over 4 endpoints, I'm planning to cache this and not update it. They have a semi-bulk extract service, but it's unclear that the CC BY licensing applies to data from that service.

I'm not sure it's going to retrieve all features (not sure what happens when new features are added mid-download with respect to the pagination, but hey it's better than nothing I guess until there is a simpler way to retrieve this.

andrewharvey:
starting with Rural properties, later on will try to include Urban properties too

PR #3978, opened 2018-04-18
Add note to: us/md/statewide

MiniCodeMonkey:
Added note to Statewide MD dataset about incorrect coordinates. A customer helped track down this oddity.

For now, we're just going to filter out those addresses, but I wanted to add a note here as well.

I assume that there's no good way to do this on the OA-side right now? (Without needing to do custom scripts)

### DIFF
--- a/sources/us/md/statewide.json
+++ b/sources/us/md/statewide.json
@@ -7,7 +7,7 @@
         "country": "us",
         "state": "md"
     },
-    "note": "Does not cover the entire state. Referred to link from https://www.arcgis.com/home/item.html?id=20878dd2c20949a8a14a5a282ab98807",
+    "note": "Does not cover the entire state. Referred to link from https://www.arcgis.com/home/item.html?id=20878dd2c20949a8a14a5a282ab98807. Also note that about 25,000 addresses is incorrectly positioned at -81.5291885,37.5772618 (you might want to filter those out)",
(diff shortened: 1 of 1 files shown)


sources/us/md/statewide.json
```diff
@@ -7,7 +7,7 @@
         "country": "us",
         "state": "md"
     },
-    "note": "Does not cover the entire state. Referred to link from https://www.arcgis.com/home/item.html?id=20878dd2c20949a8a14a5a282ab98807",
+    "note": "Does not cover the entire state. Referred to link from https://www.arcgis.com/home/item.html?id=20878dd2c20949a8a14a5a282ab98807. Also note that about 25,000 addresses is incorrectly positioned at -81.5291885,37.5772618 (you might want to filter those out)",
```
nvkelso: That's the "Null Island" of their state plane system – that's likely to happen in other datasets, too. Would be amusing to set an (optional) null island coord for each dataset and filter those automatically, or tag them as "visiting Null Island" in export.

ingalls:
Thanks for the updated note @MiniCodeMonkey 

One thing I would suggest doing is filtering each source by its reported coverage and counting the number dropped as a percentage of a total. This is one of the initial steps we do on loading data internally. It led to fixes of many of the `coverage` tags as well as dropping a lot of `null island` style points.

Adding a QA/Cleaning pipeline is something I would love to do in the next few months but will focusing on getting building polys/parcels added first per: https://github.com/openaddresses/machine/pull/699

PR #3987, opened 2018-05-07
Update to latest New York State data

iandees:
Craig sent me a new data dump for New York State.


trescube:
Re-run this please

migurski:

arch0345:
Source has since been updated in #6246

Issue #170, opened 2014-05-02
Portugal


ingalls:
http://www.portugal.gov.pt/en.aspx
gabinete.seaot@mamaot.gov.pt

ingalls:
Stale - no response. Someone else is welcome to take this on.

sbma44:
I corresponded with Felipe Cruz back in January. He's blogged about open data efforts in Portugal, and pointed me toward their open data portal:

![image](https://cloud.githubusercontent.com/assets/31717/9228457/9971993e-40e6-11e5-8ff1-9d9b5529d763.png)

Portugal [is building a cadaster](http://www.dgterritorio.pt/static/repository/2014-04/2014-04-03105202_b511271f-54fe-4d21-9657-24580e9b7023$$5D83BE99-238C-4727-83D4-712E7C3188A0$$7140EDFA-BFEA-4E41-81B8-F7A924284B33$$file$$pt$$1.pdf). But they historically have not had one (lost the link, apologies).

Probably worth checking in with the INSPIRE process to see if they've made progress otherwise.

joao:
@ingalls @sbma44 

Saw this ticket in 2014, did some research back then and no resource was available. I thought I had commented but noticed today I didn't, sorry.
Going to share the research I did, as there was no progress made in the meantime in Portugal and this might serve as future reference for anyone looking on how to get the data.

The only public dataset that comes even closer to what is necessary to add Portugal to OpenAddresses, is one of made available by the National Post office, which has all the streets in Portugal, with their postal-codes and house numbers.
It is provided by the national post office, but it is lacking geographic coordinates.
You can see it here in a repository:
https://github.com/centraldedados/codigos_postais/tree/master/data

Also on this page you can see more details on that dataset (better Google Translate it):
http://centraldedados.pt/codigos_postais/


The open data portal of the Government, built in 2011 and still in 'beta', is a ghost town:
http://www.dados.gov.pt/
It has always been more focused in publishing spreadsheets with aggregated conclusions without raw data (ex: percentage of unemployment in each district).
Unfortunately all of these 'open data portals' in the country are outsourced projects to big consulting companies, that work on them for a couple of months, with no improvement or upgrade plan afterwards. Then the projects are just kept in maintenance mode, as it has happened with one referenced above, IGEO (http://www.igeo.pt)

I will try in the coming months to get hold of a contact inside one of these government departments to see if it possible to make anything they have publicly available.

sbma44:
[...]

PR #3991, opened 2018-05-15
Portugal countrywide

sergiyprotsiv:
Thanks to @joao in https://github.com/openaddresses/openaddresses/issues/170 we have a new properly licensed dataset for Portugal. I did not include any license information myself, perhaps it can be added later.

One issue is that the source for Madeira gives strange links (the domain is different from the rest of the regions), and these files cannot be accessed via the links provided. I have tried a few options to figure out where the files are, but they were unsuccessful.

Perhaps someone can give it a try, these are three files linked to here: http://inspire.ine.pt/AD/atom/CDG_AD_BNM_raa_AC25_Atom_en.xml. We need the ones that end with "EPSG4326.zip".

On the other hand, there seem to be all addresses in Madeira, so maybe the missing files are superfluous anyway.


nvkelso:
Is this ready for merge?

iandees:
I'll merge it – no sense in waiting for Travis if the hook builds it OK.

PR #4041, opened 2018-05-30
Updated Finland with URLs to 2018-05-15 data

stefanb:


migurski:
Looks great, thank you @stefanb!

PR #4130, opened 2018-08-10
added Happy Valley-Goose Bay, NL, CAN

trescube:


iandees:
For posterity, this was fetched from http://maps.happyvalley-goosebay.com/layers/CivicAddresses2.js, which Justin found at http://maps.happyvalley-goosebay.com/.

trescube:
Thx!  I'll add a note to the source.

PR #4160, opened 2018-08-15
Switch to dynamic Riverside CA source

ingalls:
Our current riverside, CA source is missing a lot of new subdivisions like the one above


migurski:

arch0345:
https://github.com/openaddresses/openaddresses/commit/53c2c92ca2ce1ffc85aa4bb0b2e890760b0b52c1 updated the source for Riverside County. Should this PR be closed?

PR #4228, opened 2018-10-05
La Reunion Encoding

ingalls:
Our La Reunion source has an encoding issue as can be seen from the screenshots below.

screenshot from 2018-10-05 13-43-58
screenshot from 2018-10-05 13-42-17

This PR will be removed once the encoding issue is identified and fixed.

### DIFF
--- a/sources/fr/la_reunion.json
+++ b/sources/fr/la_reunion.json
@@ -17,8 +17,7 @@
         "lon": "lon",
         "lat": "lat",
         "city": "nom_commune",
-        "postcode": "code_post",
-        "encoding": "ISO-8859-1"
+        "postcode": "code_post"
     },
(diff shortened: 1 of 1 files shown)


ingalls:
Fixed:

screenshot from 2018-10-05 14-49-10

jdesboeufs:
It is more probably an issue in source file.

ingalls:
@jdesboeufs the source file seems to be correctly coded in unicode, our ETL server was being told on our side to try to convert it from `ISO-8859-1` to Unicode. However since the source was already correctly coded, this was producing erroneous output.

Assuming correct unicode from the source seems to have fixed the problem. If you find any addresses where it has not be sure open a new ticket and give me a shout!

PR #4233, opened 2018-10-10
Repentigny

ingalls:
Creates a source from https://github.com/openaddresses/openaddresses/pull/4229

Closes: https://github.com/openaddresses/openaddresses/pull/4229

### DIFF
--- a/sources/ca/qc/repentigny.json
+++ b/sources/ca/qc/repentigny.json
@@ -11,10 +11,13 @@
             ]
         }
     },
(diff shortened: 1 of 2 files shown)


ingalls:
Doesn't look like this source is valid UTF8

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc9 in position 193: invalid continuation byte
2018-10-10 16:55:23,513  WARNING: Nothing processed
```

ingalls:
iconv guessed it as `iso-8859-1` ie `Latin 3`. 

Going to see what that produces, but running it locally made it look like the correct guess.

PR #4244, opened 2018-10-31
updated data url

trescube:


migurski:
This doesn't seem to be doing the thing.

trescube:
It appears to be throttled as it takes *forever* and ultimately times out.  I'll try caching in chunks with pyesridump.

trescube:
Re-run this please

trescube:
This may be fixed, I'm having luck locally with pyesridump.

PR #4249, opened 2018-11-07
Added countrywide Finland source (bilingual)

stefanb:


migurski:
Wow! How does this compare to our current FI sources? Should we replace those, or keep this additional one?

vesameskanen:
Hi! This looks like a good change.  We have asked VRK to distribute their source data via a fixed URL, but so far without success. 

I noticed that at the moment, there is no Finland data available at all, as the old sources are skipped, and the new countrywide data set is - for some reason - not included. Any idea when the transition is over and Finland is back on the map again?

stefanb:
Yes, it would be much easier if the data was available with a fixed URL. To ease the quarterly changes i made it dependant on 1 url in 2 source files instead of 19 URLs in 38 source files.

The data is already available:
http://results.openaddresses.io/sources/fi/countrywide-fi
http://results.openaddresses.io/sources/fi/countrywide-sv
Please check if everything is ok there. Content should be the same as before in all 19 regions.

The data will be available in the full openaddresses archive after the next batch run today or on Monday - see http://results.openaddresses.io/sets/

vesameskanen:
I can confirm that Finland data is available again and works as expected. Many thanks for your contribution!

Issue #2598, opened 2017-02-28
Lee County, GA data issue


justinelliotmeyers:
several times when I try to import tables into arcgis, they error out on the STREET field. typically I just change the length of the field to 500 and don't care. I'm getting ready to fill in suburb voids for ATL and brought in the state. I found this in the STREET field for Lee

"Atkinson_out.csv Berrien_out.csv Brantley_out.csv Camden_out.csv Charlton_out.csv clean.sh Colquitt_out.csv Cook_out.csv Glynn_out.csv Lanier_out.csv Lee_out.csv Mitchell_out.csv node_modules package.json README.md run.sh test.sh Turner_out.csv util "

I've seen stuff like this before in other tables, not sure how it's happening - just wanted to bring to groups attention.

Thanks,
Justin

nvkelso:
@justinelliotmeyers Can you clarify which OpenAddress data file you're seeing the `Atkinson_out.csv ...` content, please?

justinelliotmeyers:
Sorry!

https://s3.amazonaws.com/data.openaddresses.io/runs/162051/us/ga/lee.zip

migurski:
I peeked in the log, and it looks like our data source has these values. It was cached in late 2015, clearly we need a new one: https://s3.amazonaws.com/data.openaddresses.io/runs/162051/output.txt

justinelliotmeyers:
![lee_ga](https://user-images.githubusercontent.com/5324316/48422789-6216b500-e72d-11e8-973d-7cca92cb1e8f.png)


STAY TUNED!

ingalls:
Tracking fix: https://github.com/openaddresses/openaddresses/pull/4256

PR #4256, opened 2018-11-14
Update GA Lee source

ingalls:
Fixes: https://github.com/openaddresses/openaddresses/issues/2598


ingalls:
```
# us-ga-lee-1542280844219.geojson.gz
ok - s3://gis-data/openaddresses/us-ga-lee-1542280844219.geojson.gz
ok - 11096/11096 (100%) valid addresses
ok - 0 could not be parsed
ok - 0 suffered misc failures
```

Issue #646, opened 2014-10-20
us-pa-bedford


jalbertbowden:
http://www.bedfordcountypa.org/ 
Bedford County GIS

geobrando:
No GIS data. Contacted Direct of Planning Commission.

justinelliotmeyers:
https://services2.arcgis.com/tXFMtuwRfEDEFdnG/ArcGIS/rest/services/Bedford_County_Addresses/FeatureServer/0

justinelliotmeyers:
@ingalls newer source than what we have and pulls more addresses. Current source: http://results.openaddresses.io/sources/us/pa/bedford

PR #4263, opened 2018-11-19
US-PA-Bedford

ingalls:
Closes: https://github.com/openaddresses/openaddresses/issues/646

### DIFF
--- a/sources/us/pa/bedford.json
+++ b/sources/us/pa/bedford.json
@@ -9,17 +9,13 @@
         "state": "pa",
         "county": "Bedford"
     },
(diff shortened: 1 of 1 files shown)


ingalls:
```
# us-pa-bedford-1542632457249.geojson.gz
ok - s3://gis-data/openaddresses/us-pa-bedford-1542632457249.geojson.gz
ok - 28351/28352 (99.9%) valid addresses
ok - 1 outside of boundary (dropped)
ok - 0 could not be parsed
ok - 0 suffered misc failures

```

Issue #252, opened 2014-07-04
San Mateo County, CA


daguar:
I was unable to find a file online for San Mateo County, CA, so have contacted GIS and will follow up when I get a response.

geobrando:
Great work @daguar. It's currently in the sources

https://github.com/openaddresses/openaddresses/blob/839fba4480145573ea02b092b4cb49bafac1f964/sources/us/ca/san_mateo_county.json

PR #4270, opened 2018-11-27
Data download issue with San Mateo, California

jdstone-loansnap:
The original file download link was invalid, as was the path to the file within the zip file. Fixed both.

### DIFF
--- a/sources/us/ca/san_mateo.json
+++ b/sources/us/ca/san_mateo.json
@@ -15,7 +15,7 @@
     "type": "http",
     "compression": "zip",
     "conform": {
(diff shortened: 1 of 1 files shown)


migurski:
Actually, this might not be a bug Machine. Can you update your PR with this on line 18:

```
         "file": "SAN_MATEO_COUNTY_ACTIVE_PARCELS.shp", 
```

jdstone-loansnap:
Thanks, I did see the output log and the sample output.  I found this closed issue from earlier, [#252](https://github.com/openaddresses/openaddresses/issues/252), that looks like a similar issue was run into in the past.

migurski:

jdstone-loansnap:
@migurski, we can close this pull request if you'd like or think we should.

PR #4279, opened 2018-12-11
Update Waukesha County, Wisconsin City and Unit

andymarek:


ingalls:
```
# us-wi-waukesha-1544625287195.geojson.gz
ok - 206394/206875 (99.7%) valid addresses
ok - 339 have an unsupported number format
ok - 1 are missing street names
ok - 0 could not be parsed
ok - 141 suffered misc failures
```

Issue #2205, opened 2016-12-07
Mariposa county, California


nvkelso:
Part of "finish California" sprint.

nvkelso:
Another Parcel Quest county... via ([src](http://ice.ucdavis.edu/doc/parcel_data_contacts)):

```
Contact Information & Data Requests

Please direct questions to gis@mariposacounty.org or call the Mariposa County 
Planning Department at (209) 966-5151.

GIS Parcel Data Requests

In order to continue to provide the best possible GIS services to our own county 
agencies, Mariposa County has elected to license a GIS base map from ParcelQuest, 
a private company based in Folsom, California. As this map is a licensed, copyrighted 
product, Mariposa County is not at liberty to redistribute ParcelQuest's base map to 
public inquirers. 

However, if you would like to purchase a copy of the same map used by 
Mariposa County, you may contact ParcelQuest directly by visiting their website at 
ParcelQuest or calling them at 1-888-217-8999.
```

justinelliotmeyers:
https://services2.arcgis.com/wEula7SYiezXcdRv/ArcGIS/rest/services/ParcelMap/FeatureServer/0

double checking source

justinelliotmeyers:
![par](https://user-images.githubusercontent.com/5324316/49680512-27b9e280-fa63-11e8-8766-adbc582f40e4.jpg)

PR #4283, opened 2018-12-12
US CA Mariposa

ingalls:
Closes: https://github.com/openaddresses/openaddresses/issues/2205

### DIFF
--- a/sources/us/ca/mariposa.json
+++ b/sources/us/ca/mariposa.json
@@ -1,0 +1,20 @@
+{
+    "coverage": {
+        "country": "us",
+        "state": "ca",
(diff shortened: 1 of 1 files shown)


ingalls:
```
# us-ca-mariposa-1544639501493.geojson.gz
ok - 14295/14298 (99.9%) valid addresses
ok - 3 are missing street names
ok - 0 could not be parsed
ok - 0 suffered misc failures
```

migurski:
@ingalls are these stats checks you run something we should include in Machine?

ingalls:
@migurski would be pretty rad to include them, I have to get them into their own library. they're also currently in JS/Node

Issue #2207, opened 2016-12-07
Imperial county, California


nvkelso:
Part of "finish California" sprint.

nvkelso:
Via ([src](http://ice.ucdavis.edu/doc/parcel_data_contacts)):

Looks like good data, just need to get it.

- Map viewer: https://www.geoviewer8.com/?username=icpublic&password=icpublic
- County "GIS" page: http://www.co.imperial.ca.us/Assessor/

```
For further assistance and for questions or concerns about your property 
assessment, please call: (442) 265-4949, The Assessor’s property tax 
assessment information line or send an email at: assessor@co.imperial.ca.us .
```

Old contact?

```
Laura Cervantes, Cadastral Map/GIS Supervisor
County Assessor Office
940 W. Main Street, Suite 115
El Centro, CA 92243
760.482.4247
lauracervantes@co.imperial.ca.us
```

<img width="1391" alt="screen shot 2016-12-06 at 21 46 40" src="https://cloud.githubusercontent.com/assets/853051/20956366/3d9f2ef2-bbfe-11e6-9876-949623f18098.png">

<img width="555" alt="screen shot 2016-12-06 at 21 50 33" src="https://cloud.githubusercontent.com/assets/853051/20956365/3d83f86c-bbfe-11e6-98f2-a8655cba03ef.png">

nvkelso:
I've sent them an email.

nvkelso:
From Justin's comment in https://github.com/openaddresses/openaddresses/issues/2515#issue-206580096...

emailed Imperial County - assessor@co.imperial.ca.us

http://www.co.imperial.ca.us/Assessor/FIRelease.html

data here: http://services.geoviewer8.com/arcgis/rest/login?redirect=http%3A//services.geoviewer8.com/arcgis/rest/services/ImperialCounty_Public/MapServer/

nvkelso:
@justinelliotmeyers How do we get a login token for that Esri server?

justinelliotmeyers:
no idea! I just e-mailed them (assessor@co.imperial.ca.us ) and asked for data/ token.

justinelliotmeyers:
we can buy, and incorporate into the project - they accept check/ CC

Sure, my direct line is listed here below:
(442) 265-1311   
 
 
 
LAURA J CERVANTES, GISP
Cadastral Mapping / GIS Supervisor
Imperial County Assessor's Office
(442) 265-1311

nvkelso:
[...]

PR #4284, opened 2018-12-12
US CA Imperial

ingalls:
Closes: https://github.com/openaddresses/openaddresses/issues/2207


ingalls:
```
# us-ca-imperial-1544646081927.geojson.gz
ok - 64128/87733 (73%) valid addresses
ok - 4 outside of boundary (dropped)
ok - 23596 have an unsupported number format
ok - 0 could not be parsed
ok - 5 suffered misc failures
```

Issue #2204, opened 2016-12-07
Alpine county, California


nvkelso:
Part of "finish California" sprint.

nvkelso:
Another Parcel Quest county... via ([src](http://ice.ucdavis.edu/doc/parcel_data_contacts)):

```
Maps & Property Information
We currently do not publish Assessor maps, Recorded Maps or property information 
such as values or ownership on this website yet, but some information is available on 
the Parcel Quest website. See the links below for online info or visit our office and we 
will help you any way we can.

Office hours are Monday through Friday 8 a.m. to 12 p.m. and 1 p.m. to 5 p.m. 
excluding holidays.

At the Assessor / Recorders Office public counter you can:
View and purchase Assessor Parcel Maps (APN Maps), Official Recorded Maps and 
Survey Maps. Copies are available for a fee

For map & property information, come by or contact the Assessors office at:

99 Water St.
P.O. Box 155
Markleeville, CA 96120
Phone: (530) 694-2283
```

Data cost sheet (doesn't mention GIS data):

- http://alpinecountyca.gov/DocumentCenter/View/504

Also (probably old info):

```
Zach Wood
Alpine Planning Department
17300 St Rt. 89
Markleeville, Ca 96120
530.694.2140 x 437
zwood@alpinecountyca.gov
```

justinelliotmeyers:
$1300 - https://www.alpinecountyca.gov/DocumentCenter/View/52

Think it is too much - thoughts @ingalls ?

Thanks!

justinelliotmeyers:
e-mailed zwood@alpinecountyca.gov to make sure that is current (*believe is dated 2005) The license seems restrictive too... http://alpinecountyca.gov/documentcenter/view/53

I'll see what we can get!

justinelliotmeyers:
They will send the address points for free! Just got off the phone with Zach:
Zach Wood
Planner III
530 694 1371
zwood@alpinecountyca.gov

We just have to explain why sharing it freely would benefit the general public/ or the county.

nvkelso:
Awesome!

justinelliotmeyers:
from previous source mentiod tonight: https://gis2.cdfa.ca.gov/server/rest/services/Parcel_DM/MapServer/1

**Will wait to get data from county - assuming it would be more current and correct - will compare as soon as they ship

PR #4288, opened 2018-12-13
US-CA-Alpine

ingalls:
Closes: https://github.com/openaddresses/openaddresses/issues/2204


ingalls:
# us-ca-alpine-1544724966432.geojson.gz
ok - 1347/2373 (56.7%) valid addresses
ok - 1026 have an unsupported number format
ok - 0 could not be parsed
ok - 0 suffered misc failures

PR #4290, opened 2018-12-14
Add Winnebago County, Wisconsin

andymarek:


iandees:
Thank you for adding these! They look good to me, but I want to let @ingalls run his script on the output before merging them.

Issue #4300, opened 2018-12-19
fix Brant, Ontario


justinelliotmeyers:
new source: http://maps.brant.ca/arcgis/rest/services/PublicData/CountyOfBrantGIS/MapServer/71

broken issue:http://results.openaddresses.io/sources/ca/on/brant

PR #4308, opened 2018-12-19
Update Brant Source

ingalls:
Closes: https://github.com/openaddresses/openaddresses/issues/4300


ingalls:
```
# ca-on-brant-1545219915447.geojson.gz
ok - 14840/16283 (91.1%) valid addresses
ok - 1441 have an unsupported number format
ok - 0 could not be parsed
ok - 2 suffered misc failures
```

PR #4309, opened 2018-12-19
City of San Juan Capistrano

ingalls:
FOIA: https://www.muckrock.com/foi/california-52/e911-civic-address-data-62997/#


ingalls:
```
# us-ca-city_of_san_juan_capistrano-1545225499491.geojson.gz
ok - 12234/12236 (99.9%) valid addresses
ok - 0 could not be parsed
ok - 2 suffered misc failures
```

PR #4310, opened 2018-12-19
City of Newport Beach

ingalls:
FOIA: https://www.muckrock.com/foi/california-52/e911-civic-address-data-63001/


ingalls:
```
# us-ca-city_of_newport_beach-1545225680910.geojson.gz
ok - 54046/54046 (100%) valid addresses
ok - 0 could not be parsed
ok - 0 suffered misc failures

```

Issue #4311, opened 2018-12-19
London, Ontario


justinelliotmeyers:
https://apps.london.ca/generateform/default.aspx?form=opendatarequests

justinelliotmeyers:
all four of these might have to be used to make one layer?
https://maps.london.ca/arcgisa/rest/services/Addresses/MapServer

justinelliotmeyers:
@ingalls This is a pretty big gap in Ontario
Should I wait to hear back from their formal data request system, or do you think we should try to just pull these?

Thanks!

justinelliotmeyers:
| Gerry BockTechnologistGeomatics DivisionCity of London
-- | --
 
300 Dufferin Ave. London, On. N6A 4L9
P: 519.661.2489 x 2345 \| Fax: 519.661.6422
gbock@london.ca \| www.london.ca

justinelliotmeyers:
| Gerry BockTechnologistGeomatics DivisionCity of London
-- | --

justinelliotmeyers:
Hi Justin,

 

Unfortunately I cannot send the data you have requested without referring to a City project and project manager;

see link below…..

 

https://www.london.ca/business/Resources/Consultant-Resources/Pages/GISDataRequest.aspx

justinelliotmeyers:
I asked how to formally request it again although I filled out their formal request!?! Stay Tuned 'eh!

justinelliotmeyers:
they sent data attached here:
https://www.dropbox.com/s/xxe4oybgpxbza2d/attachments_20181219.zip?dl=0

justinelliotmeyers:
Justin,
--
 
Spoke to my boss – Gary Irwin, the City Surveyor.
 
I was informed that addresses are now open source data.
 
 
 
 
 
Gerry Bock
 
Technologist
 
Geomatics Division
 
City of London
 
 
 
300 Dufferin Ave. London, On. N6A 4L9
 
P: 519.661.2489 x 2345 \| Fax: 519.661.6422
 
gbock@london.ca \| www.london.ca

iandees:
Awesome! 🎉

PR #4314, opened 2018-12-20
City Of London

ingalls:
Closes: https://github.com/openaddresses/openaddresses/issues/4311


ingalls:
```
# ca-on-city_of_london-1545310296313.geojson.gz
ok - 131752/131752 (100%) valid addresses
ok - 0 could not be parsed
ok - 0 suffered misc failures
```

Issue #4350, opened 2018-12-31
Aurora, SD Address Points


justinelliotmeyers:
http://ims.districtiii.org/arcgis/rest/services/AURORA/MapServer/20

PR #4351, opened 2019-01-01
US SD Aurora

ingalls:
Closes: https://github.com/openaddresses/openaddresses/issues/4350


ingalls:
```
# us-sd-aurora-1546364098403.geojson.gz
ok - 1507/1507 (100%) valid addresses
ok - 0 could not be parsed
ok - 0 suffered misc failures
```

Issue #4349, opened 2018-12-31
Alex, SD Parcels


justinelliotmeyers:
http://ims.districtiii.org/arcgis/rest/services/ALEX_PARCEL/MapServer/1

ingalls:
Broken URL :/ (My bad for getting to it late)

justinelliotmeyers:
@ingalls whoa.... http://www.districtiii.org/gis/interactive_mapping.php

I'll go through these tonight/ tomorrow and sort out. JACKPOT!

ingalls:
@justinelliotmeyers Haha  you already have! A year ago

PR #4352, opened 2019-01-01
US SD Alex

ingalls:
Closes: https://github.com/openaddresses/openaddresses/issues/4349

### DIFF
--- a/sources/us/sd/alex.json
+++ b/sources/us/sd/alex.json
@@ -1,0 +1,14 @@
+{
+    "coverage": {
+        "country": "us",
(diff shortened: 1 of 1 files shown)


ingalls:
@migurski have you seen this before? Should we consider increasing the readtimeout value or is this usually just a symptom of a deeper problem with the MapServer.

migurski:
I feel like this might be a server problem? The read timeout looks to be 300 seconds, which is very long.

ingalls:
Closing as stale

Issue #4347, opened 2018-12-31
Brule, SD Parcels


justinelliotmeyers:
http://ims.districtiii.org/arcgis/rest/services/BRULE/MapServer/16

PR #4354, opened 2019-01-02
US SD Brule

ingalls:
Closes: https://github.com/openaddresses/openaddresses/issues/4347


ingalls:
```
# us-sd-brule-1546442188388.geojson.gz
ok - 10110/23779 (42.5%) valid addresses
ok - 13643 have an unsupported number format
ok - 0 could not be parsed
ok - 26 suffered misc failures
```

PR #4361, opened 2019-01-04
Add curitiba.json

astoff:


astoff:
I couldn't open the dotmap, but looking at the csv all seems fine (despite the 1500 points without address at the beginning of the file).

Issue #4345, opened 2018-12-31
Davison, SD Parcels


justinelliotmeyers:
http://ims.districtiii.org/arcgis/rest/services/DAVISON/MapServer/4

PR #4371, opened 2019-01-08
US SD Davidson

ingalls:
Closes: https://github.com/openaddresses/openaddresses/issues/4345


ingalls:
```
# us-sd-davidson-1546966757033.geojson.gz
ok - 10160/23779 (42.7%) valid addresses
ok - 13619 have an unsupported number format
ok - 0 could not be parsed
ok - 0 suffered misc failures
```

Issue #4344, opened 2018-12-31
Day, SD Parcels


justinelliotmeyers:
http://ims.districtiii.org/arcgis/rest/services/DAY/MapServer/14

PR #4372, opened 2019-01-08
US SD Day

ingalls:
Closes: https://github.com/openaddresses/openaddresses/issues/4344


ingalls:
```
# us-sd-day-1546966723052.geojson.gz
ok - 3506/11070 (31.6%) valid addresses
ok - 1 outside of boundary (dropped)
ok - 7563 have an unsupported number format
ok - 0 could not be parsed
ok - 0 suffered misc failures
```

Issue #4373, opened 2019-01-08
Fix Jefferson County, KY


justinelliotmeyers:
broken:
http://results.openaddresses.io/sources/us/ky/jefferson

new source:
http://data.lojic.org/datasets/jefferson-county-ky-address-points

https://services1.arcgis.com/79kfd2K6fskCAkyg/arcgis/rest/services/OpenDataAddresses/FeatureServer/0

PR #4379, opened 2019-01-08
US KY Jefferson

ingalls:
Closes: https://github.com/openaddresses/openaddresses/issues/4373


ingalls:
```
# us-ky-jefferson-1546981013742.geojson.gz
ok - 424245/424245 (100%) valid addresses
ok - 0 could not be parsed
ok - 0 suffered misc failures
```

PR #4381, opened 2019-01-09
Removed postcode from Chester

MiniCodeMonkey:
As the one in the data source is for the owner's address, and thus is incorrect quite often.


trescube:
Good catch, thx, @MiniCodeMonkey!  Do you have merge permissions?

MiniCodeMonkey:
@trescube Thanks for the quick response. I think that I used to, but I don't believe that I have merge permissions anymore. (It's been a little while since my last contribution).

PR #4382, opened 2019-01-10
New URL for us-mi-washtenaw

nkorth:
I found the new URL, but it still requires a token. At least it's better than a 404. (I did test it with the temporary token I got from the web UI, and it seems to work.)


migurski:

PR #4395, opened 2019-01-11
Fixed us-va-king_and_queen

MiniCodeMonkey:
Need to verify that the regex is indeed correct before merging


MiniCodeMonkey:
Success! Regex is good.

screen shot 2019-01-11 at 10 10 24 am

iandees:
Wahoo!

PR #4417, opened 2019-01-28
Added small note to us/al/cullman

MiniCodeMonkey:


iandees:
Thanks for pointing this out.

PR #4420, opened 2019-02-05
Update santa_cruz.json

### LINKED ISSUE #4401: us/az/santa_cruz has incorrect conform and data source seems to be down
The current conform fields for the `street` includes the city name in the output. I was looking into fixing it, but the source seems to be unavailable: https://gis.santacruzcountyaz.gov/arcgis/rest/services/ParcelSearch/Parcels/MapServer/0

What do we do in case like this? Can we temporarily disable the source until it is resolved? I would argue that we would rather have no output for this source, rather than output with errors/incorrect format.

https://github.com/openaddresses/openaddresses/blob/master/sources/us/az/santa_cruz.json

geobrando:
Replacing source which partially Fixes #4401.  Split may still be "crummy" because of city name.

### DIFF
--- a/sources/us/az/santa_cruz.json
+++ b/sources/us/az/santa_cruz.json
@@ -9,7 +9,7 @@
         "state": "az",
         "county": "Santa Cruz"
     },
-    "data": "https://gis.santacruzcountyaz.gov/arcgis/rest/services/ParcelSearch/Parcels/MapServer/0",
+    "data": "https://mapservices.santacruzcountyaz.gov/wagis01/rest/services/ParcelSearch/Parcels/MapServer/0",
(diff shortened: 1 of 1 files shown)


trescube:
Address points with separated fields appears to be here:  https://mapservices.santacruzcountyaz.gov/wagis01/rest/services/Buildings/AddressPoints/MapServer/0

geobrando:
@trescube Nice! I'll close this.

geobrando:
Per @trescube Point layer with good fields found here https://mapservices.santacruzcountyaz.gov/wagis01/rest/services/Buildings/AddressPoints/MapServer/0

PR #4422, opened 2019-02-07
fixed data url, added district and region

trescube:


trescube:
Don't merge yet til I verify further, this seems a bit light on data.

arch0345:
It looks like they have two address datasets: one for buildings and another for places where there's more than one address per building

iandees:
It looks like their server is using some really old TLS configuration that none of the Python request libraries support anymore. Does the server not work at all with http?

PR #4429, opened 2019-02-26
Add Belo Horizonte, Brazil

astoff:


astoff:
Looks good to me. I'll let somebody else merge, in case they want to take a look first.

iandees:
Thanks for adding this!

PR #4434, opened 2019-02-28
added statewide Idaho, USA

trescube:
Hopefully this is true


trescube:
This doesn't appear to be the entire state but I don't know how to classify it otherwise.

nvkelso:
Nice! Please set a multipolygon coverage like we did for Texas a while back here.

nvkelso:
Bummer still no Twin Falls...

Issue #4183, opened 2018-08-29
New Metro MN source


mmdolbow:
There's a new, complete, standardized source for metro MN counties that can be used to update sources/us/mn/metrogis.json
Details are at https://gisdata.mn.gov/dataset/us-mn-state-metrogis-loc-address-points. The shapefile isn't zipped, so I recommend the map service at https://gis2.metc.state.mn.us/arcgis/rest/services/loc_address_points/MapServer/0

This would cover Anoka, Carver, Dakota, Hennepin, Ramsey, Scott, and Washington counties. I'll try a PR if I have time...

justinelliotmeyers:
do the individual counties submit data to the met council, then they update it? We do have overlapping sources, which is good, but was just curious.  Cheers

mmdolbow:
That's exactly what happens!

PR #4439, opened 2019-02-28
added statewide Minnesota, USA

trescube:


iandees:
This is the "metro" area (Twin Cities and surrounding counties), not statewide. See #4183. I think we've held off on merging this in because we already have data from the counties that feed into this aggregation.

trescube:
Makes sense.  I found on a statewide server (from Joe Elfelt's PDF) and was hoping for the best.

PR #4459, opened 2019-03-16
us/mn/carlton missing street name in STREET column

colin-umn:
StreetName had extra underscore.  Also removed street suffixes and prefixes because
they are now included in the StreetName.


colin-umn:
I have inspected the output and the street names now appear.  Please merge.

PR #4474, opened 2019-04-01
Fix us/tx/city_of_austin

thismakessand:
Change source url to updated source.  Also fixed a couple of issues where street directions were not being picked up.  

I'm seeing an issue with extra spaces in the street concatenation though, where a row like "S,,CONGRESS" is becoming "S  CONGRESS" (with two spaces).  How can I fix it (assuming it doesn't get cleaned up later in the process)?

### DIFF
--- a/sources/us/tx/city_of_austin.json
+++ b/sources/us/tx/city_of_austin.json
@@ -33,7 +33,7 @@
                 {
                     "function": "regexp",
                     "field": "address",
-                    "pattern": "(\\d+)\\.?",
+                    "pattern": "(\\d+)\\.?(.*)",
                     "replace": "$1"
                 },
                 {
--- a/sources/us/tx/city_of_austin.json
+++ b/sources/us/tx/city_of_austin.json
@@ -27,32 +27,13 @@
     "compression": "zip",
     "conform": {
         "number": {
-            "function": "chain",
-            "variable": "number_wip",
-            "functions": [
-                {
-                    "function": "regexp",
-                    "field": "address",
-                    "pattern": "(\\d+)\\.?(.*)",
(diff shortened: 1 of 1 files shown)


trescube:
@thismakessand Thanks for sorting this out!  Are these the results you were expecting?  If so, do you have merge permissions for this PR? 

image

thismakessand:
@trescube Sort of - I'm still having an issue of extra spaces in the street name after the join conform function.    
eg, for this conform definition:
``` 
"street": [
   "prefix_dir",
   "prefix_typ",
   "street_nam",
   "street_typ",
   "suffix_dir"
]
```
When there's just a `prefix_dir` and `street_nam`(and an empty `prefix_typ`) it ends up with two spaces, eg `E  RIVERSIDE`.
I checked the shapefile dbf and it does not contain spaces so I'm not sure where the space is coming from.

trescube:
In this particular case, it would probably be easier to use the [`prefixed_number` and `postfixed_street`](https://github.com/openaddresses/openaddresses/blob/master/ATTRIBUTE_FUNCTIONS.md#prefixed_number-postfixed_street-and-postfixed_unit) functions on the `full_stree` field instead of `chain`/`join`/`regexp`.  `prefixed_number` and `postfixed_street` are useful for common US address formats.  Despite what the documentation says, half addresses *are* supported (see the supporting code [here](https://github.com/openaddresses/machine/blob/master/openaddr/conform.py#L86-L99)).

trescube:
@thismakessand how does that look?

thismakessand:
@trescube It looks good now - thanks!

trescube:
Great!  Thanks again for the contribution, I'll merge.

PR #4477, opened 2019-04-01
Added city to ca/on/simcoe

Valachio:


migurski:
The trailing comma after "MUNI" needs to come out, to make this valid JSON. Thanks for making this change!

PR #4486, opened 2019-04-03
Address number suffix and prefix for Olmsted

mmdolbow:
Also, cache seems old - should have about another 1,300 records


trescube:
Needs a `,` at the end of the `number` line.

PR #4487, opened 2019-04-03
Update to Olmsted

mmdolbow:
Also, cache seems old - should have about another 1,300 records


mmdolbow:
Not sure what happened here...I tested the REST query as detailed in the log and I can't get it to fail similarly.

trescube:
Try updating the `data` url from http to https.  I noticed that when i put the http URL into the browser, it redirects to https.

mmdolbow:
good call! Trying now.

mmdolbow:

PR #4495, opened 2019-04-15
Better source for Becker

mmdolbow:
This driveway point source appears to be better. The BASENAME field only has 11 records with nulls, and ADDRESS has 15, some of which otherwise appear to have real data. So this field mapping looks like an upgrade over using ADDRESS with a postfixed_street function.

### DIFF
--- a/sources/us/mn/becker.json
+++ b/sources/us/mn/becker.json
@@ -9,18 +9,14 @@
         "state": "mn",
         "county": "Becker"
     },
-    "data": "http://gis-server.co.becker.mn.us/arcgis/rest/services/BECKER_INTERNAL_101/MapServer/34",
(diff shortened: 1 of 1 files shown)


mmdolbow:
I think I may have an even better source, researching it for the null geometry I found in the REST endpoint. The better source also has the UNIT field.

trescube:
@mmdolbow Looks good, are you happy with this?

mmdolbow:
Yup, it's the best we can do. There's actually about 13 records with null geometries, but several of those don't even have any address info. I've notified them about all of the records, so hopefully they'll fix 'em and we'll see that in future updates. Definitely an upgrade over the former approach though, IMHO.

trescube:
Sounds good, thx for tracking this down!

PR #4508, opened 2019-05-09
adding second source for Miami-Dade, FL

albarrentine:
New government source for Florida (addresses, not parcels) from the FL Dept of Revenue. Has ZIP code and city for all 67 counties, seems to have better coverage in many counties than what's there currently. Plan is to add county.json for sources not currently in OA and county2.json for sources where a county source already exists.


albarrentine:
On second thought, now that this is working, will just concatenate all the files together and make it a statewide source. Closing.

PR #4509, opened 2019-05-09
adding FL statewide government address source from the Dept of Revenue

albarrentine:
Statewide government source from https://pointmatch.floridarevenue.com/General/AddressFiles.aspx, updated semi-annually it looks like, though requires a download. Appears to have better coverage in some areas (Gadsden, Miami), possibly less in others, unclear if that's because of duplicates in the existing sources and/or apartment addresses vs. building-level for geocoding. In any case, it's a supplemental source with a postal code and city for 9.1M addresses in all 67 counties.

### DIFF
--- a/sources/us/fl/statewide2.json
+++ b/sources/us/fl/statewide2.json
@@ -1,0 +1,33 @@
+{
+    "coverage": {
+        "US Census": {
+            "geoid": "12",
+            "state": "Florida"
+        },
+        "country": "us",
+        "state": "fl"
+    },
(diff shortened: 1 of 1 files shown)


nvkelso:
Why is the extent of this map so far zoomed out from the extent of Florida? Are there errant points in the dataset?

albarrentine:
yeah looks like there was at least one invalid point in there throwing it off, removed everything outside the bounds in #4511

PR #4576, opened 2019-06-20
Changed source URL slightly for us/ny/city_of_new_york

MiniCodeMonkey:
Using the "original" format instead of "shapefile" as shapefile appears to synchronously convert the data before it can be downloaded. I suspect that this leads to data corruption and/or timeouts.

The "original" format is already a shape file so the extra data conversion is completely unnecessary.

> Note: This change was prompted by the fact that this source started outputting significantly less addresses even though the source data does not have less records 

https://results.openaddresses.io/sources/us/ny/city_of_new_york
Screen Shot 2019-06-20 at 12 26 54 PM

### DIFF
--- a/sources/us/ny/city_of_new_york.json
+++ b/sources/us/ny/city_of_new_york.json
@@ -16,7 +16,7 @@
             ]
         }
     },
-    "data": "https://data.cityofnewyork.us/api/geospatial/g6pj-hd8k?method=export&format=Shapefile",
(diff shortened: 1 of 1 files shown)


migurski:
Weird, but it brings the number back up to 959k features!

MiniCodeMonkey:
Weird indeed. It'll probably also solve some of the timeout issues that the source has seen in the past, just because their server doesn't have to do that "just in time" transformation of the whole dataset.

Issue #2619, opened 2017-02-28
Floyd County, GA


justinelliotmeyers:
data is here: https://gis01.romega.us/arcgis/rest/services

but I cannot get past 
Error: Services Directory has been disabled

Source: http://gis.romega.us

Maybe @trescube or @iandees can work their magic on this!?

iandees:
There's parcels here: https://gis01.romega.us/arcgis/rest/services/Parcels/MapServer/0?f=json

I can't tell if the address-related fields there are situs or mailing, though.

arch0345:
Resolved in #4578

PR #4578, opened 2019-06-25
Add source for Floyd County, GA

thismakessand:
Skipping the city/state/zipcode attributes from the source since they are referencing the owner mailing address, not necessarily the property address.


thismakessand:
Closes #2619

iandees:
rerun this please

Issue #4560, opened 2019-06-06
Fix Flanders Belgium Source


justinelliotmeyers:
new source: 
https://downloadagiv.blob.core.windows.net/crab-adressenlijst/Shapefile/CRAB_Adressenlijst_Shapefile.zip

broken link:
http://results.openaddresses.io/sources/be/flanders

barthanssens:
See also #3002

barthanssens:
I guess this issue could be closed, since it's solved by #3002 ?

barthanssens:
Can be closed, as #3002 fixes this

PR #4582, opened 2019-07-09
Initial PR for Belgium

barthanssens:
Initial PR for Belgium from official federal source, with address datasets from the 3 regions (Walloon, Flemish, Brussels-Capital Region). 

We will probably need to add some more (meta)data files for translating names in "facility municipalities" (see also https://en.wikipedia.org/wiki/Municipalities_with_language_facilities), but this would probably be a good start

See also #4560

### DIFF
--- a/sources/be/bru/bosa-region-brussels-fr.json
+++ b/sources/be/bru/bosa-region-brussels-fr.json
@@ -1,0 +1,42 @@
+{
+    "data": "https://opendata.bosa.be/download/best/openaddress-bebru.zip",
+    "protocol": "http",
+    "coverage": {
+         "ISO 3166": {
+             "alpha2": "BE-BRU",
+             "state": "Brussels-Capital Region",
+             "country": "Belgium"
+         },
+         "country": "be",
+         "state": "brussels-capital region"
+    },
+    "conform": {
+        "format": "csv",
+        "srs": "EPSG:4326",
+        "encoding": "utf-8",
+        "accuracy": "3",
+        "number": "house_number",
+        "street": "streetname_fr",
+        "city":  "muncipality_name_fr",
+        "postcode": "postcode",
+        "id": {
+            "function": "join",
+            "fields": ["region_code","address_id"],
+            "separator": ":"
+        },
+        "unit": "box_number",
+        "lon": "EPSG:4326_lon",
+        "lat": "EPSG:4326_lat"
+    },
+    "compression": "zip",
+    "language": "fr",
+    "website": "https://opendata.bosa.be",
+    "license": {
+        "url": "https://cirb.brussels/fr/nos-solutions/urbis-solutions/licence-urbis-open-data"
+    },
+    "contact": {
+        "name": "Federal Public Service Policy and Support" ,
+        "email": "opendata@belgium.be",
+        "address": "Simon Bolivarlaan 30, WTC III, 1000 Brussels, Belgium"
+    }
+}
--- a/sources/be/bru/bosa-region-brussels-fr.json
+++ b/sources/be/bru/bosa-region-brussels-fr.json
@@ -14,7 +14,7 @@
         "format": "csv",
         "srs": "EPSG:4326",
         "encoding": "utf-8",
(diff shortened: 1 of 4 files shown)


sources/be/bru/bosa-region-brussels-fr.json
```diff
@@ -0,0 +1,42 @@
...
+        },
+        "unit": "box_number",
+        "lon": "EPSG:4326_lon",
+        "lat": "EPSG:4326_lat"
+    },
+    "compression": "zip",
+    "language": "fr",
+    "website": "https://opendata.bosa.be",
+    "license": {
+        "url": "https://cirb.brussels/fr/nos-solutions/urbis-solutions/licence-urbis-open-data"
```
andrewharvey: I can't read french, so only going by Google Translate here, but it appears to be similar to CC-BY, allowing any use, provided attribution is made.

Per the other optional tags for license at https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md

I think we should set:

```
"attribution": true,
"attribution name": <insert producer name here>
```

sources/be/bru/bosa-region-brussels-nl.json
```diff
@@ -0,0 +1,42 @@
...
+        },
+        "unit": "box_number",
+        "lon": "EPSG:4326_lon",
+        "lat": "EPSG:4326_lat"
+    },
+    "compression": "zip",
+    "language": "nl",
+    "website": "https://opendata.bosa.be",
+    "license": {
+        "url": "https://cirb.brussels/fr/nos-solutions/urbis-solutions/licence-urbis-open-data"
```
andrewharvey: same note as above

sources/be/vlg/bosa-region-flanders-nl.json
```diff
@@ -0,0 +1,42 @@
...
+        },
+        "unit": "box_number",
+        "lon": "EPSG:4326_lon",
+        "lat": "EPSG:4326_lat"
+    },
+    "compression": "zip",
+    "language": "nl",
+    "website": "https://opendata.bosa.be",
+    "license": {
+        "url":"https://overheid.vlaanderen.be/sites/default/files/documenten/ict-egov/licenties/hergebruik/modellicentie_gratis_hergebruik_v1_0.html"
```
andrewharvey: again just using Google Translate here, but it appears to require some kind of attribution? If so we should set atttribution: true and an "attribution name"

sources/be/wal/bosa-region-wallonia-fr.json
```diff
@@ -0,0 +1,42 @@
...
+        },
+        "unit": "box_number",
+        "lon": "EPSG:4326_lon",
+        "lat": "EPSG:4326_lat"
+    },
+    "compression": "zip",
+    "language": "fr",
+    "website": "https://opendata.bosa.be",
+    "license": {
+        "url": "https://creativecommons.org/licenses/by/4.0"
```
andrewharvey: since this is CC BY 4.0, we can set "text" to "CC BY 4.0", set attribution: true, and provide an "attribution name" per https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md

sources/be/bru/bosa-region-brussels-fr.json
barthanssens: ok thanks, I've added the tags

sources/be/wal/bosa-region-wallonia-fr.json
barthanssens: ok, done

sources/be/bru/bosa-region-brussels-nl.json
barthanssens: ok, thanks, done

sources/be/vlg/bosa-region-flanders-nl.json
barthanssens: the license does indeed require attribution, I've added the attributes

PR #4591, opened 2019-07-18
new parcel source for Faribault MN

mmdolbow:
Looks like they've launched a new website with some maps. No address points yet, but it appears that these fields in the parcel service contain physical addresses.


iandees:
@mmdolbow is this ready to merge or can we close it out for the subsequent changes that happened to these sources?

mmdolbow:
Let's close - faribault and freeborn look good in master.

PR #4609, opened 2019-08-28
Update website and data URLs for Brazoria County, Texas, in sources/us/tx/brazoria.json

FRooter:
Update website and data URLs for Brazoria County, Texas, in sources/us/tx/brazoria.json, sourced from the Brazoria County Appraisal District.


FRooter:
`PARCELS.zip` may be a more useful data source for new information as it appears to be kept more up-to-date than `PARCELS_W_OWNER.zip`

PR #4618, opened 2019-09-11
au/countrywide include lot number as number

### LINKED ISSUE #4616: Lot number in the address field for au/countrywide
There are some addresses which don't technically have a street number, instead using just a lot number, eg.

LOT 218 CORNCRAKE COURT, SOUTHERN RIVER, 6110, WA, AUSTRALIA

At the moment OA's Australia countrywide data has this point but no number so people looking for 218 Corncrake Court can't find it. Would anyone mind if we put the lot number in place of the number for these cases, in lieu of a proper field for lot number?

cc @ingalls

andrewharvey:
closes #4616 

The justification here is that in these cases where there is a lot number only that number is being used for addressing. Ideally we'd have a different field for lot number, but until then I think this is the best interim solution.

### DIFF
--- a/scripts/au/gnaf.sh
+++ b/scripts/au/gnaf.sh
@@ -47,7 +47,7 @@ SELECT
         CASE WHEN number_first IS NOT NULL THEN number_first ||
             CASE WHEN number_last IS NOT NULL THEN '-' || number_last || ' ' ELSE ' ' END
         ELSE
-            CASE WHEN lot_number IS NOT NULL THEN lot_number ELSE NULL END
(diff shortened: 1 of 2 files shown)


andrewharvey:
We could include change this so the number is "LOT N" instead of just "N", that would help data consumers know this is a lot number not a regular street number.

andrewharvey:
This should improve results for ~300,000 out of ~15M points.
 
```
> select count(*) from gnaf.addresses where lot_number is not null and number_first is null;
311,228
```

andrewharvey:
Someone must have sped up the job runner :smile: these normally take at least 9 hours...

The results look good to me, I'll leave this PR open in case there is any feedback.

PR #4631, opened 2019-09-25
fix: update utah address point data

steveoh:
for some reason our data is not showing up on the map. Let's try these new config params and see if that helps. Please provide feedback if it is indeed accurate.


steveoh:
That looks better! 🎉

slibby:
Awesome!!

iandees:
Looks good, thank you!

PR #4634, opened 2019-09-26
update estonian file path.

Steckenbauer:
In 2017 Estonia had administrative reform.
the csv file location also changed.
csv format is the same.


iandees:
Thanks for making this change, @Steckenbauer. You changed the `website` attribute, but the download system will only look at `data`. When I tried to download the file you listed in `website`, it didn't seem to exist. Instead, I went to the site originally listed in `website` and downloaded the more recent version of the address data from 2019-09-10.

Steckenbauer:
As I look at the renewed data from the zip, then the data is indeed renewed

PR #4641, opened 2019-10-04
Added us/ct/new-milford source

MiniCodeMonkey:
Statewide CT dataset had slightly incorrect locations for points on Duncan Ln, this local dataset has better data.

### DIFF
--- a/sources/us/ct/town_of_new_milford.json
+++ b/sources/us/ct/town_of_new_milford.json
@@ -1,0 +1,29 @@
+{
+    "coverage": {
+        "US Census": {
+            "geoid": "0952560",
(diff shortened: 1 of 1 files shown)


trescube:
I think you'll need to prefix the file with `HostedDownloadableData/`

MiniCodeMonkey:
@trescube Hehe yeah, indeed! Just updated, thanks. Unfortunately this dataset only had parcels, but looks like things came out great

I just did an overlay between the statewide CT source and this one.

Red = Statewide Source
Blue = New Milford Source

Can confirm that the problematic addresses on Duncan Ln look spot on now.

Duncan Ln

This is good to merge in my end.

PR #4650, opened 2019-10-11
Updated field references due to a schema change

pmacMaps:
Cumberland County, PA, changed the schema for their address points layer this week.  This PR updates the fields to the new fields in the dataset.

I've also added contact information and additional fields that now exist in the dataset.


iandees:
Great! Thanks for filling out contact info, too!

PR #4658, opened 2019-10-22
Various Pennsylvania Counties

pmacMaps:
- Added back in "city" attribute for some counties that I removed in previous PR's
- added contact information for some counties (they don't have address points web services)


andrewharvey:
Thanks @pmacMaps. Let's try to get OA to re-run this as it's failed for one of the sources. You can do this by updating the commit `git commit --amend` then `git push --force`, and hopefully it should re-run.

Issue #3747, opened 2017-12-14
pl/* Poland all sources


groupnet:
Hi,

to my understanding the sources of Poland weren't updated in years because of an _unfixable character encoding issue_. Eg. see "note" in [sources/pl/pomorskie.json](https://github.com/openaddresses/openaddresses/blob/aa65726b68397f96ab9002e7ed2451532293ef41/sources/pl/pomorskie.json)
> "note": "original source is ftp://91.223.135.109/prg/punkty_adr.zip but zipfile has unfixable character encoding issues"

Actually, the datasets are updated frequently and the original source (ftp://91.223.135.109/prg/punkty_adr.zip) is still valid. My guess is that the mentioned _character encoding issue_ can be easily solved by extracting the zip file with the -O flag.

Under Linux this should work:
`unzip -O punkty_adr.zip`

Who can check if this solves the issue?

Jürgen.

PR #4668, opened 2019-11-02
update Poland sources #3747

macieg:
Tried to address the issue below:

https://github.com/openaddresses/openaddresses/issues/3747

Poland hasn't been updated for years, because of a problem with the file name encoding.
Also, there is another link for the same data that is updated regularly.

One thing is missing - updating that data to https://data.openaddresses.io/cache/...


andrewharvey:
Thanks! Per https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#core-tags

> If there isn't a good URL for the source, members of the OpenAddresses GitHub organization can upload files to https://results.openaddresses.io/upload-cache which provides a cached URL.

@openaddresses/administrators could we make @macieg a contributor so they can upload to the cache?

macieg:
The issue should be solved without the need for using the code in this PR.
Polish authorities fixed the issue on their side.
Anyway, I'd be happy to update new data to the cache :)

andrewharvey:
@macieg If that's the case then do we need a cache at all, can you just put the new upstream HTTP url straight into the data field?

macieg:
@andrewharvey good point, there is no need to put that into the cache.

Thanks @iandees for contributing

However, I see that for some reason it didn't work. Do you know the reason?

Error:
zipfile.BadZipFile: File is not a zip file

is a little bit mysterious to me

iandees:
That error means that the file downloaded at that URL was not a ZIP file. It probably was an HTML page or an empty response (i.e. the server blocked the downloader).

macieg:
@iandees The link is fine and works. Can we rerun the build?

iandees:
Rerun this please.

macieg:
Finally had some time to take a closer look :)

The structure changed significantly and it's not possible to make an automatic conversion with available conform methods. Hence I've provided my proposal (which works!).

andrewharvey:
> Finally had some time to take a closer look :)
> 
> The structure changed significantly and it's not possible to make an automatic conversion with available conform methods. Hence I've provided my proposal (which works!).

Great thanks! Unfortunately the current architecture of OA doesn't support running `scripts` automatically for you so they need to be manually run and the results uploaded to a cache and then the cache URL needs to be used as the source URL.

Per Per https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#core-tags

> If there isn't a good URL for the source, members of the OpenAddresses GitHub organization can upload files to https://results.openaddresses.io/upload-cache which provides a cached URL.

You're not required to use this cache, but it's there if you want to.

macieg:
@andrewharvey I've processed the data and it's available here - https://drive.google.com/open?id=1OpBtXKbIlq48H8aZnSyAQHh5vLxCFDHu :)

andrewharvey:
@macieg looks like you're getting closer. If the output is CSV now you'll need to update each of those source .json files to reflect it's using CSV not GML now.

@iandees or @ingalls if @macieg can be given access to upload to the cache they can use that for a stable endpoint for the data.

macieg:
@andrewharvey the link I've provided refers to csv in the expected format. I'm happy to upload that data to the cache. :)

andrewharvey:
Yep but for instance https://github.com/openaddresses/openaddresses/pull/4668/files#diff-a9ed9846e57273550d6311f29f9fdcd9R32 still references XML, so all these source json files will need to be updated based on the CSV schema in your cached/processed file. Does that make sense?

macieg:
@andrewharvey makes sense :) I've updated json files.

macieg:
@andrewharvey can I upload the data to the cache now?

iandees:
I apologize – I haven't had time to fully understand what's going on in this ticket.

Can you describe what prevents us from downloading directly from the source?

macieg:
@iandees no problem :)

XML schema has changed since it's was uploaded for the first time few years ago.
With current source processor design and methods available we are no capable to process xml files from the source properly.

To be precise, one address record from the source xml consists of such elements:
`      <prg-ad:jednostkaAdmnistracyjna>Polska</prg-ad:jednostkaAdmnistracyjna>
      <prg-ad:jednostkaAdmnistracyjna>**lubelskie**</prg-ad:jednostkaAdmnistracyjna>
      <prg-ad:jednostkaAdmnistracyjna>**bialski**</prg-ad:jednostkaAdmnistracyjna>
      <prg-ad:jednostkaAdmnistracyjna>Leśna Podlaska</prg-ad:jednostkaAdmnistracyjna>
`
We have four identical _jednostkaAdmnistracyjna_ elements. **Bolded** ones are a district and a region respectively. There is no conform method that allows to pick second or thirds element from an array.

Ideally, new method should be added.

Does it make sense?

iandees:
I added you to the organization, @macieg. You should be able to use https://results.openaddresses.io/upload-cache to upload data now. Once those are uploaded, you'll want to update the source files here with the URL that the uploader gives you.

iandees:
> no capable to process xml files from the source properly

Thanks, this helps. I forgot that this was XML data!

PR #4684, opened 2019-11-27
Update fi countrywise data

vesameskanen:
- Compression changed, so we need to host an intermediate data package in zip form
- As a bonus, the package name is now fixed, so dataloading will not break any more
- We started a scheduled process to keep the data updated as a part of national digitransit service


ingalls:
Fantastic work! Thanks for the contribution

PR #4693, opened 2019-12-24
Update France source

jdesboeufs:
Next week the [French Addresses database](https://adresse.data.gouv.fr) will switch to the French equivalent to the CC-BY.
Share-alike is no more needed.

The license in English: https://www.etalab.gouv.fr/wp-content/uploads/2018/11/open-licence.pdf

In this PR we also switch to the most up-to-date data.

Please do not merge until 2020 :)


jdesboeufs:
`gzip` compression is not supported.
How can I register our `csv.gz` source?
I have to submit a script?

jdesboeufs:
I finally provided a specific route where the CSV file is not compressed.

nvkelso:
Huzzah!

jdesboeufs:
Can anybody explain me at which frequency the data will be updated?

andrewharvey:
> Can anybody explain me at which frequency the data will be updated?

For example at http://results.openaddresses.io/sources/fr/ain you can see it's usually processed every fortnight.

> Please do not merge until 2020 :)

I've added the "in progress" label to try to make that clearer. You could also use a draft PR if you're not wanting it to be merged yet, but it's not a big deal.

sources/fr/countrywide.json
```diff
@@ -0,0 +1,33 @@
...
+      },
+      "street": "nom_voie",
+      "lon": "lon",
+      "lat": "lat",
+      "city": "nom_commune",
+      "postcode": "code_postal"
+  },
+  "email": "adresse@data.gouv.fr",
+  "language": "fr",
+  "attribution": "French addresses database (adresse.data.gouv.fr)",
```
andrewharvey: See https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md attribution is deprecated, it's better to use `attribution name` within the `license` key.

sources/fr/countrywide.json
```diff
@@ -0,0 +1,33 @@
...
+        "fields": ["numero", "rep"],
+        "separator": ""
+      },
+      "street": "nom_voie",
+      "lon": "lon",
+      "lat": "lat",
+      "city": "nom_commune",
+      "postcode": "code_postal"
+  },
+  "email": "adresse@data.gouv.fr",
```
andrewharvey: Per https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md, The email field says `Do not set unless the data provider wants to receive updates.` his the data provider opted in to this? Otherwise if it's just a generic contact email, then best to put this in the `contact` key.

jdesboeufs: We can accept this kind of notification in the future.
But since it's not implemented for now, I just moved to the contact email.

jdesboeufs:
Perfect!

Since the next database processing will be after the 1st of January, you can merge when you think it's ready.

andrewharvey:
> Since the next database processing will be after the 1st of January, you can merge when you think it's ready.

Oh except if it's merged now it will just grab the last run from the PR and use that. So if the run needs to be done in 2020, we'll need to rerun this in 2020 and then merge.

jdesboeufs:
There no problem in using current data, but they should not be released in attribution-only before 1/1/2020.

jdesboeufs:
You have my final Go! ;)
Happy New Year for everybody! 🍾

iandees:
🎉

PR #4723, opened 2020-01-06
Added more up-to-date source for us/ca/santa_clara

MiniCodeMonkey:
This uses data directly from Santa Clara County instead of a static dump from 2014


MiniCodeMonkey:
Just verified the output. Looks good to me 👍

PR #4730, opened 2020-01-07
Update goodhue.json

Garrett-W:
Fix & Update to Address Points


Garrett-W:
Fixed links to another 2 counties in MN

ingalls:
:taco: As always, thanks a million!

PR #4736, opened 2020-01-15
Update peel-region.json

kevin-peel:
We've switched our open data portals, this update points to the new data source and format.

Changed source from old portal (different web address, Zipped CSV) to new portal based on Esri platform.  Also included more information (license, language, website, etc.).


ingalls:
@kevin-peel, thanks a million for putting this together, and being one of the very few government orgs in Ontario that have such a commitment to open data.

PR #4776, opened 2020-02-04
Added missing part of street in us/co/statewide

MiniCodeMonkey:
`PreType` was missing causing things like "County Road 100" (Cr 100) to be added as just "100"


iandees:
The image preview looks weird here, but the slippy map preview looks ok.

PR #4777, opened 2020-02-04
Update URL for San Juan County WA addresses

nickpeihl:
The API url has changed for this county, so I've updated it here.


iandees:
Thanks for updating the URL!

Issue #4060, opened 2018-06-09
City and County of San Francisco source change


jasonlally:
In reference to the data source: https://github.com/openaddresses/openaddresses/blob/f090841936bd20566a7a7498b2ac1aa0296e2ab9/sources/us/ca/san_francisco.json#L13

We have an internal ETL that extracts and syncs the shapefile to a table with point geometry on the portal. Over time we are looking to swap in an internal ETL that connects closer to the source to make the ETL more resilient and easier to monitor. The zipped shapefile isn't going away anytime soon, but I just wanted to see if it would make sense to hook this up to a new endpoint. The data is the same, the endpoints change.

The portal page is here: https://data.sfgov.org/Geographic-Locations-and-Boundaries/Addresses-with-Units-Enterprise-Addressing-System-/dxjs-vqsy

The CSV bulk link is here: https://data.sfgov.org/api/views/dxjs-vqsy/rows.csv?accessType=DOWNLOAD

The GeoJSON bulk link is here: https://data.sfgov.org/api/geospatial/c59h-mc8k?method=export&format=GeoJSON

The Shapefile bulk link is here: https://data.sfgov.org/api/geospatial/c59h-mc8k?method=export&format=Shapefile

1. Is there a preference on source type w/r/t the ETL engine for open addresses?
2. Does it make sense to swap in the new endpoint? There's no end of life yet on the zip shapefiles at the existing source, but by switching any changes to data will be documented and communicated at the portal level.
3. I'm happy to put in the PR if it makes sense

jasonlally:
On second look, I would not use the shape file bulk link because of the truncation issue. The CSV and GeoJSON will retain the full names in the schema.

arch0345:
Resolved in #4778

PR #4778, opened 2020-02-04
Update San Francisco source

jasonlally:
This change updates the source for San Francisco addresses. Previously it was updating against a shapefile, but this file is deprecated in favor of a nightly updated dataset at https://data.sfgov.org/Geographic-Locations-and-Boundaries/Addresses-with-Units-Enterprise-Addressing-System/ramy-di5m

The endpoint to the data is a bulk geojson dataset of all the addresses published from San Francisco's Enterprise Address System. Added the Enterprise Address System in the attribution source for clarity in case additional sources of address data are added for San Francisco.

Also updated the license to Open Data Commons Public Domain Dedication and License (which is the license applied to all of San Francisco's open data).

Addresses: #4060

### DIFF
--- a/sources/us/ca/san_francisco.json
+++ b/sources/us/ca/san_francisco.json
@@ -31,8 +31,6 @@
           "street_name",
           "street_type"
       ],
-      "city": "San Francisco",
-      "region": "California",
       "unit": "unit_number",
       "postcode": "zip_code",
       "id": "eas_fullid"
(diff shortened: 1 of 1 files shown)


jasonlally:
Couldn't get the geojson source to work, but that's okay. There's a shapefile endpoint off the same underlying data. Looks like it passed all the checks.

Would love to get this merged in so we have the freshest data going to open addresses!

iandees:
Thanks for the update!

Issue #1979, opened 2016-09-15
Add addresses in suburbs around Philadelphia, PA


nvkelso:
Status:
- [ ] **Research sources** and propose best source (and it's data license)
- [ ] **Review proposed source**
- [ ] **Approve source** based on review comments
- [ ] **Add** to OpenAddresses

_NOTE: Proposed source notes should include basic column mapping to OpenAddress data model._

davidchiles:
Looks like the two counties missing are Delaware and Montgomery Counties. 

This is as close as I can get to the data on Montgomery County: http://propertyrecords.montcopa.org/pt/arcgismaps/arcgismapadvanced.aspx

And I'm not sure what's going on in Delaware County: http://www.co.delaware.pa.us/planning/mapping/GISDataLayers.html

justinelliotmeyers:
Deleware Co update - DelmutoJ@co.delaware.pa.us - There is a regional NG911 group that will be addressing this data so it’s possible it will happen sooner than later but no guarantee. 
I’ll keep you posted if anything changes.
-Julie

justinelliotmeyers:
Montgomery reply:
Hi Justin,
 
You will need to speak with our IT Department. We do not have address points in shapefile as of this time, but it is something we are working towards. However, our address data is available through our IT Data Sales: http://www.montcopa.org/439/Data-Sales

Costs here: http://www.montcopa.org/DocumentCenter/View/2498

nvkelso:
Well, $20k / year fee isn’t great.

But can you ask them if we have rights after that to use commercially and
to redistribute, please?

nvkelso:
Still a hole.

justinelliotmeyers:
followed up with Deleware Co update - DelmutoJ@co.delaware.pa.us

justinelliotmeyers:
emailed boahelp@montcopa.org asking about TOU after purchase

justinelliotmeyers:
no data

PR #4783, opened 2020-02-14
2 new sources + 2 source updates

MiniCodeMonkey:


sources/ca/bc/nanaimo.json
```diff
@@ -5,17 +5,20 @@
         "city": "Nanaimo",
         "geometry": { "type": "Point", "coordinates": [-123.96, 49.17] }
     },
-    "data": "https://www.nanaimo.ca/open-data-catalogue/Download/Index?container=nanaimo&entitySetName=Addressing&downloadID=74",
-    "website": "https://www.nanaimo.ca/open-data-catalogue",
+    "data": "https://nanmap.nanaimo.ca/arcgis/rest/services/NanMap/Polygons/MapServer/15",
```
MiniCodeMonkey: The existing source was missing some streets, specifically `650 Terminal Ave`, which is present in this source. This source also appears to be continuously updated.

sources/us/fl/miami2.json
```diff
@@ -0,0 +1,34 @@
...
+            "geoid": "12086",
+            "name": "Miami-Dade County",
+            "state": "Florida"
+        },
+        "country": "us",
+        "state": "fl",
+        "county": "Miami-Dade"
+    },
+    "attribution": "Miami-Dade County",
+    "data": "https://gisweb.miamidade.gov/arcgis/rest/services/MD_Emaps/MapServer/72",
```
MiniCodeMonkey: This source has addresses that the other source does not have. I was debating what to do here, as this source actually have slightly fewer addresses in total -- so I decided to keep them both. Let me know what you guys think.

Specifically, I was looking at `16055 SW 117th Ave` which is present in this source but not in the source from `us/fl/miami.json`

sources/us/in/hamilton.json
```diff
@@ -9,8 +9,9 @@
         "state": "in",
         "county": "Hamilton"
     },
-    "data": "http://gis.hamiltoncounty.in.gov/ArcGIS/rest/services/E911/MapServer/0",
-    "protocol": "ESRI",
+    "data": "ftp://ftp.hamiltoncounty.in.gov/GIS/Data/Address Points/E911_Addresses_web.zip",
```
MiniCodeMonkey: Based on https://results.openaddresses.io/sources/us/in/hamilton the previous source appears to be stale. This file however looks to be updated continuously (it was last updated today).

The new source has `1021 Domino Dr` and the old source did not (this is the reason that I looked into this)

sources/us/pa/montgomery.json
```diff
@@ -0,0 +1,26 @@
...
+        "US Census": {
+            "geoid": "42091",
+            "name": "Montgomery County",
+            "state": "Pennsylvania"
+        },
+        "country": "us",
+        "state": "pa",
+        "county": "Montgomery"
+    },
+    "data": "https://opendata.arcgis.com/datasets/7cdc3fca3262445499ba1ee9422ba0e1_0.zip",
```
MiniCodeMonkey: Previous discussions on https://github.com/openaddresses/openaddresses/issues/1979 -- seems like a lot has happened since 2016. Yay!

iandees:
It looks like the second Maiami-Dade source isn't finishing, but I'm ok with having two and maybe it will finish later.

MiniCodeMonkey:
@iandees Not sure what happened there, it processes fine when I ran machine in single mode locally. Perhaps a timeout issue?

PR #4785, opened 2020-02-17
Update city_of_saint_johns.json

JosephKuchar:
The "house number" attribute is in the CIVIC field, STREETNUM is a unique identifier for the street.


iandees:
Thanks!

PR #4796, opened 2020-02-19
Update saskatoon.json - New temporary data source

JosephKuchar:
The Saskatoon data is not being processed correctly because of html embedded in the kml file. I've parsed it with BeautifulSoup and have a temporary csv on the CSBP Github, hopefully this works until a more permanent processing solution can be developed.
This is also my 2nd attempt to merge this branch, forgot to specify the geocoordinates last time.

### DIFF
--- a/sources/ca/sk/saskatoon.json
+++ b/sources/ca/sk/saskatoon.json
@@ -11,13 +11,12 @@
         "state": "sk",
         "city": "Saskatoon"
     },
-    "data": "https://github.com/CSBP-CPSE/LODE-ECDO/raw/master/scripts/Addresses/Saskatoon.csv",
+    "data": "https://data.openaddresses.io/cache/uploads/andrewharvey-openaddr/783410/Saskatoon.csv",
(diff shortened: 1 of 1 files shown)


andrewharvey:
Looks like OA is having trouble setting the filename when downloading from GitHub.

I've uploaded this to the cache at `https://data.openaddresses.io/cache/uploads/andrewharvey-openaddr/783410/Saskatoon.csv` if you wanted to use that URL instead?

iandees:
You should only need to specify the filename if there's more than one file. Try removing the filename parameter altogether?

JosephKuchar:
> Looks like OA is having trouble setting the filename when downloading from GitHub.
> 
> I've uploaded this to the cache at `https://data.openaddresses.io/cache/uploads/andrewharvey-openaddr/783410/Saskatoon.csv` if you wanted to use that URL instead?

Thanks! This, and removing the filename parameter, seems to have done the trick.

PR #4798, opened 2020-02-21
Added 2 Canadian sources

MiniCodeMonkey:
Note: The Kimerbley source *should* process correctly after this has been merged in https://github.com/openaddresses/machine/pull/763

### DIFF
--- a/sources/ca/bc/city_of_cranbrook.json
+++ b/sources/ca/bc/city_of_cranbrook.json
@@ -11,6 +11,7 @@
             ]
         }
     },
+    "website": "https://map.cranbrook.ca/Html5Viewer/?viewer=CityViewer",
     "data": "https://map.cranbrook.ca/arcgis/rest/services/Updated_Mainmap/MapServer/31",
     "protocol": "ESRI",
     "conform": {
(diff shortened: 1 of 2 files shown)


andrewharvey:
What license is this data made available under?

MiniCodeMonkey:
That's a great questions! Both of these are published as open data without warranty. I have not been able to find a specific license attached to them.

For Cranbrook, here is the disclaimer on map load:

```
This map is for general purposes only.
The City of Cranbrook makes no warranties regarding the
accuracy or suitability of the map for any purpose.
This map is not intended for navigation or legal use or purposes
and should not be relied on as such.
The City of Cranbrook will not be liable for any damage, loss
or injury resulting from the use of the map or information
on the map.
The map may be changed at any time.
```

nvkelso:
That’s for the web app. Is there data specific license? Or email contact?

MiniCodeMonkey:
The web app is the source of this data source. Kimberley data was provided by Christopher Jones from their planning department. No specific license is attached with this data that I have been able to find.

MiniCodeMonkey:
These are both looking good now 👍

PR #4800, opened 2020-02-25
Update montreal.json

### LINKED ISSUE #4790: Montreal, Quebec - Possible to separate address ranges?
Hi All,

I noticed that the Montreal source file (https://github.com/openaddresses/openaddresses/blob/master/sources/ca/qc/montreal.json) uses the "TEXTE" column for the street number, which actually contains slightly weirdly formatted street number ranges (e.g., 108-112 would be written \# 108-12). However, the original source file also contains ADDR_DE and ADDR_A columns, which are essentially start and end street numbers for each entry.
Would it be worthwhile to add a step in processing for this file to use those columns and duplicate rows? So that instead of a single \# 108-12 entry, you would have separate rows for 108, 110, and 112 all with identical geocoordinates? This is something I plan to do for my own purposes (separate entries will make it easier to match addresses in my scripts), but maybe there's a clever way to do it so that it can be used by OpenAddresses as well.
Cheers,
Joseph

JosephKuchar:
New source file for Montreal where street ranges have been separated into individual rows (
Fixes #4790).

### DIFF
--- a/sources/ca/qc/montreal.json
+++ b/sources/ca/qc/montreal.json
@@ -21,6 +21,8 @@
             "GENERIQUE",
             "SPECIFIQUE"
         ],
+        "lon":"LON",
+        "lat":"LAT",
         "file":"Montreal.csv",
         "format": "csv"
     }
--- a/sources/ca/qc/montreal.json
+++ b/sources/ca/qc/montreal.json
@@ -21,9 +21,9 @@
             "GENERIQUE",
             "SPECIFIQUE"
         ],
-        "lon":"LON",
-        "lat":"LAT",
-        "file":"Montreal.csv",
+        "lon": "LON",
+        "lat": "LAT",
+        "file": "Montreal.csv",
         "format": "csv"
     }
 }
--- a/sources/ca/qc/montreal.json
+++ b/sources/ca/qc/montreal.json
@@ -5,7 +5,7 @@
         "city": "Montreal",
         "geometry": { "type": "Point", "coordinates": [-73.63, 45.51] }
     },
-    "data": "http://donnees.ville.montreal.qc.ca/dataset/63317d72-426e-41fa-bfef-b0f28bb06ce1/resource/393e5c01-f94b-4190-9a77-fcf9e4b7b394/download/adresse.zip",
+    "data": "https://github.com/CSBP-CPSE/LODE-ECDO/raw/master/scripts/Addresses/Montreal.zip",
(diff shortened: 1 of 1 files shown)


JosephKuchar:
Processing script and processed data available here: https://github.com/CSBP-CPSE/LODE-ECDO/tree/master/scripts/Addresses

MiniCodeMonkey:
Hi there! Thanks for spending time on this and sharing it.

Just a quick question on the logic. Does this mean that the same lat/lon is used for all house numbers in the range? I.e. there's no range interpolation or anything like that?

I am inclined to argue that it might be best to keep the source as-is for two reasons:

* By using a static (and external) script to generate the data, we will no longer be able to get live updates from the source -- it would manually have to be updated each time
* I would leave it up to the consumer of the source to decide how to go about dealing with address ranges (i.e. I already have specific logic in place that processes ranges shorter than a set interval and skips the rest because they are too coarse)

JosephKuchar:
> Hi there! Thanks for spending time on this and sharing it.
> 
> Just a quick question on the logic. Does this mean that the same lat/lon is used for all house numbers in the range? I.e. there's no range interpolation or anything like that?

The script I've applied does no interpolation - I'd say that interpolation introduces uncertainty, whereas the 'splitting' I've done is just an equivalent representation of the data that's already there. If someone wanted to get back to ranges, they could do so fairly easily.

> I am inclined to argue that it might be best to keep the source as-is for two reasons:

There's a bit of a discussion on this in the related issue, which you may want to see if you haven't already: https://github.com/openaddresses/openaddresses/issues/4790

> * By using a static (and external) script to generate the data, we will no longer be able to get live updates from the source -- it would manually have to be updated each time

Yup, that's true. 
> * I would leave it up to the consumer of the source to decide how to go about dealing with address ranges (i.e. I already have specific logic in place that processes ranges shorter than a set interval and skips the rest because they are too coarse)

My issue with this is that it places a burden on the final consumer of the data to go back to the original source to see what the text in the NUMBER column actually represents. For example, entries like "#6902-04" or "#5807-09A" could easily be interpreted as Street Number - Unit pairs rather than street number ranges. Before the processing is when it's clear what these things are, so it's at that point that I'd say it makes the most sense to parse them. 
Of course it can be processed after the fact, but it's also simplest to do at this stage because the original data has columns that specify the start and end numbers explicitly so that they don't have to be imputed from the text string.

PR #4801, opened 2020-02-25
Updated us/nc/statewide to use more current source

MiniCodeMonkey:
The previous source was a static dump that presumably was a snapshot of addresses from 2014

### DIFF
--- a/sources/us/nc/statewide.json
+++ b/sources/us/nc/statewide.json
@@ -7,27 +7,22 @@
         "country": "us",
         "state": "nc"
     },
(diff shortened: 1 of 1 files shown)


ingalls:
:tada: Excited for this one, didn't realize it was frozen to 2014

MiniCodeMonkey:
The only downside here is that we loose a bit of data granularity as the previous source had zip codes for addresses (the zip code here appears to be unpopulated for now) -- and street address was split out from the house number. But oh well...

iandees:
👏

PR #4809, opened 2020-02-27
Update city_of_lethbridge.json

JosephKuchar:
Add license information

### DIFF
--- a/sources/ca/ab/city_of_lethbridge.json
+++ b/sources/ca/ab/city_of_lethbridge.json
@@ -11,14 +11,15 @@
             ]
         }
     },
-    "data": "http://gis.lethbridge.ca/lethwebgisarcgis/rest/services/OpenData/odl_parcels/MapServer/0",
+    "data": "https://opendata.arcgis.com/datasets/16e189882ac94a5091618ee0e0f74ce7_0.zip",
     "license": {
         "url": "https://www.lethbridge.ca/Pages/OpenDataLicense.aspx",
         "attribution": false
     },
-    "protocol": "ESRI",
+    "protocol": "http",
+    "compression": "zip",
     "conform": {
-        "format": "geojson",
+        "format": "shapefile",
         "accuracy": 2,
         "number": {
             "function": "prefixed_number",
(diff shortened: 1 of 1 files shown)


sources/ca/ab/city_of_lethbridge.json
```diff
@@ -12,6 +12,8 @@
         }
     },
     "data": "http://gis.lethbridge.ca/lethwebgisarcgis/rest/services/OpenData/odl_parcels/MapServer/0",
+    "license": {"url": "https://www.lethbridge.ca/Pages/OpenDataLicense.aspx",
+               "attribution":false},
```
iandees: Please follow the existing JSON formatting on these.

```suggestion
    "license": {
        "url": "https://www.lethbridge.ca/Pages/OpenDataLicense.aspx",
        "attribution": false
    },
```

JosephKuchar: Sure - I'll make an effort to make sure I'm following the right format. For my own clarification, when do you use 
license : {url: ..., attribution: ...}, 
and when do you use 
license: licenseurl
attribution: ...

That is, when is license its own dictionary, and when is it not?

iandees: You can read more about it here: https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#optional-tags

The URL-only `license` attribute is supported but deprecated. Ideally it would always be a dictionary.

JosephKuchar: Okay, sounds good. Do you use the attribution name field to include text like "Contains information licensed under..." or do you simply flag it as attribution required?

JosephKuchar:
The issue seems to be with the processing of the ESRI API - "2020-02-27 19:14:21,118  WARNING: Could not download ESRI source data"
I noticed Lethbridge is missing from my recent OA download, so I think the source itself is possibly broken (though it looks fine to me).
The Lethbridge Open Data Portal hosts the same data as a CSV or shapefile: http://opendata.lethbridge.ca/datasets/16e189882ac94a5091618ee0e0f74ce7_0/data

iandees:
Looks good!

Issue #4819, opened 2020-03-04
update address source for Municipality of Anchorage


mknapp-source:
Hi there - 

Not sure how to get this done, but we need to update the source to the address data being used for Anchorage, Alaska.

Old info (from 2018)-->  "data": "http://munimaps.muni.org/ftp_gis/address_shp.zip",
"website": "http://www.muni.org/Departments/it/GIS2/Pages/MOAGISData.aspx",

Our prefered new source is a feature service hosted in ArcGIS Online

New info (updated daily)-->  "data": "https://services2.arcgis.com/Ce3DhLRthdwbHlfF/ArcGIS/rest/services/Addresses_Hosted/FeatureServer/0",
"website": "https://moa-muniorg.hub.arcgis.com/pages/data",

Thanks!

Mike Knapp
Municipality of Anchorage

iandees:
Hi Mike! Thanks for letting us know. I'll go change that now.

mknapp-source:
Thanks Ian….

I also noticed that the “house_num” and “half_add” fields should now be “HOUSE_NUMBER” and “HALF_ADDRESS”.
And “zipcode” should now be “ZIPCODE”.

Thanks for the super fast response.

-Mike

iandees:
Thanks again for telling us about the change!

PR #4820, opened 2020-03-04
Update Anchorage

iandees:
For #4819.


mknapp-source:
Awesome, thanks again.

PR #4823, opened 2020-03-08
Update San Bernardino download link

nickpeihl:
It's been almost two years and the old link is long gone. 

😞 I wish they weren't so ephemeral.


iandees:
Thanks for updating the URL.

PR #4825, opened 2020-03-10
Add Whatcom County, WA

nickpeihl:
I confirmed with the Whatcom County GIS Administrator that the City of Bellingham distributes the E911 address location dataset for the entire Whatcom County.


iandees:
Thanks for adding this!

Issue #4831, opened 2020-03-16
BVI Addresses


justinelliotmeyers:
https://www.bvimapping.gov.vg/arcgis/rest/services/MAPS/ADDRESSES/FeatureServer

PR #4832, opened 2020-03-17
British Virgin Islands

ingalls:
Closes: https://github.com/openaddresses/openaddresses/issues/4831


ingalls:
Screenshot from 2020-03-17 07-08-39

The number field function I used must be case sensitive

PR #4839, opened 2020-03-21
reverting to Shapefile source for Broward as ESRI source no longer exists

albarrentine:
Shapefile has been updated for 2020


arch0345:
This has since been updated in #6490

PR #4841, opened 2020-03-24
Added us/ny/suffolk

MiniCodeMonkey:
This captures some addresses that are not present in the NY statewide data source


MiniCodeMonkey:
@iandees Thanks for fixing my botched field names 👍

iandees:
Thanks for adding this!

PR #4853, opened 2020-03-26
data url changed for Estonia

iSudo:
latest address data can now be fetched from here

### DIFF
--- a/sources/ee/countrywide.json
+++ b/sources/ee/countrywide.json
@@ -3,7 +3,7 @@
         "country": "ee",
         "ISO 3166": {"alpha2": "EE", "country": "Estonia"}
     },
-    "data": "https://data.openaddresses.io/cache/uploads/iandees/adc2bf/valjavote.zip",
+    "data": "https://api.dev.peatus.ee/estonia-addresses",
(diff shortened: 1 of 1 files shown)


iSudo:
i made changes in target service, now should work as expected

iSudo:
@iandees can you please check this and possibly merge? We built this service in order to provide latest address data for openaddresses, and we are also in need of the latest openaddresses data.

iandees:
I appreciate the work you put into making this work, but I'm a little bit worried about pointing to an unofficial data source. It looks like the service you built is proxying or redirecting the download? How often does the data update?

iSudo:
Hi, yes, the download is being proxied, its not the original place of addresses, true, but it is official, I'm working with the Estonian Land Board, and this is what ive come up with. Our national addresses are being updated approximately once a month.

iSudo:
Is there something else I could do to make this pullrequest acceptable?

iandees:
Sorry, I was still thinking about it. I'll merge it for now but would prefer to see a more direct link.

iSudo:
Thanks, when can we expect this to reflect in results.openaddresses ?

iandees:
The system refreshes data at least once a week, so probably by this weekend.

PR #4919, opened 2020-04-10
Update New York state URL

iandees:


arch0345:
This has since been updated in #6246

PR #4962, opened 2020-04-20
fix(us-nc-statewide): Updated source to be zipped archive instead of MapServer

MiniCodeMonkey:
This appears to be the same data just in a single gdb file instead! This should fix the timeout issues we were having.

### DIFF
--- a/sources/us/nc/statewide.json
+++ b/sources/us/nc/statewide.json
@@ -7,11 +7,14 @@
         "country": "us",
         "state": "nc"
     },
-    "data": "https://services.nconemap.gov/secure/rest/services/NC1Map_Parcels/MapServer/0",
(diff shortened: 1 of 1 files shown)


iandees:
Thanks for updating this. Do you have any indication about how often this file is updated?

MiniCodeMonkey:
Unfortunately, I do not. I guess it would be easy to see if it's continuously updated after adding this to OA (the address # should change for every batch run or so).

The link is from this page:
https://www.nconemap.gov/pages/parcels

I assume that it would update as often as the individual parcel snapshots.

I agree that getting it directly from the MapServer would be ideal, but I think it's just too much data for it to be viable option without timeouts (it's probably also causing a significant load on their servers).

PR #5020, opened 2020-05-04
Fix some MN counties

Garrett-W:
Douglas, Grant, Renville, and Yellow Medicine


iandees:
The Yellow Medicine county source doesn't seem to be picking up any attributes: https://results.openaddresses.io/runs/832803/dotmap/index.html#lat=44.71030&lng=-96.27589&z=17.7599

Can you check the attribute names on that layer please?

PR #5023, opened 2020-05-06
Snyder, Somerset, and Sullivan Counties PA Updates

pmacMaps:
Update contact info:
- Snyder County, PA
- Somerset County, PA
- Sullivan County, PA

Update map service URL and fields for Snyder County, PA


iandees:
Thanks again for updating these!

PR #5073, opened 2020-05-20
Sullivan County, PA, USA, source update.

pmacMaps:
Replace CSV file with Esri REST service for parcels.
Use function to parse single address field.


pmacMaps:
For some reason, the function combined the number and street together in the street field.

iandees:
The data looks pretty messy, but I see numbers extracted in most cases. Thanks for working on this!

PR #5080, opened 2020-05-27
Fix street fields for us/al/jefferson

MiniCodeMonkey:


MiniCodeMonkey:
`street_name` was accidentally duplicated instead of having `street_type` as the second field

Issue #4982, opened 2020-04-22
some additional Pennsylvania addresses


hhm0:
From [PASDA](https://www.pasda.psu.edu/) (searches that I used to find the data sets: [here](https://www.pasda.psu.edu/uci/SearchResults.aspx?Keyword=parcel) and [here](https://www.pasda.psu.edu/uci/SearchResults.aspx?Keyword=address))
- Indiana County
  - source: https://www.pasda.psu.edu/uci/DataSummary.aspx?dataset=1943
  - type: parcels
  - fields:
    - STATE: two digit state
    - ZIP:  9 digit zip code (nnnnn-nnnn)
    - CITY
    - ADDRESS_1: usual address field, sometimes has PO box or similar
    - ADDRESS_2: usually empty, at times has address when ADDRESS_1 has PO box, or has PO box when ADDRESS_1 has address, sometimes has additional information on ADDRESS_1 (e.g. suite number)
    - COMBINED_S: usually same as ADDRESS_1, sometimes has partial address etc, occasionaly has valid address when others have invalid address
- Lycoming County
  - source: https://www.pasda.psu.edu/uci/DataSummary.aspx?dataset=1798
  - type: address points
  - fields:
    - LAT: latitude, decimal
    - LON: longitude, decimal
    - NUMBER: house number
    - NAME: street name without suffix
    - TYPE: street suffix/type
    - ADDRESS: house number and street name with suffix
    - MUNICIPALI: municipality
    - ZIP_CODE: 5 digit zip code
- Union County
  - source: https://www.pasda.psu.edu/uci/DataSummary.aspx?dataset=1992
  - type: address points
  - fields:
    - Lat: latitude, decimal
    - Long: longitude, decimal
    - FULLADDR: house number and street
    - Add_Number, ADDRNUM: house number
    - AddNum_Pre, AddNum_Suf, ADDRNUMSUF: empty?
    - FULLNAME: street name
    - StreetName: street name without prefix and suffix
    - St_PreDir: street (direction) prefix
    - St_PreMod, St_PreTyp, St_PreSep: usually empty
    - St_PosTyp: street (type) suffix
    - St_PostDir, St_PostMod: usually empty
    - Inc_Muni, MUNICIPALI, MSAGComm: Municipality with suffix and mixed case (e.g. Mifflinburg Boro)
    - Post_Comm: municipality in all caps without suffix, sometimes differs from MUNICIPALI etc. fields
[...]

PR #5091, opened 2020-06-12
Add Indiana County, PA

pmacMaps:


pmacMaps:
After previewing the results, I don't think this data source is usable.  It appears that when the parcel address matches the owner mailing address, then it's correct.  

But there are many features where the address appears to be an owner address from another city.

Feel free to close this PR without merging, or leaving open as reference.

pmacMaps:
This is related to #4982

ingalls:
Thanks for giving this a go @pmacMaps - bummed it didn't work out

PR #5094, opened 2020-06-18
Update Columbia County, PA:

pmacMaps:
- update URL for map service and fields
- add contact information


pmacMaps:
This county has a proxy running on their ArcGIS Server environment.  I thought maybe it would work, but it appears not. 

It was worth a shot.

PR #5104, opened 2020-07-09
Updating NC County Data

medewitt:
This pull request updates the source location for addresses for Alamance and Randolph counties in North Carolina. No real changes for Alamance other than location. There are some more substantial changes on the Randolph data.


iandees:
Thanks for the update!

medewitt:
@iandees thanks for everything you do and the fix to my jsons!

Issue #5123, opened 2020-07-23
Trying to understand why ct/statewide is broken


blackmad:
Just out of curiosity(while building a pelias index) I wanted to understand why ct/statewide was broken and if it was possible to fix it.

https://data.openaddresses.io/runs/879601/output.txt

The debug output from OA seems to be getting back an html form, but when I run the query manually from python (based on what that debug log is showing) I get a perfectly correct response. Is there a better way to understand exactly what the OA scraper is sending? I couldn't figure out instructions to run the OA scraper locally on just one source. Is there a doc that would help me?

Here's what I get in local ad-hoc testing

```
>>> r = requests.post('https://www.ctgismaps2.ct.gov/arcgis/rest/services/CT_Open_Maps/Address_Points/MapServer/0/query', {'f': 'json', 'returnGeometry': True, 'outFields': '*', 'geometryPrecision': 7, 'outSR': '4326', 'where': 'OBJECTID > 0 AND OBJECTID <= 1000'})
>>> str(r.json())[0:5000]
[...]

PR #5125, opened 2020-07-23
Switch to https

iandees:
For #5123


iandees:
For posterity, it doesn't look like this has any unit information even though the fields seem to be there.

PR #5129, opened 2020-07-28
Update contact information for various Pennsylvania counties

pmacMaps:
These are only contact information updates.  Hopefully the data sources are still good so the tests don't fail.


iandees:
Looks like Greene county data disappeared, but otherwise this looks good. Thanks for adding the contact information!

PR #5130, opened 2020-07-28
au/act buildings

andrewharvey:
@ingalls is v2 schema ready to go?

### DIFF
--- a/sources/au/act/statewide.json
+++ b/sources/au/act/statewide.json
@@ -1,4 +1,5 @@
 {
+    "schema": 2,
     "coverage": {
         "ISO 3166": {
             "alpha2": "AU-ACT"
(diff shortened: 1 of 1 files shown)


andrewharvey:
"16 284.3 ERROR: Nothing processed: 'buildings' layer not currently supported"

looks like I may have been a bit too eager with this PR.

ingalls:
@andrewharvey If we can hold off pushing v2 sources until we have completed the migration that would be great, but it's finalized. 
There is an active staging deployment here: http://staging.openaddresses.io/data The new processing code lives here: https://github.com/openaddresses/batch and the migration is being finalized here: https://github.com/openaddresses/migrate

andrewharvey:
I'll close this for now, if buildings are supported later, I can do a new PR.

ingalls:
Building should be supported within the next week, the support has been added to `batch-machine`, I just haven't wired up support for it in the schema & `batch` itself

PR #5139, opened 2020-07-30
update au/nsw to July 2020

andrewharvey:


ingalls:
Hey @andrewharvey would you have time to update to the V2 Source schema? Would love to get this merged. If you're swamped I'm happy to take it on.

andrewharvey:
I ran into an issue building the data, when doing the esri dump it broke half way, so I tried to resume but I think I either ended up with duplicate points or missing points. So when I get a chance I'll re-run the dump from scratch.

Once I do that I'll update this PR  to v2.

andrewharvey:
closing this until I have a chance to run a fresh dump again (running into timeout errors in the process and Spatial Services have launched a new endpoint which might be better, needs investigation)

PR #5182, opened 2020-08-25
NS, CA

ingalls:
Update NS Province file


ingalls:
Closes: https://github.com/openaddresses/openaddresses/issues/5245

PR #5212, opened 2020-08-26
Fix us/az/statewide data URL

MiniCodeMonkey:
Existing URL was 404'ing


ingalls:
@MiniCodeMonkey Thanks a ton for the PR! I had to pull your commit into a new PR as you just discovered for us that the new batch CI job only runs for branches in this repo. Will look into making sure it runs for everyone

PR #5213, opened 2020-08-27
Cook, OR

ingalls:


iandees:
Looks like this is for Crook county, not Coos.

ingalls:
You're absolutely correct, thanks for the catch

PR #5222, opened 2020-08-27
Update City of Guelph

thismakessand:


ingalls:
Thanks a ton for your contribution! Looks Great! 

https://batch.openaddresses.io/run/119/

PR #5240, opened 2020-09-01
Update city_of_brampton.json

JosephKuchar:
Update link and license info from Open Data portal source


sources/ca/on/city_of_brampton.json
```diff
@@ -2,7 +2,7 @@
     "coverage": {
         "country": "ca",
         "state": "on",
-        "city": "Burlington",
```
ingalls, line 5: Oooh you got me, guilty of copy and paste

Issue #5255, opened 2020-09-03
Göteborg, Sweden


PierreMesure:
Hi,

The municipality of Göteborg recently published a lot of geodata, including its addresses as CC0. The format they use if Shapefile (.shp).

I can try to have a go at it but I'd love feedback!

ingalls:
Merged here: https://github.com/openaddresses/openaddresses/pull/5256

PR #5256, opened 2020-09-03
Göteborg, Sweden

PierreMesure:
Related issue: #5255 

Feedback welcome, this is my first one and it is for now untested.

A few comments for the person reviewing:
- the archive also contains .prj, .dbf, .cpg and .shx. Should they be left unused?
- is the file path `adresser_shp_v1` needed since it is the top level folder? I'm also afraid the data gets updated and the version number changes, breaking the code.
- I chose to use the variable nr_littera för street number, it's a combination of numbers and letters (7A for example), could that be a problem?

### DIFF
--- a/sources/se/municipality_of_goteborg.json
+++ b/sources/se/municipality_of_goteborg.json
@@ -24,7 +24,7 @@
           "number": "nr_littera",
           "street": "gatunamn",
           "format": "shapefile",
-          "file": "adresser_shp_v1/adresser.shp"
+          "file": "adresser.shp"
         }
       }
     ]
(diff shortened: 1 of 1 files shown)


nvkelso:
Very exciting to see Swedish data opening up and landing in Open Addresses :)

ingalls:
This is fantastic! thanks a ton for the contribution

PierreMesure:
Thanks for merging so quickly! Do you know when the data will be reflected on the map at openaddresses.io?

ingalls:
@PierreMesure are you talking about the main map at: https://openaddresses.io/? 

If so, we just switched to our new processing system, which in the near future will allow us to expand the scope of the project to include building footprints & parcels. I have an open ticket here: https://github.com/openaddresses/batch/issues/35 for updating dotmap on the main page for new sources. Until that is done, only sources processed by the old system will appear.

Issue #4680, opened 2019-11-19
Wrong Counties in US West Extract


Jnb2387:
The Colorado folder downloaded from the US West extract has addresses from New York state. 
All of Washington County pulls Washington county NY. I am not aware of a service url that will pull Washington County CO.

Atomahawk:
Error persists in most recent parcel extract for us/co/washington_county

arch0345:
Fixed in #5263

PR #5263, opened 2020-09-09
File at us/co/washington should actually be us/ny/washington

iandees:
As reported via email.


arch0345:
Related issue:
* #4680

PR #5285, opened 2020-09-17
Update city_of_ajax.json

JosephKuchar:
Added license information.


ingalls:
@JosephKuchar There are 3 PRs for the city of AJAX, just wondering which is the canonical and which ones I should close

JosephKuchar:
> @JosephKuchar There are 3 PRs for the city of AJAX, just wondering which is the canonical and which ones I should close

This should be the canonical one, but I see the data didn't pull properly. I'll see if the data source also needs to be changed then update the same branch.

PR #5293, opened 2020-09-17
Update city_of_welland.json

KHobbs3:
Updated license information.


ingalls:
@KHobbs3 I think the attribution and sharealike need to be boolean: `true` instead of a string `"true"`

PR #5299, opened 2020-09-17
Update city_of_welland.json

KHobbs3:
Updated data link.


ingalls:
@KHobbs3 Looks like the `number`/`address` fields need to be updated

ingalls:
@KHobbs3 something is wonky about the coordinates in this one. They all seem to be the same point. Have you tested opening it in QGIS to make sure it is valid?

KHobbs3:
> @KHobbs3 something is wonky about the coordinates in this one. They all seem to be the same point. Have you tested opening it in QGIS to make sure it is valid?

The map looks okay for me in QGIS but the coordinates do look extremely similar in the table so I'll keep investigating to try to sort that out! Thanks

PR #5314, opened 2020-09-17
Update saskatoon.json

KHobbs3:


ingalls:
@KHobbs3 This one looks like it doesn't have any coordinates. Do we have a better source by chance?

KHobbs3:
> @KHobbs3 This one looks like it doesn't have any coordinates. Do we have a better source by chance?

Not that I've come across so far but I flagged it on our end as well. Thanks!

ingalls:
Going to close this PR. If another source is found, feel free to create a new PR for saskatoon

PR #5315, opened 2020-09-17
Update okanagan_similkameen.json

KHobbs3:
Updated license information.


iandees:
Instead of using the download links for arcgis hub sources, please use the MapServer or FeatureServer that arcgis downloads from. You can usually find those links in the "Data Source" link under "About".

PR #5346, opened 2020-09-21
Update gatineau.json

KHobbs3:


ingalls:
@KHobbs3 We don't support the `json` format. Looking at the data source it appears to be valid GeoJSON. Can you try updating the format tag to `geojson` and seeing if it pulls?

PR #5353, opened 2020-09-21
Update city_of_niagara_falls.json

KHobbs3:


ingalls:
@KHobbs3 It looks like the `ca/bc/north_cowichan` source is failing as the zip file contains multiple shapefiles. Can you update the conform object with a `file` attribute to specify which shapefile it should pull from?

PR #5354, opened 2020-09-21
Gatineau, QC

ingalls:
Closes: https://github.com/openaddresses/openaddresses/pull/5346

### DIFF
--- a/sources/ca/qc/gatineau.json
+++ b/sources/ca/qc/gatineau.json
@@ -23,10 +23,9 @@
                     "attribution name": "CC BY 4.0",
                     "attribution": true
                 },
-
                 "protocol": "http",
                 "conform": {
-                    "format": "geojson"
+                    "format": "geojson",
                     "number": "NO_CIVIQUE",
                     "street": [
                         "TYPE_VOIE",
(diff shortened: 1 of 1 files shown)


JosephKuchar:
Note for the order of addresses in conform,

"NOM_VOIE" : "Lacaune",
"LIEN_VOIE" : "de",
"TYPE_VOIE" : "Rue"

The street name here is "Rue de Lacaune", so the order should be TYPE, LIEN, NOM, ORIENTATION. 

Not sure why it's failing, the log from data-pls suggests it's a compression thing, but it looks like Nick has already removed the compression?

iandees:
The source file has a BOM ([byte order marker](https://www.freecodecamp.org/news/a-quick-tale-about-feff-the-invisible-character-cd25cd4630e7/) and it looks like our GeoJSON pipeline doesn't properly remove it. We'll probably need to add some code to the downloader to work around it.

ingalls:
@iandees just a note, I've broken out the downloader here: https://github.com/openaddresses/batch-machine for the purposes of batch. Whoever takes a look we can back/forward port it to machine/batch-machine

PR #5359, opened 2020-09-22
Update regina.json

KHobbs3:


ingalls:
Same here, any particular reason for the change from a MapServer to a shapefile?

PR #5360, opened 2020-09-22
Update city_of_st_john.json

KHobbs3:

### DIFF
--- a/sources/ca/nb/city_of_st_john.json
+++ b/sources/ca/nb/city_of_st_john.json
@@ -9,12 +9,19 @@
         "addresses": [
             {
                 "name": "city",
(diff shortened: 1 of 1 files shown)


ingalls:
@KHobbs3 Any reason for the change from a MapServer to a shapefile? Generally we very much prefer MapServers,

KHobbs3:
> @KHobbs3 Any reason for the change from a MapServer to a shapefile? Generally we very much prefer MapServers,

No problem here, I'll switch it back. For those whose data source links ("website") also change, the MapServer link is swapped out because the licenses are more clearly associated with the new data link. I'm going through previous changes now and trying to correct for this and leave the MapServer link where possible.

PR #5366, opened 2020-09-23
Update city_of_banff.json

KHobbs3:


ingalls:
Ooof, we don't support zip files within zip files. Is it possible to reach out to the source and get them to have non-zipped files within the main bundle zip? Double zipping compressed files is generally not really useful.

KHobbs3:
> Ooof, we don't support zip files within zip files. Is it possible to reach out to the source and get them to have non-zipped files within the main bundle zip? Double zipping compressed files is generally not really useful.

Sent them a request!

PR #5379, opened 2020-09-23
Update province.json

KHobbs3:
Updated data link and license information.

### DIFF
--- a/sources/ca/ns/province.json
+++ b/sources/ca/ns/province.json
@@ -13,13 +13,16 @@
         "addresses": [
             {
                 "name": "state",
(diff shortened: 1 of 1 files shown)


sources/ca/ns/province.json
```diff
@@ -13,13 +13,16 @@
         "addresses": [
             {
                 "name": "state",
-                "data": "https://nsgiwa.novascotia.ca/arcgis/rest/services/BASE/BASE_NS_CivicAddress_File_WM84/MapServer/0",
+                "data": "https://data.novascotia.ca/api/geospatial/tntn-er5g?method=export&format=Shapefile",
```
iandees, line 16: The old URL seems to exist still. Why did we change this URL? I'd prefer to use Esri endpoints as they're usually kept up to date.

iandees (changes requested):

PR #5382, opened 2020-09-23
Update city_of_kingston.json

KHobbs3:
Updated data links and added license information.

### DIFF
--- a/sources/ca/on/city_of_kingston.json
+++ b/sources/ca/on/city_of_kingston.json
@@ -18,22 +18,21 @@
                 "name": "city",
                 "protocol": "http",
                 "compression": "zip",
-                "data": "https://apps.cityofkingston.ca/data/gis/Civic_Address_SHP.zip",
-                "website": "https://www.cityofkingston.ca/explore/data-catalogue",
+                "data": "https://opendatakingston.cityofkingston.ca/explore/dataset/civic-addresses/download/?format=geojson&timezone=America/New_York&lang=en",
+                "website": "https://opendatakingston.cityofkingston.ca/explore/dataset/civic-addresses/information/?disjunctive.street_type_suffix",
+                "license": {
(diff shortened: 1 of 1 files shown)


sources/ca/on/city_of_kingston.json
```diff
@@ -18,24 +18,31 @@
                 "name": "city",
                 "protocol": "http",
                 "compression": "zip",
-                "data": "https://apps.cityofkingston.ca/data/gis/Civic_Address_SHP.zip",
-                "website": "https://www.cityofkingston.ca/explore/data-catalogue",
+                "data": "https://opendatakingston.cityofkingston.ca/explore/dataset/civic-addresses/download/?format=geojson&timezone=America/New_York&lang=en",
```
iandees, line 21: Looks like you're asking to download it as geojson but specify a format of "shapefile" below.

iandees (changes requested):

JosephKuchar:
Looks like the issue may be the column names. I'm looking at the geojson, these are the variables:

"address_number"
"street"
"postal_code"
"unit"

Probably there's an inconsistency between the csv column names and geojson column names.

JosephKuchar:
Looking at the log from data-pls, it looks like it is having issues with the geojson:

2020-09-29 13:16:54,820   DEBUG: Content-Type says "application/json" for https://opendatakingston.cityofkingston.ca/explore/dataset/civic-addresses/download/?format=geojson&timezone=America/New_York&lang=en
2020-09-29 13:16:54,820   DEBUG: Content-Disposition disagrees: "civic-addresses.geojson"
2020-09-29 13:16:54,823 WARNING: Error in process_one.process()

Maybe @iandees or @ingalls can tell what's going on. I downloaded the csv to see if it would be better, but it's also a bit weirdly formatted.

arch0345:
Superseded by #6943

PR #5387, opened 2020-09-23
Update city_of_ottawa.json

KHobbs3:

### DIFF
--- a/sources/ca/on/city_of_ottawa.json
+++ b/sources/ca/on/city_of_ottawa.json
@@ -28,9 +28,9 @@
                 "protocol": "ESRI",
                 "conform": {
                     "format": "geojson",
-                    "unit": "Address Unit Number",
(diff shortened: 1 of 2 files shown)


ingalls:
@KHobbs3 MapServers don't support spaces in the field names, they are "pretty printed" on the arcgis site.

to get the actual underlying field names - go to the mapserver url: https://maps.ottawa.ca/arcgis/rest/services/Municipal_Address/MapServer/0

And they will be present at the bottom of the page. One usually needs an example to help guide which field is which so append `/query` to the end of the mapserver URL:

https://maps.ottawa.ca/arcgis/rest/services/Municipal_Address/MapServer/0/query

On the query page, fill out the following:
```
Query: 1=1
Out Fields: *
```

Then click Query (GET) and it should provide you with a list of examples

PR #5402, opened 2020-09-24
Create cape_breton.json

KHobbs3:
Created source file for Cape Breton, NS.


JosephKuchar:
I took a look at the logs and data-pls failed on a 'lat' key error. It looks like coordinates aren't specified in the json, the coordinate columns are "X Map Coordinate" and "Y Map Coordinate".

PR #5403, opened 2020-09-24
Update ch/countrywide.json --> Correct the x and y

gismountains:


gismountains:
Incorreect -->
image

andrewharvey:
hi I'm not sure what's going on here, when I open the CSV in QGIS with EPSG:2056 and use the original/current fields

"lon": "GKODE",
"lat": "GKODN",

It correctly positions in CH. Don't know if there is an issue somewhere else in the OA processing code either because the data bot preview doesn't look right with your change either.

:shrug:

iandees:
I think this is a bug in batch: https://github.com/openaddresses/batch/issues/90

iandees:
I'm going to close this as the lat/lon columns seem to be picked correctly in the source.

PR #5405, opened 2020-09-26
upgrade au/nsw to 26 Sept 2020 extract

andrewharvey:
Swap to new source at https://portal.spatial.nsw.gov.au/portal/apps/sites/#/home/items/9b2990428ef4411da7c61ae56b2c6806


andrewharvey:
any reason why CircleCI needs to be authorised before I can see the logs?

andrewharvey:
The GeoJSON is in GeoJSONSeq so just one feature per line, not wrapped in a FeatureCollection, maybe this is the issue...

PR #5417, opened 2020-09-28
Update city_of_guelph.json

KHobbs3:
updated license information.


JosephKuchar:
Looks like the issue with CircleCI is that the quotations around "share-alike" aren't consistent in your edit. I imagine if that's fixed it will work!

PR #5418, opened 2020-09-28
Update city_of_hamilton.json

KHobbs3:
Updated license info and data links (ESRI link has error code 403).


JosephKuchar:
Looks like the issue is the protocol should be http, but is still ESRI.

PR #5434, opened 2020-09-28
Update city_of_oakville.json

KHobbs3:
Updated data link (ESRI link has Error 403).

### DIFF
--- a/sources/ca/on/city_of_oakville.json
+++ b/sources/ca/on/city_of_oakville.json
@@ -13,7 +13,7 @@
         "addresses": [
             {
                 "name": "city",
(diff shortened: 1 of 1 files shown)


KHobbs3 (commented):
https://maps.oakville.ca/oakgis/rest/services/ArcGISOnline/AGO_Oakville_Misc_Layers/MapServer/15 gives me Error 403 (Services Directory has been disabled). Is this the same for you?

iandees:
This is a common way to prevent easy browsing. The downloader will use JSON output format, which isn't blocked:

https://maps.oakville.ca/oakgis/rest/services/ArcGISOnline/AGO_Oakville_Misc_Layers/MapServer/15?f=json

I was trying to trigger a change to show that it still works.

PR #5459, opened 2020-11-20
Update port_moody.json

JosephKuchar:
Fix variable names


arch0345:
This has since been fixed in #6349

PR #5468, opened 2020-12-23
Switzerland: ch/contrywide fix missing postcodes and city names

SilvrDuck:
The latest [ch/countrywide](http://results.openaddresses.io/sources/ch/countrywide) data was having blank `POSTCODE` and `CITY` columns.

There was a mismatch with the column names of the [source data](https://data.geo.admin.ch/ch.bfs.gebaeude_wohnungs_register/CSV/CH/CH.zip), so this PR fixes it.

### DIFF
--- a/sources/ch/countrywide.json
+++ b/sources/ch/countrywide.json
@@ -37,8 +37,8 @@
                     "street": "STRNAME",
                     "lon": "GKODE",
                     "lat": "GKODN",
-                    "city": "PLZNAME",
-                    "postcode": "PLZ4",
+                    "city": "DPLZNAME",
+                    "postcode": "DPLZ4",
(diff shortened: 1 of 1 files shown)


iandees:
Looks like this is being affected by the lat/lon bug in https://github.com/openaddresses/batch/issues/90

zifeo:
@iandees anything blocking the merge? We are waiting on that to re-enable openaddresses.

iandees:
There's a bug that's preventing our system from processing the source properly. I can merge it, but the data output will be affected by the bug mentioned above.

zifeo:
@iandees Thanks. This is why we get the image rotated and flipped?

iandees:
Yep.

PR #5478, opened 2021-02-16
Adding São Paulo city Data

feromes:
Hi,

We have processed data from parcels in São Paulo city and now we are able to share with others.

Thanks a lot


feromes:
Just a little mistake on Jsonfile solved.

PR #5483, opened 2021-02-17
DC Parcels

ingalls:
### Context

DC Parcel Source

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2014
+Copyright (c) 2021 OpenAddresses
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without modification,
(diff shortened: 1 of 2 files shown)


ingalls:
Screenshot from 2021-02-17 16-04-33

Very exciting - processed data from the live service. Unfortunately this uncovered a couple bugs 

- [ ] Street/Number are being output as the properties? Didn't see this in local testing - should just be PID
- [ ] No Preview Image, Expected but should update the image creater to support polys
- [ ] Use GitDiff to attempt to only run modified layer in diff? To prevent 2 sources from being run when only 1 was added?

iandees:
> No Preview Image, Expected but should update the image creater to support polys

Preview image is also not working for points: https://github.com/openaddresses/openaddresses/pull/5481

ingalls:
Working on this here: https://github.com/openaddresses/batch/pull/113

iandees:
🎉 🎉 🎉

iandees:
Great work, Nick!

PR #5485, opened 2021-02-20
add au/act parcels

andrewharvey:
awesome work on getting parcels accepted @ingalls, I'll add what I can for AU/NZ.

### DIFF
--- a/sources/au/act/statewide.json
+++ b/sources/au/act/statewide.json
@@ -39,6 +39,7 @@
         ],
         "parcels": [
             {
+                "name": "state",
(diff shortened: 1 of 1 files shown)


nvkelso:
Are the preview maps showing parcel centroids instead of polygons? Is the plan to show the polygons?

andrewharvey:
> Are the preview maps showing parcel centroids instead of polygons? Is the plan to show the polygons?

Yeah looks like it, can see it better on #5486. It threw me, I downloaded the data to check indeed polygons were coming through. Would be better if actual polygons were shown, or if only centroids then I'd suggest to add a text note otherwise it will cause confusion.

PR #5488, opened 2021-02-20
add au/vic/geelong parcels

andrewharvey:


andrewharvey:
Not entirely sure why this failed, since it's actually the address which failed, gives a warning
```WARNING: Unknown source conform format shapefile-polygon``` which is strange because this is documented, and hasn't changed from what was already here...

PR #5496, opened 2021-02-20
add au/nsw parcels

andrewharvey:


andrewharvey:
Strange I didn't change addresses here at all but now it's failed:

```
    match = re.sub(pattern, convert_regexp_replace(replace), row[fxn["field"]])
KeyError: 'street_wip'
```

iandees:
Is tianjara.net your domain?

iandees:
I downloaded the address data and it doesn't seem to have the `street_wip` column. Maybe the data we're downloading changed?

andrewharvey:
> Is tianjara.net your domain?

Yes. I used to only have 100KB/s upload, so would do the processing on a remote server, and then just leave the file there so that was easier than using the HTTP web form to upload to the cache. But now I can probably upload these to the cache.

> I downloaded the address data and it doesn't seem to have the street_wip column. Maybe the data we're downloading changed?

street_wip isn't from a column, I followed the example for chain functions at https://github.com/openaddresses/openaddresses/blob/master/ATTRIBUTE_FUNCTIONS.md#chain where street_wip is just the variable name.

The data hasn't changed since the last successful run I think.

andrewharvey:
closing in favour of #5828,  I'll open a new issue about the issue with chain functions.

PR #5503, opened 2021-02-22
add au buildings from Microsoft

andrewharvey:
preemptive PR for buildings

### DIFF
--- a/sources/au/countrywide.json
+++ b/sources/au/countrywide.json
@@ -37,6 +37,24 @@
                     "region": "region"
                 }
             }
+        ],
+        "buildings": [
+            {
+                "name": "country",
+                "protocol": "http",
+                "data": "https://usbuildingdata.blob.core.windows.net/australia-buildings/Australia_2020-06-21.geojson.zip",
+                "website": "https://github.com/microsoft/AustraliaBuildingFootprints",
+                "compression": "zip",
+                "license": {
+                    "url": "https://opendatacommons.org/licenses/odbl/",
+                    "attribution": true,
+                    "attribution name": "Microsoft",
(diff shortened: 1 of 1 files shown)


andrewharvey:
@iandees it looks like all the cache files were wiped

> openaddr.cache.DownloadError: 403 response from https://data.openaddresses.io/cache/uploads/andrewharvey-openaddr/ec6370/au-feb2021.zip

This file I just uploaded a week or so ago and it worked before. When I try to download the latest cache file from http://results.openaddresses.io/sources/au/countrywide to restore it, it also gives a 403.

Is this a problem on the OA infrastructure end?

iandees:
Yea, looks like it. @ingalls I think we need to add a bit of code to batch that fetches data with URL `https://data.openaddresses.io` from `s3://data.openaddresses.io` instead, so we can skip the CDN costs. I'll file a ticket on that.

iandees:
The data wasn't wiped, I just added a referrer check and machine-batch isn't doing the S3 hack that I thought it was doing. We can fix these! See https://github.com/openaddresses/batch/issues/134

ingalls:
I've bumped this as I've added S3 support to `batch`. @andrewharvey Do you use slack? Would love to get you on the OA slack if you do.

ingalls:
```
NoSuchBucket: The specified bucket does not exist
```

andrewharvey:
@ingalls yes please add me `andrew@alantgeo.com.au`

ingalls:
Sent you an invite - @iandees just has to accept it

ingalls:
rerunning with bug fix here: https://batch.openaddresses.io/run/885

PR #5541, opened 2021-03-11
add nz buildings

andrewharvey:


andrewharvey:
Not sure what the problem is here I can download 2.76G from the data url.

andrewharvey:
Log reports a "MemoryError".

Issue #5694, opened 2021-04-13
Issue with us/fl/_loveland data


collinjc:
We're using openaddresses data in conjunction with Pelias and have noticed a few issues with the data coming back from the us/fl/_loveland dataset. First of all, it seems that this data is missing ZIP codes. More frustrating, however, is that there appears to be a great deal of bad data where the house number and street have the same value duplicated. For example:
{"type":"Feature","properties":{"hash":"22f4d81746a5a6f6","number":"325","street":"325","unit":"","city":"mayo","district":"lafayette","region":"fl","postcode":"","id":""},"geometry":{"type":"Point","coordinates":[-82.9529545,29.9342789]}}
From what I can tell, there is no 325 325 Mayo, FL.

collinjc:
This may need to be a separate issue, but I'm unable to see the results of this change, seemingly due to a failure in processing the _loveland data. [https://results.openaddresses.io/sources/us/fl/_loveland](https://results.openaddresses.io/sources/us/fl/_loveland) is showing that the job has been failing since 2021-03-23 due to the source file being unavailable. This is presumably causing the "old" data to still be returned.

arch0345:
Loveland source was removed in #6554

arch0345:
Resolved in #6554

PR #5727, opened 2021-04-14
Improve Florida Loveland dataset parsing for invalid addresses

orangejulius:
As mentioned in https://github.com/openaddresses/openaddresses/issues/5694 there is some invalid data in the Loveland Florida dataset.

From my examination, about 35,000 of the 9 million addresses are coming through with a housenumber and streetname that are identical.

Looking at the source data, this is coming from records where the address field contains only a number, something like `"325"`.

The regular expression we are currently using to parse the data will cause both the housenumber _and_ street to use that single number in this case.

This PR attempts to improve the regex so that the street column is empty instead.

### DIFF
--- a/sources/us/fl/_loveland.json
+++ b/sources/us/fl/_loveland.json
@@ -22,7 +22,7 @@
                     "street": {
                         "function": "regexp",
                         "field": "phy_addr1",
-                        "pattern": "^(?:[0-9]+ )(.*)",
+                        "pattern": "^(?:[0-9]+)\\w+(.*)",
                         "replace": "$1"
                     },
                     "city": "city",


orangejulius:
Wow the new job stats are awesome!

And I think it means this PR does what I intended:
image

collinjc:
If the issue stems from the source data having a house number but no street, is there value in keeping that data around at all? Could it not be updated to simply ignore such records?

orangejulius:
@collinjc the OA team is discussing removing this dataset all together, so that might happen too.

But for Pelias in particular, we won't import records that don't have both a housenumber and a streetname, so this will solve your issues there while keeping the majority of the data, which is likely mostly valid, around.

collinjc:
@orangejulius Great, thanks for the update and info!

PR #5753, opened 2021-04-22
Update source url with stable link

NairolfL:
Hello,
I'd like to submit this change to update data in Luxembourg.
Same source as before, but up to date link.
Actually this new link is a redirect (302) to the latest up to date version of that file, so it shouldn't need any more change afterwards.

There is ~7k more addresses in that new file

### DIFF
--- a/sources/lu/countrywide.json
+++ b/sources/lu/countrywide.json
@@ -11,7 +11,7 @@
         "addresses": [
             {
                 "name": "country",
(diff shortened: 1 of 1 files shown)


NairolfL:
Well if seems that the job fails, is it because it doesn't follow the redirect ? 
Otherwise, the current up to date link without redirect is : https://download.data.public.lu/resources/adresses-georeferencees-bd-adresses/20210419-050701/addresses.geojson
But that's going to change

jeffdefacto:
This is now fixed from https://github.com/openaddresses/openaddresses/pull/7776

PR #5754, opened 2021-04-28
Update CONTRIBUTING.md to remove shapefile-polygon reference

andrewharvey:
It appears that this format value is deprecated based on the list of supported formats.


ingalls:
Thanks @andrewharvey. I still process these files correctly on the backend but I'm slowly removing sources that have it and replacing them with `shapefile` as we can easily determine this algorithmically.

Issue #4265, opened 2018-11-21
US Virgin Islands Data


justinelliotmeyers:
https://www.dropbox.com/sh/166qphozl216sr4/AACJiRprAvaqOGOPPvd2_Zwta?dl=0

Randal Hale <rjhale@northrivergeographic.com>

...and then after we talked I started looking for the addresses going 
"Oh noooooo where are they" - BUT - I found them!

The legal correct address  is the "address" field. We kept the old 
address (Legal_desc) to tie it back to the parcel. I'm just looking and 
realizing they (USVI) took the zip code out because they think zip codes 
on the islands are a waste of memory cells.

If this cranks back up again I'll send you the rest of the addresses.

Randy

On 11/20/18 4:07 PM, J M wrote:
> Hey, it's Justin. Thanks for the call today, I appreciate all the 
> insight into the Virgin Island addresses! Does sound like a pretty 
> wild stories :-) I was wondering if there's any way you could send me 
> the official shapefile data to use in the open addresses project.
>
> I'll create a official issue and the repository for it, I'll also pink 
> Ian and let him know you said hi! Have a good Thanksgiving, thanks!!
>
> Justin

ingalls:
This data is fantastic! I reached out to the current GIS admin for the USVI for an updated copy of the data. They were very responsive & helpful. As soon as I get a shapefile back I'll get this loaded into the system

justinelliotmeyers:
@ingalls gracias amigo!

PR #5781, opened 2021-06-02
US Virgin Islands

ingalls:
Closes: https://github.com/openaddresses/openaddresses/issues/4265


ingalls:
:bug: Currently missing street names

PR #5792, opened 2021-06-21
update au/countrywide for MAY21 release

andrewharvey:


andrewharvey:
It looks like the cache upload service is broken, hope my files weren't lost otherwise I'll need to re-process and upload again.

iandees:
Looking at the S3 bucket contents, I think maybe the URL was typoed from a 6 to a 5 in the hash string. I fixed that and kicked off a new CI build.

andrewharvey:
Oh thanks for checking! Normally I copy paste but this time I just replaced the hash manually.

iandees:
Looks good! Merging.

PR #5799, opened 2021-07-02
Adding website and license information to CO statewide

fishcharlie:
As shown in the About tab of [this link](https://data.colorado.gov/State/Statewide-Aggregate-Addresses-in-Colorado-2020-Pub/n7je-akky), the license for this data is `Public Domain`.

I have verified that if you click Export, and choose the CSV file type, the URL is identical to the existing `data` field in the source file. This verifies that the data source is identical to what we are pulling from.

You can also find screenshots of this attached below.

Screen Shot 2021-07-01 at 7 25 03 PM

Screen Shot 2021-07-01 at 7 25 14 PM

### DIFF
--- a/sources/us/co/statewide.json
+++ b/sources/us/co/statewide.json
@@ -13,6 +13,12 @@
             {
                 "name": "state",
                 "data": "https://data.colorado.gov/api/views/ge5d-pvi7/rows.csv?accessType=DOWNLOAD",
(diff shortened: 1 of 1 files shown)


fishcharlie:
Any reason why this would be failing the check?

ingalls:
@fishcharlie Looks like it is unrelated to your change. Thanks a ton for the PR!

PR #5800, opened 2021-07-02
Denver, CO - Buildings, Parcels, and other updates

fishcharlie:
This PR makes the following changes to Denver, CO:

1. Adds building data
2. Adds parcel data
3. Adds license and website information to existing addresses entry

My only concern about this PR is I can't find a way to test the data. So although I think I formatted everything correctly and everything, I have been unable to test it to verify the output is correct.

Is there anyway to test everything before submitting a PR?

### DIFF
--- a/sources/us/co/denver.json
+++ b/sources/us/co/denver.json
@@ -36,6 +36,20 @@
                 "compression": "zip"
             }
         ],
+        "parcels": [
+            {
+                "name": "county",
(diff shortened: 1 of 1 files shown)


fishcharlie:
I think I fixed this. Therefore I think it is ready to be reviewed and merged (if approved). Please let me know if there are any issues I need to correct.

ingalls:
This looks fantastic! Thanks a ton for your last few contributions.

Issue #5802, opened 2021-07-06
Fix Switzerland Source Projection


flatwhite0928:
Hi 👋 , the projection of ch/countrywide (Switzerland) addresses looks incorrect. Could you please fix it?
 
- Job #78735 ch/countrywide - addresses - country: https://batch.openaddresses.io/job/78735
![image](https://user-images.githubusercontent.com/31902919/124652493-31e62c80-de6a-11eb-8ee4-fa1d671d3486.png)

PR #5813, opened 2021-07-08
CH Countrywide

ingalls:
### Context

Fixes CH Countrywide projection issues and updates to the latest data

Closes: https://github.com/openaddresses/openaddresses/issues/5802

### DIFF
--- a/sources/ch/countrywide.json
+++ b/sources/ch/countrywide.json
@@ -1,4 +1,5 @@
 {
+    "schema": 2,
     "coverage": {
         "ISO 3166": {
             "alpha2": "CH",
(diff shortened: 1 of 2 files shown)


ingalls:
```
/usr/local/src/batch/lib/validator.js:106
        if (isNaN(Number(feat.geometry.coordinates[0])) || feat.geometry.coordinates[0] < -180 || feat.geometry.coordinates[0] > 180) errs.geometry.push('Feat exceeds +/-180deg coord bounds');
                                       ^
TypeError: Cannot read property 'coordinates' of null
    at Validator.addresses (/usr/local/src/batch/lib/validator.js:106:40)
```

PR #5821, opened 2021-07-09
Norway

ingalls:


arch0345:
Was this resolved in https://github.com/openaddresses/openaddresses/pull/5813/commits/65cb818e6370b2285515ad66f87cc7ffe40df42c? Not sure which one's newer, and it looks like they have different attributes.

PR #5836, opened 2021-07-12
update au/tas/launceston with new source

andrewharvey:


andrewharvey:
looks like conform.py issues based on the log https://batch.openaddresses.io/job/129097/log

PR #5840, opened 2021-07-16
Adding building unit attributes to us/co/denver

### LINKED ISSUE #5838: us/co/denver Duplicate Entries
In `us/co/denver` there are duplicate entries for `4303 N Brighton Blvd Rear`. OpenAddresses Hash: `a4f552c5a3f25ca0`.

In the raw data it looks like these should be two different entries:

```
4303 N Brighton Blvd Bldg 1 Rear
4303 N Brighton Blvd Bldg 2 Rear
```

I'm not sure how to fix this. I think adding those properties to the `conform` property will add a bunch of bad data to the output.

Any ideas how to proceed here?

fishcharlie:
Adding building unit attributes to us/co/denver.

Someone with more experience here should probably verify this worked correctly.

Closes #5838

### DIFF
--- a/sources/us/co/denver.json
+++ b/sources/us/co/denver.json
@@ -27,6 +27,8 @@
                         "POSTDIRECT"
                     ],
                     "unit": [
+                        "BUILDING_T",
+                        "BUILDING_I",
                         "COMPOSITE_",
                         "COMPOSIT_U"
                     ]
--- a/sources/us/co/denver.json
+++ b/sources/us/co/denver.json
@@ -29,8 +29,8 @@
                     "unit": [
                         "BUILDING_T",
                         "BUILDING_I",
-                        "COMPOSITE_",
-                        "COMPOSIT_U"
+                        "UNIT_TYPE",
+                        "UNIT_IDENT"
                     ]
                 },
                 "website": "https://www.denvergov.org/opendata/dataset/city-and-county-of-denver-addresses",


fishcharlie:
It looks like this failed. As you can see, the `unit` field is `Bldg 1 Bldg 1`. I don't think it should have duplicates like that.

```
{"type":"Feature","properties":{"hash":"f36fca2341981d4d","number":"4303","street":"N Brighton Blvd","unit":"Bldg 1 Bldg 1","city":"","district":"","region":"","postcode":"","id":""},"geometry":{"type":"Point","coordinates":[-104.9715914,39.7779286]}}
```

iandees:
It looks like what we had before was right:

image

But you found a case where their composite field wasn't populated properly:

image

So I'm going to change the fields it uses for unit to include the two building fields and the two unit fields.

iandees:
Thanks for spotting this. I'm going to merge in this change and we can make further changes with other PRs if needed.

fishcharlie:
@iandees Sounds good. In order to test this on my end, I'm relying on the `https://batch.openaddresses.io/api/data` API endpoint to get the S3 path.

How often does that get updated?

Is there a better method to get this data earlier? It looks like the `https://batch.openaddresses.io/api/job` API endpoint has the job listed from this PR, but that isn't guaranteed to be stable or reviewed.

iandees:
@ingalls can you give some more info on @fishcharlie's questions here?

PR #5897, opened 2021-09-06
Added DeWitt County, IL (blocker: need libpostal parsing)

MiniCodeMonkey:


MiniCodeMonkey:
@iandees The SITUS has city/zip postfixed, I can't recall a situation in the past where that has been the case. Any thoughts on how we can extract the street name from this within OA?

iandees:
Hmm. I don't know if we have a good solution for this. I think we were talking about bringing in support for libpostal to parse strings like this into the parts but never did.

PR #5928, opened 2021-10-25
Remove duplicate source

MiniCodeMonkey:
This is the same source as https://github.com/openaddresses/openaddresses/blob/master/sources/us/wa/snohomish_county.json


MiniCodeMonkey:
@ingalls Going to merge this in

PR #5983, opened 2022-01-02
Replaced individual us/tx county sources using 2019 data with statewide sources using 2021 data

MiniCodeMonkey:
[tnris](https://data.tnris.org/collection?c=94502179-9389-4bfa-b753-5e43f6d477bf) has now released updated address data for Texas.

This replaces all individual sources that reference the now legacy 2019 data with a single statewide datasource.

Screen Shot 2022-01-02 at 1 16 38 PM


iandees:
Thanks for doing this. We had separated them out into one file per county because it didn't cover the whole state. This seems like a much nicer way to update the data, though.

MiniCodeMonkey:
@iandees Yeah, it looks like the statewide version is new for the 2021 data. Very convenient improvement!

Issue #5984, opened 2022-01-05
US source street and number drop questions 


ilissablech:
👋  @ingalls. A few OA source drops I noticed. Appreciate your eyes, but no rush. 🙇 

## us/ak/matanuska_susitna_borough
This was updated in https://github.com/openaddresses/openaddresses/pull/5867, but do they just have no number or street data to provide?
![image](https://user-images.githubusercontent.com/25749483/148261853-40462298-08c0-4118-ae5e-d9b1f03e19b8.png)

## us/mn/st_louis
number and street dropped count nov19-nov26 2021
![image](https://user-images.githubusercontent.com/25749483/148261971-b1dffc0f-1364-47ce-a9df-e5abd9db5565.png)

## us/pa/crawford
number and street dropped count oct 29-nov5 2021
![image](https://user-images.githubusercontent.com/25749483/148263171-e99a19ee-1d91-4e00-a396-172b7a3ba372.png)

ingalls:
- [ ] us/ak/matanuska_susitna_borough
- [ ] us/mn/st_louis
- [ ] us/pa/crawford
- [ ] us/il/adams
- [ ] us/ky/pendleton
- [ ] us/nd/bottineau
- [ ] us/nd/cass
- [x] us/nd/city_of_fargo

PR #5985, opened 2022-01-10
Fargo, ND

ingalls:
Ref: https://github.com/openaddresses/openaddresses/issues/5984


ingalls:
Address Layer Pass: https://batch.openaddresses.io/job/205158

PR #6008, opened 2022-02-14
Fixed ES script

StephanGeorg:
Fixed script to generate ES cached data. 

- URLs are redirected to HTTPS now. Fails if curl `-L` option is missing to follow redirects
- `build` directory is not added so script failed 
-  Syntax error in `scripts/es/download_gml.py`


ingalls:
Thanks a ton @StephanGeorg

PR #6027, opened 2022-03-02
Remove us/nm/sandoval, because it doesn't contain data for Sandoval county

iandees:
It actually contains some data for San Diego county, CA, which we have a better source for.


iandees:
This is based on a report emailed to us.

PR #6035, opened 2022-04-01
New & updated data in ca/on/Peel

kevin-peel:
Two changes were made for this request.

**ca/on/peel-region.json**
We've added a landuse column into our data and I've added the relevant tag within the JSON to allow OpenAddresses to parse it.

**ca/on/city_of_mississauga.json**
The City has released building footprints and parcels to their open data over the past couple of months.  I have added both to the JSON.  They have not released addresses; however, the peel-region addresses cover the city.

### DIFF
--- a/sources/ca/on/city_of_mississauga.json
+++ b/sources/ca/on/city_of_mississauga.json
@@ -1,0 +1,50 @@
+{
+    "coverage": {
+        "country": "ca",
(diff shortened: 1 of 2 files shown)


ingalls:
This is absolutely fantastic! As always, thanks a million for your work on the inside!

PR #6042, opened 2022-04-21
Update us/ca/san_joaquin with new source

MiniCodeMonkey:
Previous source was a one-off export from 2014


MiniCodeMonkey:
@ingalls Looks like the batch service is taking a small break today. Any thoughts on what might be going on?

MiniCodeMonkey:
Looks like a batch service error. A couple of recent jobs are running into the same error https://batch.openaddresses.io/job

MiniCodeMonkey:
This PR needs to be merged in for this job to process correctly: https://github.com/openaddresses/pyesridump/pull/85

MiniCodeMonkey:
@ingalls Thanks! Looks like the `pyesridump` dependency still needs to be updated on the `batch` side of things.

PR #6061, opened 2022-05-18
Update us/ca/san_mateo.json

mcolyer:
I saw that San Mateo county is currently empty on the map. I'm not totally sure what the issue is but I'm guessing that this is at least the first step towards fixing this.

The fields inside of the shapefile might also have changed but I'm unsure how to verify it.

I see these related but unmerged PR: https://github.com/openaddresses/openaddresses/pull/4270


mcolyer:
So I think I sorted out the field name and I tried changing the format to `shapefile` as the docs seem to suggest that's the new name for that format. 

I'm not sure quite else what to try, there is a geojson field available too but not sure if that would help or not https://www.smcgov.org/isd/gis-data-download.

mcolyer:
So it looks like the data is good now but I just realized that the zip contains two shapefiles. I think they are truncating the first at 100,000 entries and the second file contains the rest. What's the best way to support that?

ingalls:
Thanks a ton for putting this together! I've found the ESRI endpoint that appears to be where the address points are pulled from and compiled your PR here: https://github.com/openaddresses/openaddresses/pull/6062

PR #6071, opened 2022-06-19
it: Update Lombardia field mapping

HandyMenny:
The shapefile fields have changed.
Unfortunately not all cities have the street field populated. I don't think this file will be updated anymore, as I can no longer find any public reference to these files


iandees:
Thanks for improving the mappings!

PR #6072, opened 2022-06-20
Update South Tyrol source

HandyMenny:
- Removed  South Tyrol Ladin given that names in this language are no longer available
- Updated the url with the new WFS endpoint
- Removed district field. In other sources this field is used for province, while in this case it was used for "Fractions" i.e. a part of a city
- Switched to geojson format because csv provides coordinates in "Point(X Y)" format (regxp is not supported for lat/lon) and shapefile export is too slow (> 2 minutes)
- Added server-side reprojection because openaddress doesn't support reprojection with json format, there is probably a loss of accuracy, but I don't see many alternatives


ingalls:
:tada:

PR #6073, opened 2022-06-20
Update Verona Source

HandyMenny:
Changes:
- Updated file link (update 13/07/2018)
- Added "exponent" to street number (example 1/A, A is the exponent)

Some questions:

- This file isn't updated regularly and every time it updates its url changes. Is there any option in the json to prevent it from being reprocessed every time? (It's kind of a waste of resources)
- Is it useful to have city, district and region hardcoded in conform? Or is it just ignored?
- It looks like that the "validity check" considers street numbers with "exponent" invalid, why?

### DIFF
--- a/sources/it/34/city_of_verona.json
+++ b/sources/it/34/city_of_verona.json
@@ -16,7 +16,7 @@
         "addresses": [
             {
                 "name": "city",
-                "data": "http://file.comune.verona.it/opendata/territorio/2014/SHP_NUMERAZIONE_CIVICA_point%2026-02-21014.zip",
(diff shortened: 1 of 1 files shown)


ingalls:
@HandyMenny This looks fantastic! Thanks a ton for contributing.

- I have a plan to eventually respect the `frequency` tag but development is an unpaid side hobby so just have to balance my time
- We generally consider the `A` to be a unit number, is this not the case for verona?
-

HandyMenny:
> * We generally consider the `A` to be a unit number, is this not the case for verona?

yes it's used for that purpose, but in Italy we consider it part of the house number by writing it as 3/A or 3A.
If you want to put them in two separate fields we can do that, but we have to fix almost all Italian sources :)

PR #6077, opened 2022-08-03
Add Validate Schema

ingalls:
### Context

Up until know we have performed minimal data transformation in order to maintain the dataset as a source of authoritative data. We're in a good position to start performing optional additional transformations to clean up data to be more consistent and usable by end users.

This PR modified the schema to include a new `transform` object to manage per-source transforms on address layers


iandees:
Closing as stale (opened 2022-08, never revisited, no CI). Please reopen if there's renewed interest in a schema validator.

PR #6086, opened 2022-09-01
Add address source for ca/ns/kings

MiniCodeMonkey:


bili29:
Modisi, North Sulawesi

PR #6089, opened 2022-09-24
Add id to us/ca/butte

mcolyer:
Adds the APN as the id of the address.


iandees:
Is the APN unique per address or per parcel?

mcolyer:
I think it’s unique per parcel.

iandees:
Yay previews are back!

PR #6114, opened 2022-12-30
Update URL for Marion, FL

iandees:


iandees:
The Marion county server has a crappy SSL certificate that the urllib3 library is refusing to connect to.

iandees:
(See https://stackoverflow.com/questions/38015537/python-requests-exceptions-sslerror-dh-key-too-small)

PR #6133, opened 2023-01-20
Fix Wallowa County, Oregon

arch0345:
It looks like there was a copy-paste error when creating the file for Wallowa County.


iandees:
Good catch. Thanks for fixing that.

PR #6217, opened 2023-01-24
Fixed mappings de/nw/city_of_cologne

StephanGeorg:


StephanGeorg:
Hey @iandees. Thanks for merging. 

One question: https://batch.openaddresses.io/data/218/history shows that the change was already processed but on https://batch.openaddresses.io/data it still shows the old job and data. Is there any step needed to enable the latest processed data? Thanks for clarification.

iandees:
Yea, something seems to be wrong with the batch summary/display map. I'm not sure how it works, so we'd need to lean on @ingalls to debug.

StephanGeorg:
It's not only that. It is also that `https://batch.openaddresses.io/api/data` returns `"job": 316054` as the latest jobId even if https://batch.openaddresses.io/data/218/history shows a newer one.

PR #6240, opened 2023-02-01
Try flipping lat/lon for Spain sources

### LINKED ISSUE #6222: spain: flipped lat/lon coordinates
I was validating my data and saw a big difference in Spain, 14M of addresses was no longer located in Spain.

So I imported all data from Spain into QGIS and here is the result:
image

In purple the polygon of Spain flipped thanks to [ST_FlipCoordinates](https://postgis.net/docs/ST_FlipCoordinates.html). 

As you can see, affected data are: `es/25829`, `es/25830`, `es/25831` and `es/32628`. You can easily do the check with a random line and [geojson.io](https://geojson.io/)

iandees:
Fixes #6222

### DIFF
--- a/sources/es/25829.json
+++ b/sources/es/25829.json
@@ -23,8 +23,8 @@
                 "protocol": "http",
                 "compression": "zip",
                 "conform": {
-                    "lon": "lat",
-                    "lat": "lon",
+                    "lon": "lon",
+                    "lat": "lat",
                     "number": "number",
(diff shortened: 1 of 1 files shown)


iandees:
Closing — this PR net-diffs to zero. It has two commits: the original lat/lon flip, followed by my own follow-up commit undoing it ("I think it was a bug in our extract software"). Nothing left to merge.

PR #6242, opened 2023-02-02
Update us-pa-allegheny address source

### LINKED ISSUE #6241: Allegheny County, PA missing addresses + attributes
It looks like a lot of addresses in Allegheny County, PA are missing (in the map below, they should be about as dense as Washington County to the south):

Screenshot 2023-02-02 at 11 54 08 AM

From the [history](https://batch.openaddresses.io/data/1852/history) it looks like addresses dropped from 626k to 39k in June 2022. They also lost all their attribute data.

This looks like a problem in the source data. You can browse it on the County's [open data portal](https://openac-alcogis.opendata.arcgis.com/datasets/626854df63a84eeaa2b3fca9b0a3b911/explore?layer=0&location=40.322389%2C-80.132279%2C15.21) and see the same missing addresses.

rbrtmrtn:
Closes #6241

Also adds contact and website for us-pa-allegheny addresses and parcels

### DIFF
--- a/sources/us/pa/allegheny.json
+++ b/sources/us/pa/allegheny.json
@@ -14,7 +14,7 @@
         "addresses": [
             {
                 "name": "county",
-                "data": "https://gisdata.alleghenycounty.us/arcgis/rest/services/Addressing/Addressing_AddressPoints/MapServer/0",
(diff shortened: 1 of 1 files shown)


rbrtmrtn:
Force-pushed to remove contact for buildings (this was throwing an error)

PR #6276, opened 2023-02-16
Update Los Angeles County, California

arch0345:


arch0345:
It looks like jobs for addresses/parcels passed, but the one for buildings keeps timing out
https://batch.openaddresses.io/run/1664
https://batch.openaddresses.io/run/1742

PR #6287, opened 2023-02-17
Update Coos County, Oregon

arch0345:


arch0345:
> It looks like these aren't pulling in attributes for addresses. Can you see if they updated attribute names on the source?

Looks like they're still using the same fields. Around a third of the parcels don't have an address:
image

iandees:
Ah I didn't see that they were sourced from parcels 👍🏻

PR #6358, opened 2023-02-25
Update Columbia County, Oregon

arch0345:
The shapefile from their website has a dbf with separated address fields.

Source zip: [Tax23.zip](https://www.columbiacountyor.gov/media/Assessor/GIS%20Data/Tax23/Tax23.zip)
Joined zip: [taxlotSitusJoined.zip](https://github.com/openaddresses/openaddresses/files/10830161/taxlotSitusJoined.zip)


iandees:
I'm curious how long these uploaded files will last. Github makes no guarantees about file attachments on issues. On the flip side, it doesn't seem like they advertise any deletion timeframe.

PR #6394, opened 2023-04-22
Update City of Palo Alto, CA

arch0345:


iandees:
Streetid seems to be the ID into another table, not the actual street name. Needs a little kore investigation into how to resolve that.

arch0345:
I noticed when doing a query with `STREETID` in outFields that it prints out the corresponding names for each of the IDs at the beginning:
https://services6.arcgis.com/evmyRZRrsopdeog7/ArcGIS/rest/services/StreetAddressData/FeatureServer/0/query?f=json&outFields=STREETID&where=1=1

Could we use this somehow?

iandees:
We would have to make some changes in https://github.com/openaddresses/batch-machine to support the mapped names. It could be done, but isn't supported right now.

iandees:
Superseded by #8341, which covers the same source with a more complete update (addresses, parcels, buildings) and is passing CI. Closing in favor of that one.

PR #6402, opened 2023-04-22
Update City of Laguna Niguel, CA

arch0345:


iandees:
Looks like this Esri server requires authentication of some sort.

iandees:
This time it's a timeout, meaning the database behind the Esri service is probably not configured properly. I'll merge this as-is, as it might work intermittently.

PR #6421, opened 2023-04-23
Update North Carolina

arch0345:


arch0345:
Looks like all jobs got stuck, can these please be rerun?

iandees:
I think their Esri service is just too slow to download all data points from it. I think we ran into this last time we tried to use it. Can you try updating to the new GDB at https://s3.amazonaws.com/dit-cgia-gis-data/NCOM-data/addresses/addressnc-addresses-04-14-2023.zip (as found at https://assets.nconemap.gov/pages/hub/file-downloads-min.html)

PR #6447, opened 2023-05-11
Add Comox Valley Regional District

arch0345:
Closes #6076

Issue #6076, opened 2022-07-25
Comox Valley Regional District, BC, CA


BasilLangevin:
Source: http://imaptoo.comoxvalleyrd.ca/imap87320977492837458/rest/services/CVRDHouseNumbers/MapServer/70

arch0345:
Added in #6450

arch0345:
Added in #6450

PR #6450, opened 2023-05-15
Add Comox Valley Regional District

ingalls:
Closes: https://github.com/openaddresses/openaddresses/pull/6447


iandees:
For #6076

PR #6456, opened 2023-05-17
Added OSM's BANO (French Open address database)

ycouble:
I tried to write the json spec for the dataset, but I'm not quite sure about the conform part.

- dataset needs to be un-gzipped
- there is no header in the csv file, headers are given in their readme (https://bano.openstreetmap.fr/data/lisezmoi-bano.txt)

Hope this can help and the DB is elligible


iandees:
Yep, for reference https://github.com/openaddresses/openaddresses/pull/236.

PR #6547, opened 2023-06-15
Add Madison County, FL

arch0345:
Source zip (from [FDOR download portal](https://pointmatch.floridarevenue.com/General/AddressFiles.aspx)): [MADISON_2023-07-01.zip](https://github.com/openaddresses/openaddresses/files/11761462/MADISON_2023-07-01.zip)

### DIFF
--- a/sources/us/fl/madison.json
+++ b/sources/us/fl/madison.json
@@ -1,0 +1,59 @@
+{
+    "coverage": {
+        "US Census": {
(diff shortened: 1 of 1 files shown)


iandees:
Hmm, looks like some data on null island for the address points. That's OK, but I imagine it might confuse people. In an ideal world we could tell the county that the data is bad.

PR #6016, opened 2022-02-24
swap au/vic/statewide to gdb source

andrewharvey:

PR #6909, opened 2023-08-29
skip removed files and skip non json files in run_changed_sources

andrewharvey:
Removed files won't exist and sometimes non-JSON files exist within the sources directory (eg. .md for documentation).


andrewharvey:
@iandees I don't know Python so :crossed_fingers: this works. Aims to address the failure in https://github.com/openaddresses/openaddresses/pull/6016

PR #6912, opened 2023-08-31
au/act new source

andrewharvey:


sources/au/act/statewide.json
```diff
@@ -10,9 +10,9 @@
     "layers": {
         "addresses": [
             {
-                "name": "state",
-                "data": "http://actmapi.act.gov.au/datadownload/Shapefile/MGA94_55/ACT_data.zip",
-                "website": "http://app.actmapi.act.gov.au/download.html",
+                "name": "ACTGOV Addresses",
```
iandees, line 13: It doesn't like the spaces you added in the names. This could likely be fixed with some correct escaping in the script or by switching the names back to what they were.

andrewharvey: oops I realised now the documentation says alphanumeric + underscores.

andrewharvey:
Looks like it doesn't like these geojson files @iandees :shrug:

iandees:
These files are generated from an ArcGIS endpoint if you want to switch to the endpoint directly instead of using their download version.

andrewharvey:
> These files are generated from an ArcGIS endpoint if you want to switch to the endpoint directly instead of using their download version.

I had assumed the direct GeoJSON links were safer and more reliable, but maybe that's no the case. Regardless I'm trying ESRI now.

iandees:
> I had assumed the direct GeoJSON links were safer and more reliable

They _should_ be fine. It's not clear why process-one is unhappy with the file. We don't see sources with GeoJSON downloaded directly very often.

PR #6913, opened 2023-08-31
Update brisbane_city_council.json

andrewharvey:
This updates au/qld/brisbane to use fresh 2024 data and does additional processing to select the first geometry of an address where multiple geometries exist in the data.


andrewharvey:
Some records like

`,147,ABBOTSFORD,Road,,BOWEN HILLS,4006,Uprn (Official Location),HAMILTON,L.2 RP.847828 PAR NTH BRISBANE,6964892.91|6964894.581,504030.63|504030.131                                                                                           `

Use a pipe in the easting/northing column. Can't see how we can handle this without pre-processing.

iandees:
What the heck. CSV isn't that hard. Just use it! 😭

PR #6926, opened 2023-09-06
Update Montana

arch0345:


iandees:
The buildings dataset seems suspiciously small. I wonder if they only LIDAR'd a couple spots?

arch0345:
Looks like I forgot to specify the layer, should be fixed now

PR #6943, opened 2023-09-10
Update city_of_kingston.json

arch0345:
Closes #5382


iandees:
I think the GeoJSON parsing code in machine looks at the first GeoJSON feature ([here](https://github.com/openaddresses/batch-machine/blob/master/openaddr/conform.py#L651)) and uses that to build the columns for the CSV to pass through the rest of the system. I think some other feature in the GeoJSON has a property that the first feature doesn't have, which is breaking the GeoJSON -> CSV conversion step here.

iandees:
Closing as stale (no activity since 2024-11). Feel free to reopen or resubmit — we'll also independently check whether Kingston's source has changed.

PR #6952, opened 2023-10-09
Update statewide.json

dpsspatial:
The path to the CO Statewide address point file has been updated. There are issues with the old file that have been corrected in this file. These issues go upstream to Geocode.Earth, for example, and are used by our District's School Finder.


arch0345 (changes requested):
Thanks for updating this! The `file` attribute should be changed to reflect the new file name for the gdb

PR #6958, opened 2023-10-15
Update Texas

### LINKED ISSUE #6860: 2023 Texas statewide addresses
I noticed that TNRIS has released its 2023 address points, however it looks like they only have downloads for each individual county as opposed to one covering the whole state:
https://data.tnris.org/collection/?c=578ac3ee-0d9f-4bd3-bdeb-509af9b5b646

arch0345:
Closes #6860


iandees:
Hah, looks like the CI ran out of disk space.

PR #6965, opened 2023-10-22
Add Kentucky statewide addresses

arch0345:


justinelliotmeyers:
Copyright confirmed: Public Domain

Wed, Mar 27, 2024, 9:04 AM Harding, Watson C (KOHS) <[watson.harding@ky.gov](mailto:watson.harding@ky.gov)> wrote:

Good Morning Justin,
I checked with the folks at the Division of Geographic Information, who manage KY’s GIS data clearinghouse, and this was their reply:
“All data in the Open Data Portal is in the public domain. With that being said, we’ve always been advised to include a standard disclaimer along with each published item/metadata record.

PR #6968, opened 2023-10-27
Update to make es script (Spain) working with current python and current data

frodrigo:
The current data for whole Spain are broken and no more up to date.
- reversed lat/lon
- city name are missing for all records

I updated the scripts to
- switch from pyhton2 to pyhton3
- fix the Makefile

Improved:
- the doc
- strip street name starting with space

Note, I do nothing about the reversed lat/lon but the current data produced by this PR are now ok.

Fix also #6222, close other PR #6240


iandees:
Thanks for improving this.

frodrigo:
@iandees Thank for merging. I have to upload the result of the script into the cache ? (and how ?)

frodrigo:
Running the batch-machin on the output show an error. I will fix the scripts.

PR #6969, opened 2023-10-31
Fix es script and  source

frodrigo:
Fix ouput of script by removing intermediate header. So it makes the data right for batch process.

Fix the source config to also use the municipality name.


frodrigo:
CI error out of this PR scope.

PR #6980, opened 2023-11-15
Update Oregon statewide addresses

arch0345:
This gdb contains statewide address data extracted from NAD r8. The current source we're using is missing address data from some counties. Not sure how long this file will last on Slack, so it might be worth uploading it to data.openaddresses.io


iandees:
We need to bring back the data upload tool on the website...

PR #7035, opened 2024-01-04
Update denver.json

dpsspatial:
Removing 2 redundant and error-inducing sources. Please reach out to me if there are any more clarifications needed. As we have been using OA and City of Denver address points (as well as parcels, and many other Denver open data sources), I feel the proposed changes below are justified: 

- Denver Parcels - main reason is there are missing zip codes for all addresses, and otherwise, contains redundant information that is more accurately found in the Denver Addresses file, which is maintained perfectly by the City and County of Denver. The missing zip codes are also causing issue through Geocode.Earth and our Denver Public Schools geocoding system, which pulls Denver and CO through the US_WEST aggregate. 

- building footprints 2014 - in addition to being 10 years out of date, the data within provides no additional geocoding information

### DIFF
--- a/sources/us/co/denver.json
+++ b/sources/us/co/denver.json
@@ -40,35 +40,7 @@
                 "protocol": "http",
                 "compression": "zip"
             }
-        ],
-        "parcels": [
-            {
-                "name": "county",
-                "website": "https://www.denvergov.org/opendata/dataset/city-and-county-of-denver-parcels",
-                "license": "CC BY 3.0",
-                "attribution": "City of Denver Open Data Catalog",
-                "data": "https://www.denvergov.org/media/gis/DataCatalog/parcels/shape/parcels.zip",
-                "protocol": "http",
-                "compression": "zip",
-                "conform": {
-                    "format": "shapefile",
-                    "pid": "SCHEDNUM"
-                }
-            }
-        ],
-        "buildings": [
-            {
-                "name": "county",
-                "website": "https://www.denvergov.org/opendata/dataset/city-and-county-of-denver-building-outlines-2014",
-                "license": "CC BY 3.0",
-                "attribution": "City of Denver Open Data Catalog",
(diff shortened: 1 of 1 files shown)


iandees:
I don't think I'm willing to remove these. The parcel and building datasets that we record aren't necessarily for geocoding. If your primary usecase is geocoding, I would suggest sticking with the address datasets.

dpsspatial:
@iandees If the project is to collect and aggregate addresses, what is the purpose of building footprints, especially if they are not maintained by the data owner? Additionally, what is the purpose of parcel outlines if they are otherwise containing redundant information? 


> I don't think I'm willing to remove these. The parcel and building datasets that we record aren't necessarily for geocoding. If your primary usecase is geocoding, I would suggest sticking with the address datasets.

iandees:
The project's name is OpenAddresses, but we also gather parcel and building data alongside addresses.

dpsspatial:
As a long-time user of several downstream applications of this project, my feedback is the extra non-address datasets are causing issues with geocoding, which I should hope is the primary application of the OpenAddresses project.

iandees:
I don't understand why the non-address datasets would cause issues with geocoding. Your processing should skip over them, right?

dpsspatial:
It's more the redundant parcel data and missing zip codes that causes the issue. The building footprints are just unnecessary and out of date.

iandees:
It's better to have some data, even if it's out of date or unsupported, than no data. If it's not useful for geocoding, then please feel free to skip it when processing the data.

dpsspatial:
> It's better to have some data than no data, even if it's out of date or unsupported. If it's not useful for geocoding, then please feel free to skip it when processing the data.

Data integrity quote of the year!!!

iandees:
Feel free to put it on a t-shirt, as long as you attribute it to this thread.

While we were going back and forth, I looked for more recent data and updated the buildings dataset here in the above PR. Thanks for pointing it out!

PR #7036, opened 2024-01-05
Update Denver source to use more recent buildings

iandees:
As pointed out in https://github.com/openaddresses/openaddresses/pull/7035, the buildings dataset for Denver was using older data. This updates to use a more recent (2020) release.


dpsspatial:
There is no address information in this dataset.

iandees:
Yes, correct. It's a buildings dataset.

PR #7073, opened 2024-02-12
added other address layer props, added buildings/parcels layers

trescube:
Updates:
- added `unit`, `city`, and `postcode` to addresses layer
- added `parcels` layer
- added `buildings` layer


trescube:
Re-run this please.

iandees:
I'm not sure that the re-run thing works with "data-pls" (it was a feature of the older CI system). It looks like that job failed due to timeout. The server might not be up to handling the type of queries we make?

PR #7093, opened 2024-03-26
Update us/nc/statewide

### LINKED ISSUE #7003: Update North Carolina statewide
The state's NC OneMap spatial portal has a statewide address map:

https://www.nconemap.gov/datasets/nc-onemap-addresses-addressnc/explore?location=35.292884%2C-80.516145%2C14.99

that has this underlying ArcGIS REST service:

https://services.nconemap.gov/secure/rest/services/AddressNC/NC1Map_Addresses/MapServer/1

rmx90210:
Use live address points, parcels, and buildings from ArcGIS Online (may stay more up to date than the ZIP files being used previously)

Closes #7003

### DIFF
--- a/sources/us/nc/statewide.json
+++ b/sources/us/nc/statewide.json
@@ -12,16 +12,14 @@
         "addresses": [
             {
                 "name": "state",
-                "data": "https://s3.amazonaws.com/dit-cgia-gis-data/NCOM-data/addresses/AddressNC-addresses-07-03-2024.zip",
                 "website": "https://www.nconemap.gov/datasets/nconemap::address-points-addressnc/about",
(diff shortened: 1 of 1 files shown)


iandees:
I think we had problems downloading from the Esri endpoint for the state for some reason (a slow database server behind it, if I remember right) and had to switch to their download package.

bertday:
Thank you for the background @iandees. Indeed it looks like the job is hanging. I opened openaddresses/batch#380 just to document that the job has been running for 12 hours but seems to be non-responsive (should it have timed out by now?)

In any case, happy to close this PR for now if the source is already known to be no-good.

PR #7094, opened 2024-03-26
Slovenia: use first_non_empty function for street name

### LINKED ISSUE #7091: Slovenia countrywide lost about half of its streets
https://batch.openaddresses.io/data/591/history#map=5.84/45.717/16.54

image

Last run with streets intact was Jan 25, 2023, which coincides with the last commit affecting this source (9d033d76882a812193622768fe47989e8aa664e5). So this seems like an upstream data issue.

Kindly throwing this on your radar @stefanb, if you happen to see anything amiss! 😄

stefanb:
Fallback to city(village) name if there are no streets (common for smaller places)

Fixes #7091

Also updated attribution, removed year attribute

### DIFF
--- a/sources/si/countrywide.json
+++ b/sources/si/countrywide.json
@@ -20,10 +20,9 @@
                     "text": "CC BY 4.0",
                     "url": "https://creativecommons.org/licenses/by/4.0/",
                     "attribution": true,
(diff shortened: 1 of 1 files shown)


iandees:
@ingalls it looks like data-pls is using an old schema from somewhere. Can you point me in the right direction of updating that?

PR #7118, opened 2024-04-09
Update street column in us/sc/jasper.json

anthony-moreau-qapa:
The street labels currently use the FULL_STREET field but this field contains everything including the street number. Change to get only the street label using other fields so the output is consistent with other sources.


trescube:
No idea why this failed, I kicked it off again.

trescube:
@ingalls any idea what the error `AssertionError: R2_BUCKET must be set` means?

trescube:
@anthony-moreau-qapa can you create a branch for this repo w/o forking?  Otherwise I can do it.

PR #7171, opened 2024-04-21
Update CONTRIBUTING.md to remove mention of deprecated upload-cache

andrewharvey:
Since the upload cache URL is deprecated and no longer functioning, remove it from the contributing guide.


iandees:
Can you try uploading something with https://batch.openaddresses.io/upload ? It's meant to replace that URL.

andrewharvey:
I've updated the URL in this PR, though it's not working for me but that's a separate issue to track compared to the documentation update.

image

iandees:
Thanks! We'll look at the upload system...

PR #7189, opened 2024-04-30
Update Alpine County, CA

arch0345:


iandees:
Looks like there might be a typo in the schema 🤔

iandees:
Ah nope, typo in the source file here. I misread the error from pre-commit. Updated!

iandees:
The `regexp` function uses this code [here](https://github.com/openaddresses/batch-machine/blob/master/openaddr/conform.py#L914-L915). I wonder if `re.sub()` has a different behavior when the `pattern` doesn't match at all 🤔

skylarmt:
This PR should be closed because https://github.com/openaddresses/openaddresses/pull/7520 was merged instead.  And my regexes are a lot easier on the eyes :)

PR #7213, opened 2024-05-09
La Plata County, CO, USA: updated data urls for addresses and parcels

trescube:


trescube:
This source seems to be configured for 100 records/request, causing 355 requests which eventually times out.

iandees:
Closing as superseded — master already updated to the live `gis.lpcgov.org` URL. The URL proposed in this PR (`gis.laplata.co.us`) now just redirects to that same domain, so there's nothing left for this PR to add.

PR #7241, opened 2024-05-19
update attribution string for au/qld/statewide

andrewharvey:


sources/au/qld/statewide.json
```diff
@@ -21,28 +21,27 @@
...
                 },
                 "note": "must use web interface & email workflow to download fresh data.",
                 "conform": {
                     "format": "gdb",
                     "file": "DP_QLD_DCDB_WOS_CUR.gdb",
-                    "layer": "QLD_ADDRESS_QAMF",
-                    "unit": "UNIT_NUMBER",
+                    "layer": "QLD_LOCATION_ADDRESS",
+                    "unit": [
+                        "UNIT_TYPE",
```
andrewharvey, line 33: it's probably better to leave out the unit type, as the type is not as important and more likely to confuse data consumers. In a future OA schema it could go into its own field.

image

iandees: The "unit" is meant for sub-address stuff like this. We want to get "APT 1" or "SHED" in this field.

andrewharvey: Yes but I'd just expect `["UNIT_NUMBER", "UNIT_SUFFIX"]` here, otherwise data consumers need to parse out the prefix and need to implement rules specific to each OA source since what these prefixes mean is specific to the source. For a map, you'd likely want to omit the prefix and label like "1A" but for a full postal address you'd want the prefix, eg. "Unit 1A", expanding the "U" into "Unit".

iandees: I think it's still useful to include UNIT_TYPE here, otherwise you can't tell the difference between "APT 1  A" and "U 1 A" or "VLLA 1 A". I agree that the data source's encoding/shortening in "UNIT_TYPE" isn't ideal, but it's still useful and important information to include in our output.

If we know what the encodings are via documentation, we could use a conform function to expand them, I think.

PR #7245, opened 2024-05-22
Update mx data countrywide.json

crebollobr:
This data is from INEG processed by addressforall.org

### DIFF
--- a/sources/mx/countrywide.json
+++ b/sources/mx/countrywide.json
@@ -11,31 +11,25 @@
         "addresses": [
             {
                 "name": "country",
-                "data": "https://www.dropbox.com/s/vd0i1enmc69145j/mexico_ne.zip?dl=1",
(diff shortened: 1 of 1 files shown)


iandees:
Sorry for radio silence on this: I hadn't merged this yet because I was still trying to understand the PDF linked in the license. It doesn't look like "sharealike" should be true here. We should also link to the original license at https://www.inegi.org.mx/inegi/terminos.html. And we should make sure any processing that was done is described in the pull request and scripts used to create the output are documented (and possibly stored in this repo's `scripts/` directory).

ThierryAJean:
@iandees Here is the description of how the data was filtered: 
https://github.com/digital-guard/preserv-MX/tree/ineginewversion/data/_pk0002.01#-geoaddress
complemented by download/rebuild steps description at https://github.com/digital-guard/preserv-MX/blob/ineginewversion/data/_pk0002.01/attachment.md

iandees:
Thanks for the links and explanation, Thierry.

PR #7252, opened 2024-05-28
Update Mexico Countrywide

crebollobr:
update INEGI data
fix license share-alike flag and url of license

### DIFF
--- a/sources/mx/countrywide.json
+++ b/sources/mx/countrywide.json
@@ -11,31 +11,25 @@
         "addresses": [
             {
                 "name": "country",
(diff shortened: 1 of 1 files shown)


iandees:
This is an updated version of the change proposed in #7245, where there is more discussion about this data source and how it was built.

iandees:
Thanks for the change and updating the data!

I'll merge this in as-is, but I'm curious if we can add back the city information to the address.

crebollobr:
I will work in put city in data.
This is next step.
I will merge again with new data
thanks

iandees:
Perfect, thank you!

PR #7253, opened 2024-05-30
Add US National Address Database source

jwass:
Adds the US NAD (https://www.transportation.gov/gis/national-address-database) source file

### DIFF
--- a/sources/us/countrywide.json
+++ b/sources/us/countrywide.json
@@ -1,0 +1,39 @@
+{
+    "coverage": {
+        "ISO 3166": {
+            "alpha2": "US",
+            "country": "United States"
+        },
+        "country": "us"
+    },
+    "schema": 2,
+    "layers": {
+        "addresses": [
+            {
+                "name": "country",
+                "protocol": "http",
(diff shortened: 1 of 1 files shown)


jwass:
* I wasn't able to test this on a subset of the NAD as I can't get the osgeo/gdal dependency installed on my system at the moment.

Should the data be downloaded and put on OA infrastructure? I just included the main download link for now.

iandees:
The `data` attribute should be a URL to download the data. It's best if that can be a URL directly from transportation.gov in this case.

iandees:
Looks like the CI system ran out of disk space downloading this 😄 . I'll look at bumping it up like we did with the weekly run in https://github.com/openaddresses/batch/commit/1252532842ae78ddf6a665e00ef5e76a597a71f2.

jwass:
> The data attribute should be a URL to download the data. It's best if that can be a URL directly from transportation.gov in this case.

Ha - I just realized I didn't include the url to the actual zip file but just the link text... Thanks for the fix.

iandees:
I think this is failing with out of memory now. It was failing with disk space before. The weekly run should complete, though.

Issue #7257, opened 2024-06-14
Possible incorrect mapping for NZ data


mtravis:
Hi there

Just taking a look at the [NZ countrywide.json](https://github.com/openaddresses/openaddresses/blob/master/sources/nz/countrywide.json#L43-L44) and it struck me that the mappings for don't seem quite right. 

```
"city": "suburb_locality",
"district": "town_city", 
```

If you query the data you see something like this

` 'number': 75A, 'street': McElwee Street, 'unit': , 'city': Jervoistown, 'district': Napier,`

but in the source data it's the other way around? 

![image](https://github.com/openaddresses/openaddresses/assets/1799238/f6c633b8-b02d-4d5b-8af9-2d7a5a7964a2)

@andrewharvey  I'm assuming this could be intentional so the suburbs level is captured? Is this common elsewhere in OA schemas?

hamishcampbell:
Hi there,

Not an expert but familar with the sources. Will ask others to verify, but:

Given the hierarchy of addresses, the current configuration for city/disctrict looks **correct**. We essentially don't use the "distrcit" (or territorial authority) for addressing purposes, and the data is structured such that `suburb_locality` + `town_city` will give an addressable location, dispite the apparent mismatch. (e.g. from the example above Jervoistown is a suburb of the city of Naper).

The source data used above is the "NZ Street Address" dataset, which was superceded in Jan 2023 by the "NZ Addresses" data available at https://data.linz.govt.nz/layer/105689-nz-addresses/ - it looks like the reference to layer ID 53353 needs ot be updated to 105689. The linked data dictionary at https://data.linz.govt.nz/document/24489-nz-addresses-data-dictionary/view/ describes how it should be used.

Note that it includes full address examples. It looks like `full_address_number` can be used in place of the existing `number` definition (since it is "All number components concatenated for an address").

andrewharvey:
@mtravis I think to be consistent with other global sources within OA, the NZ "town_city" field should be mapped to the OA "city" field, however OA doesn't have a field corresponding to the NZ "suburb_locality" and since this field is essential for addressing in NZ, bumping the NZ "town_city" up to the OA "district" to leave space for the NZ "suburb_locality" in the OA "city" field as in the current schema mapping is the best solution we have.

The hierarchy order of both the NZ data (`suburb_locality`, `town_city`) and OA (`city`, `district`, `region`) must be retained with the mapping we use.

[...]

PR #7258, opened 2024-06-17
update nz/countrywide source

andrewharvey:
1. Updates the source URL per https://github.com/openaddresses/openaddresses/issues/7257#issuecomment-2171881758

> the "NZ Street Address" dataset, which was superceded in Jan 2023 by the "NZ Addresses" data available at https://data.linz.govt.nz/layer/105689-nz-addresses/

2. Fixes a small number of number formatting edge cases where there was both an `address number_suffix` and a `address_number_high`. For example "1A-7A Puka Place, Inglewood" which shows the suffix applying to both the low and high number parts, which is something the current mapping we use gets wrong as OA will format it as "1A-7"

### DIFF
--- a/sources/nz/countrywide.json
+++ b/sources/nz/countrywide.json
@@ -11,8 +11,8 @@
         "addresses": [
             {
                 "name": "country",
-                "data": "https://data.linz.govt.nz/services;key=441c0097257f45ab84d5ea4d24723f23/wfs?service=WFS&request=GetFeature&outputformat=CSV&typeNames=layer-53353",
-                "website": "https://data.linz.govt.nz/layer/53353-nz-street-address/",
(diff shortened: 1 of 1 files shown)


iandees:
Are we happy with the mapping here? The output looks OK to me.

andrewharvey:
> Are we happy with the mapping here? The output looks OK to me.

Hang on I'm just taking a look at the output now and immediately noticed that some include "Level N" in the number which isn't right.

andrewharvey:
Ok the `full_address_number` includes level data, which is hard to split out, so I think best we revert back to the previous behaviour of formatting the address number, and just accept that we won't have the prefix/suffix on the high ranges.

andrewharvey:
This looks good to me now @iandees, as mentioned a small number of the address number fields are slightly incorrect (1A-7) instead of (1A-7A), which I think is okay until there is a better solution.

PR #7262, opened 2024-06-23
Update countrywide.json

crebollobr:
Update inegi data with region and district. I do not found city information on data.


crebollobr:
i will change some data

Issue #7266, opened 2024-06-25
Updating sources/fi/countrywide-fi.json


jwass:
Looks like sources/fi/countrywide-fi.json hasn't been updated in 3 years. I've downloaded updated data from 2024-05-17. Looks like the source is .7z so I'd just need to convert to .zip.

I'll do it if there's a place for me to upload.

iandees:
If you want to upload to https://www.dropbox.com/request/rUcmasTnxZkiDR6rJc4l I will put it in our S3 bucket.

We used to have a way for contributors to upload directly themselves, but it doesn't work at the moment.

jwass:
@iandees Done. Also put up #7267 for posterity

iandees:
Ok, uploaded. Use the following URL in a source update:

```
https://data.openaddresses.io/cache/uploads/iandees/9cc3a4/suomi_osoitteet_2024-05-14.zip
```

PR #7267, opened 2024-06-26
Add script to process fi/countrywide-fi data

jwass:
Just downloads the .7z, unzips it using `7z x`, then zips it back up with zip


jwass:
Accompanies the upload in #7266

PR #7275, opened 2024-07-02
Remove run_changed_source job running on Github

iandees:
Since the data-pls process seems to be running ok now.


iandees:
Just so it's not failing and causing confusion on PRs from forks...

PR #7304, opened 2024-08-02
Update City of Norfolk, VA

### LINKED ISSUE #7303: USA Data: Norfolk VA -New Data Available, Existing one seems outdated
New Updated data is available at - https://data.norfolk.gov/Real-Estate/Norfolk-Master-Address-List/jagz-9a37/about_data

arch0345:
Closes #7303


iandees:
Looks like the CVS code is capitalizing the key names from conform. I don't remember how we got around this before.

It probably makes sense to fix batch-machine so that it doesn't call `upper()` on the key names...

PR #7306, opened 2024-08-03
Acerra is in Campania (ISO_3166-2 IT-72)

musuruan:
I used the correct ISO_3166-2 code for Campania


iandees:
Failure was related to Mapbox vector tiles.

PR #7307, opened 2024-08-06
Fixes for Biella and Verona

musuruan:
Renamed files to avoid city_of_ just like Torino, Venice, Bologna, Ferrara, etc.
   
Removed "region" from Biella. It's not necessary and moreover Biella is in Piemonte and not in Lombardia.
    
Updated website for Biella Open Data.


iandees:
Looks like number localization errors? The decimal for the latitude and longitude fields is a comma in the source data. We don't have a great way to resolve that right now. 🤔

musuruan:
How was Verona data committed in the first place? I've just renamed the file...

iandees:
Yea, we can deal with that later. Thanks!

iandees:
The number localization error appears to be a warning that we can ignore here.

I think Mapbox has started to serve error messages to our map image rendering pipeline, which breaks that step of the build.

PR #7363, opened 2024-08-17
Update Trempealeau County, WI

arch0345:


iandees:
"Unknown type: MULTISURFACE" 🤯 

Not sure how to resolve that

iandees:
Looks like MULTISURFACE is a geometry type that supports curves and was introduced in gdal ~3.13. Unfortunately Shapely (the Python library we use to interface with gdal) doesn't support the (multi)surface type.

We might have to find another source.

iandees:
I updated to a gdal 3.11 image over in https://github.com/openaddresses/batch-machine/pull/99 and it still doesn't seem to work.

PR #7419, opened 2024-09-08
Add statewide address source for North Dakota (us/nd)

skylarmt:


skylarmt:
The parcel coordinates are in a different layer and I didn't realize at first.  It should be possible to link the two layers since they have common unique keys, with every parcel having zero or more addresses.  Averaging the parcel lat/lon points would get a coordinate pair near the center of the parcel.  I don't know how to do this with OpenAddresses.

Here's an ArcGIS map where you can click on any parcel, then click "Table", scroll to the far right, and get the address(es) in that parcel: https://www.arcgis.com/apps/mapviewer/index.html?layers=ac6da1176038457db16e8debe3f1abaf

Here's a direct download link to the GDB data: https://ndgov.maps.arcgis.com/sharing/rest/content/items/c9e6956684c14fa299ee9e38c90dc003/data

Hopefully someone with more experience than me with this can make it work.  All the data seems to be there.

PR #7428, opened 2024-09-19
Schleswig Holstein districts

jeffdefacto:
About 126k features in the dataset have district information so we should add them to the OA dataset.


jeffdefacto:
Actually looking closer at the OA schema, district seems to be higher level than city while these are subcity districts so I believe this doesn't fit. Closing the PR.

PR #7443, opened 2024-09-23
Updated SK data source

jeffdefacto:
The previous datasource was returning an empty csv and the data was never actually building. This is the latest data from the government website. The license on the new data is CC0 now which I've updated and changed attribution to false.

https://data.slovensko.sk/datasety/b27f57f1-7e76-45e0-8968-631f9176b2e9

### DIFF
--- a/sources/sk/countrywide.json
+++ b/sources/sk/countrywide.json
@@ -11,35 +11,35 @@
         "addresses": [
             {
                 "name": "country",
-                "data": "http://epsilon.sk/sk-countrywide.zip",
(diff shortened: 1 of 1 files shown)


jeffdefacto:
It would be good to get the data hosted onto OA for resiliency but its fine to ship this using my web hosting for now.

iandees:
Use `https://data.openaddresses.io/cache/uploads/iandees/112e44/Adresy_kraje_04_2023_UTF8.csv.zip`

jeffdefacto:
Thank @iandees!

PR #7445, opened 2024-09-24
Adding german city names using constants

jeffdefacto:
This adds cities using constants to define them for some of the smaller datasets


Vainock:
You will need to replace `constant` with `static`. There is no `constant` function.

PR #7451, opened 2024-09-30
CZ country data update

jeffdefacto:
Updated the datasource to latest release and current license. Data looks to be drop in compatible with only minor changes to the config schema. The source is a file of zipped municipal CSVs that I've combined into one. Using my hosting to test out the PR and then https://jeffu.dev/data/CZ_addresses.csv.zip will need to be rehosted.

### DIFF
--- a/sources/cz/countrywide.json
+++ b/sources/cz/countrywide.json
@@ -20,7 +20,7 @@
                 },
                 "protocol": "http",
                 "compression": "zip",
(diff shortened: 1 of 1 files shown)


jeffdefacto:
@iandees The data looks good. Can you rehost the zip when you have a chance?

iandees:
Use https://data.openaddresses.io/cache/uploads/iandees/bbd719/CZ_addresses.csv.zip

PR #7457, opened 2024-10-07
Add sources for de/by

Vainock:
Adds available sources for Bavaria, Germany.


iandees:
Hmm. This is the first time we've combined data for multiple disparate coverages into a single file. Why not split them up into multiple files?

PR #7460, opened 2024-10-08
Updating Japan Regional Data

jeffdefacto:
This updates all Japan regions to 2023 data, adds city and region columns, and fleshes out license details.

The data urls appears to be stable and will just need the versions bumped up each year.


jeffdefacto:
I believe the data is ready to go now

PR #7472, opened 2024-10-14
Improve Germany sources

Vainock:
Make de/be consistent with other states, improve number fields for de/nw and add new sources

Closes https://github.com/openaddresses/openaddresses/pull/7445
Close https://github.com/openaddresses/openaddresses/issues/4217


Vainock:
@ingalls Please publish v5.3.0 of the oa npm package, which includes the fixes for the constant function. Following that requires updating the package to that version in the batch repository which then fixes https://batch.openaddresses.io/job/469808. Thank you!

Vainock:
Seems like this still needs pushing the above mentioned updates to the cloud.

iandees:
🤔 Guess this user deleted themselves.

PR #7484, opened 2024-10-17
Propose Schema V3

Vainock:
The source schema defines what the entire OpenAddresses project is built upon: +2600 files that provide instructions on how to convert some structure into a defined structure. This conversion step uses files as its input that are often handwritten, copied together from other files, were never consistently updated according to their defining schema and have never been truly verified to match what they actually intend to describe.

Goals and considerations redesigning the schema:
- Make it consistent
- Make it country-neutral
- Define a clear logical structure
- Define clear patterns
- Give freedom only where intended and needed
- Remove deprecated, nonsensical, duplicated and non-existing things

Changes to all files:
- Add patterns to strings
- Add requirements to objects and arrays such as a minimum amount of items
- Improve and add more descriptions

Changes to the main file:
- Update schema version to 2020-12
- Improve the `coverage` object, making it more logical and structured
- Remove the unused `official_website` and `official_phone` objects. This data is already provided per layer.
- Make the `layers` object contain only a single layer object (previously it contained arrays containing the layers), making a source contain only one of each layer type. Currently, different regions (counties, cities, ...) are split into their own files. This change makes it clear to make use of that style.
- The base.json now contains all objects that were previously duplicated in all layer files, providing a better structure and clear separation.

Changes to all layers:
- Remove the `name` object. The `coverage` object already describes... the coverage. The way this is currently used seems redundant.
- Remove the `email` object. I find the actual purpose described in the CONTRIBUTING.md questionable, unrealistic and prone to misuse. An `email` object already exists in the `contact` object.
- Lowercase the `ESRI` field in the `protocol` object for consistency
- Remove possibilty to provide a single string for the `license` object. This behaviour was deprecated.
- Remove the deprecated `attribution` object.
- Convert the `year` object to a `date`, creating the possibilty to insert a specific date.
- Conditionally require the `date` object only when `frequency`=`once`
- Remove unused `timeout` object. Proving this type of information in a source file is in my opinion wrong and unrealistic. This should better be handled in the executing script.
- Remove references to the unused `base` function which does not exist in the codebase.
- Conditionally require the `layer` object only when `format`=`gdb`

Changes to the addresses layer:
- Convert `city`, `region` and `district` objects to an `address_levels` object. Taken from the OvertureMaps docs, this object makes addresses country-neutral. Finding the equivalents for	 these terms is not so easy for all countries. This makes it clear which level one wants to describe.
- Require the `format` for clarity
- Remove the `country` as this can be inherited from the `coverage`
- Conditionally require objects when `format`=`csv` (such as `csvsplit` or `headers`)

Changes to the pacels layer:
- For consistency, rename the `pid` object to `id`
- Allow the `id` to be a string and not just a function. It has always beed used this way but never officially described as such.

Ideas:
- Convert the `frequency` object to a `static_data` boolean that indicates whether the data is changing or static, such as for preprocessed files in the cache (equivalent to `frequency`=`once`). Does it really make a difference when the data is crawled weekly and not just monthly?  Why not crawl all non-static sources monthly and publish on the same day each month?

I'm of course willing to make the needed migrations and source code changes. I'm encouraged to continue working with you on improving this project in every aspect.

### DIFF
--- a/schema/schema_v3/base.json
+++ b/schema/schema_v3/base.json
@@ -1,0 +1,222 @@
+{
+    "type": "object",
+    "if": {
+        "properties": {
+            "frequency": {
+                "const": "once"
+            }
+        }
+    },
+    "then": {
+        "required": [
+            "date"
+        ]
+    },
+    "else": {
+        "not": {
+            "required": [
+                "date"
+            ]
+        }
+    },
+    "required": [
+        "protocol",
+        "data"
+    ],
+    "properties": {
+        "protocol": {
+            "type": "string",
+            "enum": [
+                "http",
+                "ftp",
+                "esri"
+            ]
+        },
+        "data": {
+            "type": "string",
+            "description": "A URL referencing the dataset. This should point to the raw data and not a web portal. If there isn't a good URL for the source, process the file to the desired format, upload it to a third party file hosting service and ask for uploading that file to the cache which then gives you this URL.",
+            "minLength": 1
+        },
+        "website": {
+            "type": "string",
+            "description": "A URL referencing the data portal.",
+            "minLength": 1
+        },
+        "compression": {
+            "type": "string",
+            "enum": [
+                "zip"
+            ]
+        },
+        "license": {
+            "type": "object",
+            "if": {
+                "required": [
+                    "attribution_name"
+                ]
+            },
+            "then": {
+                "required": [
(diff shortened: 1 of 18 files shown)


iandees:
Before proposing a solution could you describe specific problems you've encountered that you're interested in trying to solve?

PR #7486, opened 2024-10-21
Spain full country dataset

jeffdefacto:
https://www.cartociudad.es/ has full country data available that is significantly different than our current datasets so seems worth including.

There is some way to download them through INSPIRE [here](https://inspire-geoportal.ec.europa.eu/srv/eng/catalog.search#/datasetdetails?country=es&view=themeOverview&theme=none&resourceId=spaign_cartociudad_addresses) and that may be the better long term path for maintenance if someone can figure it out.

They also have the data available as geopackages through a site [here](https://centrodedescargas.cnig.es/CentroDescargas/buscadorCatalogo.do) which is where I ended up getting it, although its not possible to automate the download process unfortunately. The data comes in 52 regional geopackages but has some non address layers and needs to be converted to a compatible format anyway so I just combined them during that process into one file for simplicity.

### DIFF
--- a/sources/es/countrywide.json
+++ b/sources/es/countrywide.json
@@ -1,0 +1,48 @@
+{
+    "coverage": {
+        "ISO 3166": {
+            "alpha2": "ES",
+            "country": "Spain"
+        },
+        "country": "es"
+    },
+    "schema": 2,
+    "layers": {
+        "addresses": [
+            {
+                "name": "country",
+                "data": "https://jeffu.dev/data/spain_addresses.csv.zip",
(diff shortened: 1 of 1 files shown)


jeffdefacto:
This seems to be all working. @iandees can you rehost this dataset please?

https://jeffu.dev/data/spain_addresses.csv.zip

iandees:
https://data.openaddresses.io/cache/uploads/iandees/bbc9ff/spain_addresses.csv.zip

jeffdefacto:
Thanks Ian. Its ready to ship now.

PR #7490, opened 2024-10-24
Piemonte dataset updates

jeffdefacto:
Updates the statewide and Torino datasets


jeffdefacto:
@iandees Looks like the pietmont dataset will need to be rehosted. I couldn't get the WFS layer to completely work on its own unfortunately.

https://jeffu.dev/data/pietmont_addreses.csv

PR #7494, opened 2024-10-27
Update address source for City of Huntsville, AL

skylarmt:


skylarmt:
I think data-pls had a glitch, it has no error log and the only change was adding a typo to a field name.

iandees:
Can you use https://maps.huntsvilleal.gov/server/rest/services/Addresses/AddressPoints/MapServer/0 instead? Rest services tend to be updated more frequently?

PR #7496, opened 2024-10-27
Update address source for Cullman County, AL

skylarmt:
On hold pending https://github.com/openaddresses/batch-machine/issues/64
Note: The parcel and building sources are broken and I couldn't find working ones.


skylarmt:
The source apparently has latitude and longitude included but as normal data fields, not as geometry, and the batch server ignores where I set the fields manually.

iandees:
We support this for Esri endpoints, now 👍🏻

PR #7508, opened 2024-10-31
Latvia country wide dataset

jeffdefacto:
Ingestion of cc-by 4.0 latvia dataset.

Meta data is here with the license.
https://geometadati.viss.gov.lv/geoportal/rest/document?id={E7CD2078-0767-4686-B169-EDE511C23998}

The url updates every Monday with a new link using the current date so this will need to be periodically updated in the json to ingest fresh data. E.g. this week's url is [https://grafws.kadastrs.lv/atom/ad/20241028/Address.gml.zip](url) and next week will be [https://grafws.kadastrs.lv/atom/ad/20241104/Address.gml.zip](url)

Latest URL can be found here.
https://grafws.kadastrs.lv/atom/ad/atom_Address.xml

### DIFF
--- a/sources/lv/countrywide.json
+++ b/sources/lv/countrywide.json
@@ -12,7 +12,9 @@
             {
                 "name": "country",
                 "data": "https://grafws.kadastrs.lv/atom/ad/20241028/Address.gml.zip",
+                "note": "URL updates every monday to that date. Adjust to refresh data.", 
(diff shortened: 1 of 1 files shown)


iandees:
It doesn't seem like the street name is coming through. I wonder if it's just the sample?

jeffdefacto:
@iandees Thats just the sample not containing any. Rural addresses, which make up most of the dataset, have a name rather than a street or house number. Street addresses are present in places like Riga though and I confirmed that they are in the dataset above.

iandees:
Gotcha!

PR #7523, opened 2024-11-01
Update address fields for City of San Jose, California

skylarmt:


iandees:
For the record, I prefer to have the list of individual fields so that if we want to split them apart for a future schema update it would be a little bit easier if we know that there are individual fields in the source.

PR #7531, opened 2024-11-01
Updated Toscana data

jeffdefacto:
Updates Toscana to latest data which is from June 2022.

There are 40k or so more addresses, but most notably the license changed at some point to CC-BY 4.0.

Source is still at http://www502.regione.toscana.it/geoscopio/download/grafo_stradale/ and the data unfortunately still comes as a zip within a zip so won't build directly.

### DIFF
--- a/sources/it/52/statewide.json
+++ b/sources/it/52/statewide.json
@@ -13,7 +13,7 @@
         "addresses": [
             {
                 "name": "state",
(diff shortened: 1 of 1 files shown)


iandees:
It seems like the Mapbox tiles we're requesting for the preview image are failing for whatever reason.

Issue #3365, opened 2017-08-25
harris texas


justinelliotmeyers:
https://arcweb.hcad.org/server/rest/services/public/public_query/MapServer/0

PR #7534, opened 2024-11-02
Create address source for Harris County, Texas

skylarmt:
Their ArcGIS server seems to be very slow at queries but it eventually finished in about 10 and a half hours.  1505480 address records total.


skylarmt:
Looks like this PR closes [issue #3365](https://github.com/openaddresses/openaddresses/issues/3365)

PR #7571, opened 2024-11-21
Update Toscana attribution name

jeffdefacto:
Change `attribution text` to `attribution name` to match the typical standard


jeffdefacto:
@iandees I think the only issue here is the mapbox tiles are failing again while running the check.

iandees:
We shouldn't be using any Mapbox tiles anymore.

PR #7574, opened 2024-11-22
Update france addresses

jeffdefacto:


iandees:
Hm, I wonder where the schema it's using to check is...

jeffdefacto:
Thanks for taking a look!

iandees:
@ingalls I'm trying to tag and publish `oa` to get the schema change for gzip. When I run `npm publish` I get:

```
npm notice Publishing to https://registry.npmjs.org/ with tag latest and default access
npm error code E403
npm error 403 403 Forbidden - PUT https://registry.npmjs.org/oa - You do not have permission to publish "oa". Are you logged in as the correct user?
npm error 403 In most cases, you or one of your dependencies are requesting
npm error 403 a package version that is forbidden by your security policy, or
npm error 403 on a server you do not have access to.
npm error A complete log of this run can be found in: /Users/iandees/.npm/_logs/2024-11-25T17_04_28_437Z-debug-0.log
```

Can you please add me so that we can update the schema?

iandees:
Nick published the oa repo for us. I'm updating the dependency in #414 and then will deploy and re-run this PR.

nick-rv:
Hi, i was trying to download an up-to-date dataset of french addresses, but i saw that the job that builds it is failing : https://batch.openaddresses.io/job/498733#job:498733:log:14
This PR seems to be the solution, but i check remains.
Does this data-pls check need to be relaunched manually ?
Thanks in advance!

jeffdefacto:
@iandees The jobs seem to be stalling after downloading the .gz file. Any ideas for what is still going wrong?

iandees:
Yea, I bet the ungzip code is failing silently somewhere. Have you ever tried running `process-one` locally? You could see if it works for you locally. Otherwise I can try looking later.

jeffdefacto:
Yeah, I've been running it for the past few weeks to create less PR spam. I thought I had it working when I put the PR up but maybe not. Looks like it just gives a `Killed` messaged after downloading.

iandees:
I bet the ungzip code in Python is eating up a ton of memory

KuSh:
@jeffdefacto could you also fix the json file :
- coverage["ISO 3166"].alpha2 should be "FR" and not "FR-01" as it covers more than the "Ain" metropolitan department
- license.url using "https://www.etalab.gouv.fr/licence-ouverte-open-licence/" avoids a http redirect

@iandees if needed the csv is also available per subdivisions (~100 separate files). The global file is also currently overlaping some french overseas also available as ISO 3166-2 codes I don't know if that's a problem

I can work on using splitted files if necessary. Or see if I can help on the gunzip code part if you can point me to the problematic repo/code (could it be [that](https://github.com/openaddresses/batch-machine/blob/6aa07557b767f7b524004150bab7471c9ae9cc6a/openaddr/conform.py#L232) ?)

KuSh:
The file is processed with my fix. Ungzipping takes less than 4 seconds

But the CSV contains errors, most notably it is missing headers (which is not the case of split ones). I can adapt the source description but I don't know if it is temporary. How does oa cope with a header line that is not always there ?

It also contains some invalid lines. I'll have a look at it. Perhaps @jdesboeufs can help ?

KuSh:
The source should be fixed now. I've got a feedback from adresse.data.gouv.fr and the batch-machine has been updated to handle big gzipped files.

I don't know how to retrigger checks unfortunately.

@jeffdefacto if you could fix the 2 little problems I have reported in a previous comment, that would be awesome.

jeffdefacto:
@KuSh I just fixed those issues so hopefully it will be good to go now

KuSh:
I tested locally with openaddr-process-one in master batch-machine and it works. I don't know why it doesn't pass checks
Without resources constraint it takes around 25 minutes on my machine

nick-rv:
Hi @jeffdefacto @KuSh , thanks very much for your help.
I checked the following site and the update batch for french addresses still seems to be failing : https://batch.openaddresses.io/job/578500 
Do you see some more precise error ?
Do you have an idea on how to fix it ?
Thanks!

KuSh:
> Hi @jeffdefacto @KuSh , thanks very much for your help. I checked the following site and the update batch for french addresses still seems to be failing : https://batch.openaddresses.io/job/578500 Do you see some more precise error ? Do you have an idea on how to fix it ? Thanks!

I dont know as everything is working locally.
You should preferably open an issue to track the problem as this is a closed PR without much visibility.

jdesboeufs:
Sorry guys I don't work anymore on this.

PR #7575, opened 2024-11-22
Update schema to allow gzip

jeffdefacto:
I think this is needed still for gzip functionality


iandees:
See also https://github.com/openaddresses/batch-machine/pull/58

iandees:
I likely need to kick off another release, which I'll do in a little bit.

iandees:
This should be deployed.

PR #7581, opened 2024-12-02
Liechtenstein address refresh

jeffdefacto:
Refreshes data and now points to URL that should remain up to date.


iandees:
That's not very many addresses. I guess the country is pretty small...

PR #7588, opened 2025-01-08
Updated Canada Addresses

jeffdefacto:
Updated to December 2024 data which adds another 600k or so addresses.


pheersink:
Updated addresses now available from Statistics Canada **(July 2025)**: https://www150.statcan.gc.ca/n1/pub/46-26-0002/462600022022001-eng.htm. Please update.

PR #7603, opened 2025-01-24
Update Hillsborough county

jeffdefacto:
Fix license information and add city and region columns


iandees:
I think the `constant` function might be breaking some filtering somewhere here. I'm going to try it without the constant.

PR #7611, opened 2025-02-05
Add Overseas France

KuSh:
Those, for some, also have dedicated ISO 3166-1 alpha-2 code. Not sure how to cope with that.
Should I duplicate sources ?


iandees:
I'm not sure what you mean by your question, but we definitely don't want to duplicate sources.

KuSh:
> I'm not sure what you mean by your question, but we definitely don't want to duplicate sources.

See https://en.wikipedia.org/wiki/ISO_3166-2:FR#Subdivisions_included_in_ISO_3166-1
Some overseas France subdivisions are also available as main ISO 3166-1 entries with dedicated country codes. I don't know how you expose your data but it is perhaps better to expose FR-971 as GP and so on.

Duplicated sources are certainly not wanted but aliases could potentially be interesting

PR #7635, opened 2025-02-11
add au/wa buildings

andrewharvey:


andrewharvey:
upstream supplies a bulk download instead of the ESRI service but you need to embed a token with the request

andrewharvey:
while the outlines are "coarse" I think they should still be included, OA is just an index/aggregator, downstream consumers will still need to decide what they decide to use.

Issue #7606, opened 2025-01-31
Incomplete Massachusetts jobs


jeffdefacto:
The Massachusetts statewide address job seems to be failing to fully download the source dataset in some recent jobs. I'm not sure if its an issue with the MassGIS feed itself or if the problem lies on the batch machine side, but the jobs are still getting marked as a successfully completed so end up in the global collection. 

Here are some examples I found:

https://batch.openaddresses.io/job/526049
https://batch.openaddresses.io/job/505317

This is the most extreme with only 17k addresses making it in.
https://batch.openaddresses.io/job/530207

Any ideas on what is going wrong and/or how these can be flagged as failures?

PR #7636, opened 2025-02-12
Update Massachusetts source

jeffdefacto:
The web service is failing to fully download. This changes it to a geojson link instead.

### DIFF
--- a/sources/us/ma/statewide.json
+++ b/sources/us/ma/statewide.json
@@ -12,14 +12,14 @@
         "addresses": [
             {
                 "name": "state",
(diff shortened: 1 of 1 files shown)


ilissablech:
@jeffdefacto what's the eta for the updated set being available to download from the OA server?

iandees:
What was failing about the service download? We should be doing the same thing that the ArcGIS Hub is doing.

jeffdefacto:
@iandees Its not clear to me why its happening, but I posted some examples in an issue [here](https://github.com/openaddresses/openaddresses/issues/7606).

jeffdefacto:
@ilissablech It should be in the next global collection run this weekend.

PR #7659, opened 2025-03-03
Fix city in DE/SN

StephanGeorg:
City field was mapped incorrectly and changed to postonm.

see https://batch.openaddresses.io/data/275/history#map=0/0/0


StephanGeorg:
@iandees

Issue #7721, opened 2025-04-08
Add potentially a lot of Illinois counties


bertday:
A kind person rounded up official web maps for every Illinois county here:

https://www.illinoisgis.com/

I haven't vetted all these yet, but a glance it looks like there are quite a few counties we could add (excludes subscription-based and Regrid sources):

- [ ] Clay
- [ ] Clinton
- [ ] Crawford
- [ ] Edgar
- [ ] Edwards
- [ ] Fayette
- [ ] Ford
- [ ] Fulton
- [ ] Gallatin
- [ ] Grundy
- [ ] Hamilton
- [ ] Hancock
- [ ] Henderson
- [ ] Iroquois
- [ ] Jasper
- [ ] Jefferson
- [ ] Jo Daviess
- [ ] LaSalle
- [ ] Madison
- [ ] Marion
- [ ] Marshall
- [ ] Mason
- [ ] Monroe
- [ ] Montgomery
- [ ] Ogle
- [ ] Perry
- [ ] Pike
- [ ] Richland
- [ ] Rock Island
- [ ] Sangamon
- [ ] Scott
- [ ] Wayne

bertday:
## Clay

Has single-line addresses with no separators (e.g. DANDELION RD CLAY CITY IL 62824). In my experience these can be tricky-to-intractable to parse, but I will look around in the repo for any prior art. Or if someone has a way please let me know 😄

bertday:
## Clinton

Similar to Clay County above, these have no line separator, e.g. 315 S PLUM ST BREESE, IL 62230. I'm not sure how to handle those without NLP 🥴

PR #7722, opened 2025-04-08
Add Edgar County, IL addresses and parcels

bertday:
Relates to #7721


sources/us/il/edgar.json
```diff
@@ -6,8 +6,7 @@
             "state": "Illinois"
         },
         "country": "us",
-        "state": "il",
-        "county": "👋TODO👋"
```
bertday: Thanks @iandees, I need to remember to fill in my own template for new sources! 😂

iandees: 😄 no problem! I was doing that to see if there was some weird encoding problem causing the build failure.

iandees:
Well now why isn't data-pls running? 🤔

iandees:
(I need to get to my home laptop to look at this further. I'll be able to do that in an hour or two)

bertday:
No problem or rush @iandees, thanks for looking into it!

iandees:
Yea no kidding.

I think one of the regexes in the addresses are failing and it's not finding a match (returning a `None`).

bertday:
Sorry these turned into such a hassle 😔 Is there a way to default to an empty string if there's no regex match, or something like that?

iandees:
I think the changes I made over in https://github.com/openaddresses/batch-machine/pull/78 helped a bit here. Can you try to modify the regexp to ignore the "NO ADDRESS ASSIGNED" and capture the full zip code instead of just the last 4 digits?

bertday:
Thanks for the ping @iandees, very cool to see the addresses coming through! 🎉 

I will take a look at the regex later today. I may need to understand a bit more about how the `regexp` function works in the context of OA. I've tested the expression out (at [Pythex](https://pythex.org/)) and it captures the full ZIP code for me, but maybe the `regexp` function in batch-machine is only using the last capture group? For reference this is what it looks like currently:

```
"postcode": {
    "function": "regexp",
    "field": "PropertyAddress2",
    "pattern": "\\d{5}(-\\d{4})?$"
  }
```

iandees:
Capture groups are specified using parentheses. So in the pattern you have there, you're only going to capture the dash and four digits at the end of the string. You want to put the parens around the `\d{5}` too to include that piece of the string. Then you can put a `?:` at the beginning of the `-\d{4}` capture group to make it a non-capture group so that the `?` makes it optional.

Try `^(\\d{5}(?:-\\d{4})?)$`

iandees:
Sorry, I couldn't resist playing with a regex. Thanks for adding this source!

bertday:
No worries, glad to see it's fully added! Thanks @iandees

PR #7723, opened 2025-04-09
Add Edwards County, IL addresses and parcels

bertday:
Relates to #7721

### DIFF
--- a/sources/us/il/edwards.json
+++ b/sources/us/il/edwards.json
@@ -1,0 +1,59 @@
+{
+  "coverage": {
+      "US Census": {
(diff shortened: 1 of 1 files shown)


bertday:
Hi again @iandees, I was curious to see if https://github.com/openaddresses/batch-machine/pull/74 gets these addresses to load properly. Would it be possible re-run the [job](https://batch.openaddresses.io/job/574843)? I tried doing it myself but I don't think I have the permissions.

iandees:
Same thing on this one. Looks like the address regexp is failing to match in some cases.

Issue #7730, opened 2025-04-16
Update us/il/cook addresses and add parcels


bertday:
- Addresses
  - City is using wrong field, should be `Post_Comm`
  - Use `State` for region
- Add [parcels](https://gis.cookcountyil.gov/traditional/rest/services/CookViewer3Parcels/MapServer/0)

PR #7731, opened 2025-04-16
Update us/il/cook addresses and add parcels

bertday:
See #7730


iandees:
It looks like there might be some kind of blocking happening here to prevent repeated requests to this service. Or we're overwhelming their database.

bertday:
Ah, bummer 😕 I think there's a static parcel dataset from 2021, I can try that next.

iandees:
Looks like that helped with the addresses, but parcels are still breaking. I'll merge for now, thanks all.

PR #7739, opened 2025-04-21
Fix Spain SRS

jeffdefacto:
Fixes a projection error due to SRS typo

### DIFF
--- a/sources/es/countrywide.json
+++ b/sources/es/countrywide.json
@@ -40,7 +40,7 @@
                     "postcode": "codigo_postal",
                     "city": "poblacion",
                     "format": "csv",
-                    "srs": "EPSG:4259"
+                    "srs": "EPSG:4258"
                 }
             }
         ]


orangejulius:
Question: does this change just impact how OA displays the preview map/renders coverage? Because I don't think we saw points being offset in Spain, but just want to make sure.

iandees:
We reproject everything to 4326 as part of the download pipeline, so this would have affected output data as well as the map display here.

orangejulius:
Ah yes indeed. I visualized all the addresses in Spain and while they look correct for this city block in Barcelona...
image


over on the other side of the country in A Coruña they are clearly shifted a few meters. Very nice catch!
image

jeffdefacto:
Yeah, it was a subtle enough offset that I didn't notice during the initial PR. I only stumbled upon it while investigating an unrelated issue in the Overture release.

Issue #5531, opened 2021-03-05
Belgium: parcels


barthanssens:
Hi,

I have a few questions on providing (Belgian) parcel data:

- Is this a supported feature or still in beta ?
The documentation (https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md) still mentions `Valid layers are addressess. In the near future parcels and buildings will be supported`
But there are already a few parcel sources available, so I assume we can just mimic what e.g. Canada is doing ?(https://github.com/openaddresses/openaddresses/blob/master/sources/ca/bc/port_moody.json)

- I assume the structure/source does not have to follow the structure of the addresses (ie.: the Belgian addresses are provided by the 3 Regions, but the parcel are published by the federal level as 1 single nation-wide source)

- Is there a size limit per source ? The compressed download is a 2.5 GB zip with Shapefiles (8 GB uncompressed)

andrewharvey:
> Is this a supported feature or still in beta ?

@ingalls has built support for parcels so you can add them now. Documentation still needs to be updated, feel free to put forward a PR for the docs if you'd like to help.

> (ie.: the Belgian addresses are provided by the 3 Regions, but the parcel are published by the federal level as 1 single nation-wide source)

You can have a separate `countrywide.json` file just containing the parcels, it doesn't need to be in the same json file for each region which contains the addresses.

> Is there a size limit per source ? The compressed download is a 2.5 GB zip with Shapefiles (8 GB uncompressed)

You can try. I uploaded a 1GB zip for au/qld parcels.

andrewharvey:
drafting updated docs at #5542

barthanssens:
Thanks, I'll give it a try early next week.

barthanssens:
Looks like the shapefile containing the Belgian parcels also contains other layers (e.g. administrative borders), so I'll have to ask them to make a separate set containing just the parcel data...

https://eservices.minfin.fgov.be/myminfin-web/pages/cadastral-plans

iandees:
The system can download the whole thing and pick just the parcels. It'll take less time if they are broken up, but it should be possible.

barthanssens:
Ah that would be great... but if I read the documentation correctly, the `layer` property is currenly only supported for GDB, not for shapefile ?

iandees:
You can accomplish the same thing by specifying `file` when processing Shapefiles.

barthanssens:
Makes sense, thanks.

[...]

PR #7760, opened 2025-05-28
GH-5531: added parcels for 3 Regions + updated contact address

barthanssens:
Added the parcel info on the "fr" files of the 3 Belgian Regions (not in nl, otherwise the same shapefiles get downloaded and processed twice...). Unfortunately there is no "latest version" link, so these shapefiles won't get updated, one has to manually change the links to newer files to get the new data... 

Also used a slightly different email and physical address for FPS Policy and Support for the addresses


barthanssens:
Seems like the OA download doesn't like > 500 MB files (tried storing the data on another server, but doesn't seem to make much of a difference): 
`urllib3.exceptions.ProtocolError: ('Connection broken: IncompleteRead...`

PR #7773, opened 2025-06-13
Fix zipcodes on Seminole, FL and Clarke, GA

jeffdefacto:
These zipcodes had a trailing `.0` for some reason. The regex removes it.


iandees:
The `.0` is happening because the Esri data source lists the zipcodes as numbers, and the default rendering of a number is decimal.

jeffdefacto:
Ah yeah that makes sense. I wonder if it would be better to just cast decimal columns as integer or string as part of the processing.

PR #7776, opened 2025-06-18
Luxembourg: latest dataset

grischard:
Luxembourg: https://data.public.lu/fr/datasets/r/7b58cf20-cbb0-4970-83f7-53a277f691b8 redirects to the latest version of the geojson, updated weekly.


jeffdefacto:
Batch doesn't seem to like redirecting to the geojson but works for the shapefile. Try this:

[countrywide.json](https://github.com/user-attachments/files/20822602/countrywide.json)

Issue #7781, opened 2025-07-07
GeoJSON coordinates are not normalized correctly for some files.


davideder:
There are a number of files with coordinates that need scaled, are backwards, or should be negative, but are positive.

- [ ] us/md/howard-buildings-county.geojson
- [ ] us/ms/city_of_diberville-parcels-city.geojson
- [ ] us/oh/darke-addresses-county.geojson
- [ ] us/oh/miami-addresses-county.geojson
- [ ] us/tx/kerr-addresses-county.geojson
- [ ] us/tx/mclennan-addresses-county.geojson
- [ ] us/tx/midland-addresses-county.geojson
- [x] us/va/city_of_norfolk-addresses-city.geojson
- [x] us/wy/statewide-addresses-state.geojson
- [x] nz/countrywide-buildings-country.geojson

The coordinates from us/oh/darke-addresses-county.geojson seem to need more than just scaling.  They just seem to be in the wrong place.

davideder:
openaddresses/nz/countrywide-buildings-country.geojson seems to be EPSG:4167.
openaddresses/us/md/howard-buildings-county.geojson seems to be EPSG:2248.
openaddresses/us/ms/city_of_diberville-parcels-city.geojson seems to be EPSG:2254.
openaddresses/us/oh/darke-addresses-county.geojson seems to be SR-ORG:6686.
openaddresses/us/oh/miami-addresses-county.geojson seems to be EPSG:3735.
openaddresses/us/tx/kerr-addresses-county.geojson seems to be reversed.
openaddresses/us/tx/mclennan-addresses-county.geojson seems to be reversed.
openaddresses/us/tx/midland-addresses-county.geojson seems to be reversed.
openaddresses/us/va/city_of_norfolk-addresses-city.geojson seems to be SR-ORG:6639.
openaddresses/us/wy/statewide-addresses-state.geojson seems to be reversed.

davideder:
openaddresses/nz/countrywide-buildings-country.geojson is actually EPSG:2193 according to https://data.linz.govt.nz/layer/101290-nz-building-outlines/

iandees:
Checking in on the currently-unchecked items — 4 of the 7 are now moot because the files themselves are gone (superseded by other sources), independent of this issue:

[...]

PR #7783, opened 2025-07-07
Update Ohio sources

arch0345:
Replace LBRS downloads for each county with new statewide LBRS address layer

Also updated url for Miami County

Re: #7781


iandees:
Merged manually to master as commit 76ea38e78. Thanks for the contribution!

PR #7788, opened 2025-07-09
Ask LINZ for building GeoJSON in 4326

iandees:
For #7781


iandees:
I think our GeoJSON code might ignore SRS on GeoJSON sources because GeoJSON is supposed to always be 4326.

This PR asks the source URL to do the transform for us.

Issue #7071, opened 2024-02-12
Official addresses for Austria


StephanGeorg:
The provided source for Austria is outdated and not available anymore.

There is a new source available:

+ WEBSITE: https://geometadatensuche.inspire.gv.at/metadatensuche/srv/ger/catalog.search#/metadata/bd28c2b0-4ea5-41c2-8847-b7e8711a6512
+ DATA: https://data.bev.gv.at/download/Adressregister/AT-INSPIRE_AD_Address_CSV_20231022.zip
+ LICENSE: [CC BY 4.0 DEED](https://creativecommons.org/licenses/by/4.0/)
+ ATTRIBUTION: **© Österreichisches Adressregister, Stichtagsdaten vom TT.MM.JJJJ**

Address information is distributed across several CSV files. I used [duckdb](https://duckdb.org/) to merge the data and transform the geometries:

```sql
LOAD spatial;

-- source https://geometadatensuche.inspire.gv.at/metadatensuche/srv/ger/catalog.search#/search?resultType=details&sortBy=relevance&any=Adresse%20Relationale%20Tabellen&fast=index&_content_type=json&from=1&to=20

CREATE temp TABLE addresses AS FROM (
  SELECT 
    a.ADRCD
    , a.HNR_ADR_ZUSAMMEN
    , a.HAUSNRTEXT
    , a.HAUSNRZAHL1
    , a.HAUSNRBUCHSTABE1
    , a.HAUSNRVERBINDUNG1
    , a.HAUSNRZAHL2
    , a.HAUSNRBUCHSTABE2
    , a.HAUSNRBEREICH
    , a.HOFNAME
    , s.STRASSENNAME
    , s.STRASSENNAMENZUSATZ
    , s.SZUSADRBEST
    , s.ZUSTELLORT
    , o.ORTSNAME
    , g.GEMEINDENAME
    , a.PLZ::varchar as PLZ
    , a.EPSG
    , ST_Point(a.HW::double, a.RW::double) AS "geometry"
    , 0.0::double AS "lat"
    , 0.0::double AS "lon"
  FROM 
    'ADRESSE.csv' a
    LEFT JOIN  'STRASSE.csv' s ON a.SKZ = s.SKZ
    LEFT JOIN 'ORTSCHAFT.csv' o ON a.OKZ = o.OKZ
    LEFT JOIN 'GEMEINDE.csv' g ON a.GKZ = g.GKZ
);

-- transform coordinates
[...]

PR #7789, opened 2025-07-09
Update Austria countrywide

jeffdefacto:
Update to April 2025 data. I also deleted the old obsolete scripts and replaced it with query from https://github.com/openaddresses/openaddresses/issues/7071


iandees:
Pretty.

Issue #7813, opened 2025-08-13
No `streets` data in countrywide Norway data


Kuzat:
Looking at the `no/countrywide` data it seems like all the `street` fields look empty.
It looks like it used to work previously based on the History graph.

<img width="1265" height="322" alt="Image" src="https://github.com/user-attachments/assets/a21d621b-8dd2-47b3-a10f-8942491a511a" />

Kuzat:
Looking at the [source file](https://github.com/openaddresses/openaddresses/blob/master/sources/no/countrywide.json) and from reviewing the corresponding csv file it looks to me that the field name for `street` is `adressenavn` and not `addressenavn`.

PR #7814, opened 2025-08-13
fix typo in street field name for Norway countrywide

Kuzat:
This should hopefully fix the problem in #7813 with no `street` data for Norway's countrywide data.


iandees:
There are still a bunch of records with blank house number and street names, but I see some with the data present. Thanks for the fix!

PR #7824, opened 2025-09-14
Fix Aiken source layer

### LINKED ISSUE #7823: Incorrect Aiken, SC source
We noticed an issue with the data for Aiken, SC (nothing has addresses info!). Upon further inspection of the source, it looks like layer ID (1) is for Abbeville, not Aiken. I verified the data in QGIS and can confirm that Aiken is actually layer 2, and that the Abbeville data doesn't seem to include addresses at least in a spot check (just parcels).

I'll try to get a Docker batch environment up and submit a patch next week after confirming the fix.

ianthetechie:
Fixes #7823

### DIFF
--- a/sources/us/sc/aiken.json
+++ b/sources/us/sc/aiken.json
@@ -14,7 +14,7 @@
         "addresses": [
             {
                 "name": "county",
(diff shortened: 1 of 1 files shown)


iandees:
Looks good!

PR #7827, opened 2025-09-25
Filter out cadastral addresses de/bw/statewide

thismakessand:
About 8% of addresses in the de/be/statewide file are cadastral addresses with only cadastral ids as housenumber.

### DIFF
--- a/scripts/de/bw/statewide.py
+++ b/scripts/de/bw/statewide.py
@@ -1,0 +1,28 @@
+"""
+Filter out addresses with cadastral psuedonumbers (col 3 == 'C'). These are ids/pseudo-numbers and not house numbers.
+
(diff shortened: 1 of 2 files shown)


thismakessand:
I guess there's nothing that actually runs the scripts automatically.  Just trying to filter out some rows based on a value in col3 but can't see a way to do that.

iandees:
Yep, the scripts have to be manually run at this point.

iandees:
... and yea, we don't have a way to filter data at the moment.

If you want to try your hand at porting this over to an alltheplaces scraper (e.g. https://github.com/alltheplaces/alltheplaces/blob/master/locations/spiders/addresses/us/mn/us_mn_hennepin_addresses_us.py) you can do more complex things there because it's actually running Python code.

PR #7854, opened 2025-10-29
migrate nz/countrywide to pilot dataset

### LINKED ISSUE #7857: LINZ is planning on changing the NZ Addresses schema
This is expected to be done in March 2026, however a pilot dataset is available currently.

Proposed changes at #7854

andrewharvey:
~~**DRAFT PR do not merge yet** awaiting for this new dataset to become official on 4th March 2026.~~

LINZ have just announced changes to their NZ Addresses dataset to be published in March 2026, see https://data.linz.govt.nz/document/25898-nz-addresses-data-dictionary-pilot/

Changes include
`unit` field renamed
image

address number prefix removed
image

road name fields changed, however the `full_road_name` field we are using is unchanged

coordinate fields removed
image

the current URL will output coordinates as WKT `POINT (-46.2988688167 168.2442077)` which it doesn't appear that OA supports... therefore I changed the output to GeoJSON.

This PR applies the changes and migrates to the new dataset. We can choose to go live (merge) this now or wait for the Pilot dataset to finalise and be released due in March 2026.

cc @dwsilk

closes #7857

### DIFF
--- a/sources/nz/countrywide.json
+++ b/sources/nz/countrywide.json
@@ -11,7 +11,7 @@
         "addresses": [
             {
                 "name": "country",
-                "data": "https://data.linz.govt.nz/services;key=441c0097257f45ab84d5ea4d24723f23/wfs?service=WFS&request=GetFeature&outputformat=CSV&typeNames=layer-123113",
+                "data": "https://data.linz.govt.nz/services;key=441c0097257f45ab84d5ea4d24723f23/wfs?service=WFS&request=GetFeature&outputformat=JSON&typeNames=layer-123113",
                 "website": "https://data.linz.govt.nz/layer/123113-nz-addresses-pilot/",
                 "license": {
                     "url": "https://creativecommons.org/licenses/by/4.0/",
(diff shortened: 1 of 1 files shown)


andrewharvey:
we may face the same issue as #7788 in that we expect GeoJSON to be in EPSG:4326.

KuSh:
> we may face the same issue as #7788 in that we expect GeoJSON to be in EPSG:4326.

You can specify source EPSG to convert from with the srs attribute.

andrewharvey:
> > we may face the same issue as #7788 in that we expect GeoJSON to be in EPSG:4326.
> 
> You can specify source EPSG to convert from with the srs attribute.

We still have `"srs": "EPSG:4167"` specified, so sounds like there shouldn't be any issue.

andrewharvey:
I think this is ready to go now, but probably best to still wait until 4th March 2026 before merging.

andrewharvey:
From LINZ's latest update https://mailchi.mp/linz.govt.nz/nz-address-data-changes-feb06

> The new layers are now available via the LINZ Data Service and you can transition to these layers today. The only change on 4 March will be to remove pilot from the name. The layer IDs and schema will not change. A new ArcGIS REST service will be published on 4 March 2026. 

Based on this, I'll merge this now.

PR #7856, opened 2025-10-29
Create Yuma County, Colorado, US

skylarmt:


skylarmt:
If anyone can say why the regex chains break the batch job, that would be great.

TSC_Site_Address looks like this: "13705 COUNTY ROAD O KIRK 80824"

* House number is first (13705)
* Street is second (COUNTY ROAD O)
* All town names in the county are a single word (in this example, "KIRK"), so that makes extracting them possible.
* ZIP Code follows the town name.

iandees:
For "legacy reasons" the fields outputted from each function get prefixed with `oa:`. I think this was so that the output of the conform step didn't interfere with existing columns. This means that the second "chain" function needs to operate on the prefixed column names.

PR #7897, opened 2025-11-11
South Carolina, US statewide

### LINKED ISSUE #7896: South Carolina Addresses!
Image
https://mega.nz/file/bk501YgY#D6OBxTp3brjrKjBXcR-AOh3Llh1ltTcCRhTmHvaOAHg

jeffdefacto:
Fixes #7896

### DIFF
--- a/sources/us/sc/statewide.json
+++ b/sources/us/sc/statewide.json
@@ -13,9 +13,6 @@
             {
                 "name": "state",
                 "data": "https://data.openaddresses.io/cache/uploads/jeffdefacto/2025-11-11-05ekg/SC.zip",
-                "license": {
-                  "text": "Please note that such records used for commercial purposes is prohibited by SC Code Ann. ß 30-2-50(B)"
-                },
                 "compression": "zip",
                 "protocol": "http",
                 "conform": {


jeffdefacto:
The parcels seem to have a geometry type that batch doesn't like. Pulling them out of this PR.

`shapely.errors.GEOSException: ParseException: Unknown type: 'MULTISURFACE'`

skylarmt:
I noticed your note about the South Carolina law restricting commercial use.  While it probably applies to the *input* data, the *output* from OpenAddresses isn't covered because there's no personally identifying data.

From the law cited in the note:

> A person or private entity shall not knowingly obtain or use personal information obtained from a state agency, a local government, or other political subdivision of the State for commercial solicitation directed to any person in this State.

> "Personal information" means information that identifies or describes an individual

While the definition of "personal information" goes on to list many things as personal information, including "home address", the intent of the law seems crystal clear that by itself an address isn't covered.  All the other information listed as "personal" is either biometric data, a unique ID number (SSN, drivers license, etc), or would only be identifying when combined with a person's name (such as education level, which could be represented by a number between 0 and 16 for "none" to "4 years of college" and by itself is meaningless).  The fact that the law says "**home** address" is also important, because there's no way to tell from the data if an address is even residential at all, let alone if someone calls it a home.

tl;dr the note is misleading I think, and should probably be removed, as it places incorrect license terms (no commercial use) on the dataset when there is no such limitation in the law.

jeffdefacto:
@skylarmt I do agree that commercial use has a particular meaning in this case that can be misleading and probably only really applies to the raw parcel data. I'll pull it out of the json, but for posterity this is the terms of use that come with the datasets.


`Please note that such records used for commercial purposes is prohibited by SC Code Ann. ß 30-2-50(B):
Each state agency, local government, and political subdivision of the State shall provide a notice to all requestors of records pursuant to this chapter and to all persons who obtain records pursuant to this chapter that obtaining or using public records for commercial solicitation directed to any person in this State is prohibited.`

iandees:
Thanks for digging into this, all.

PR #7915, opened 2025-11-22
Rework France sources

KuSh:
- Split source per state, this should help jobs
- Add parcels and buildings for mainland and most overseas
- Map accuracy field using support introduced by https://github.com/openaddresses/batch-machine/pull/82
- Fix license attributes and links


KuSh:
I've finally decided to split France sources per state as the countrywide one seems to pose a problem (even if locally it works)
It will also avoid overlaps between global source and overseas that was happening already

KuSh:
@iandees I've created an other [PR](https://github.com/openaddresses/openaddresses/pull/7922) to fix schema definitions as it seems data-pls don't use those provided in the PR. I'll rebase when merged

PR #7916, opened 2025-11-23
update dataset and enrich with Addressforall postcodes

crebollobr:
- Update geographic data based on latest INEGI source.
- Add new 'postcode' column.
- Postcode data provided by Addressforall project.


iandees:
Thanks for the update!

Issue #7910, opened 2025-11-17
support accuracy at the address attribute tag level


andrewharvey:
Accuracy is provided as a processing tag at the source level, https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md#processing-tags, however some sources like `au/countrywide` this will vary per address within the source.

It would be desirable to extend this to an "Address Attribute Tag" also called `accuracy` following the same format as the source level processing tag.

KuSh:
Same for France dataset

Should be computable with some regexp attribute functions

iandees:
I thought we supported this already. There was discussion https://github.com/openaddresses/batch-machine/issues/72 about it. Looking closer at the code, though, I don't think we actually do.

I'll see if I can add it in there this coming week.

iandees:
I believe this should be supported now. Does someone want to give it a try?

KuSh:
> I believe this should be supported now. Does someone want to give it a try?

I can give a try in my current PR to update France sources if you want

KuSh:
https://github.com/openaddresses/openaddresses/pull/7915

andrewharvey:
Thanks @iandees . I tried in #7923, but I'm not sure what the problem is. In the source `conform` I set `"accuracy": "accuracy"` to directly map the `accuracy` field from the source CSV as this new per record `accuracy` field.

An integer would only be expected here when setting `accuracy` at the per source level... hang on, how do you specify a record level accuracy field now with https://github.com/openaddresses/batch-machine/pull/82, because `conform: { accuracy: ... }` is still for the source level according to (https://github.com/openaddresses/openaddresses/blob/master/CONTRIBUTING.md)?

```
ok - updating job: 716752 with {"status":"Running","version":"1.1.0","output":{"cache":false,"output":false,"preview":false,"validated":false},"loglink":"batch-prod-job/default/c011bdbeb4d545bc8790ce5c8a8e3923"}
[{"instancePath":"/layers/addresses/0/conform/accuracy","schemaPath":"#/properties/layers/properties/addresses/items/properties/conform/properties/accuracy/type","keyword":"type","params":{"type":"integer"},"message":"must be integer"}]
Error: Source does not conform to V2 schema
    at Job.fetch (file:///usr/local/src/batch/lib/job.js:87:19)
    at async flow (file:///usr/local/src/batch/task.js:82:9)
ok - updating job: 716752 with {"status":"Fail"}
false
```

[...]

PR #7923, opened 2025-11-28
support per record accuracy in au/countrywide

andrewharvey:
see also #7910


andrewharvey:
Looks good!

This is the mapping we are using

```
        CASE geocode_type WHEN 'BUILDING ACCESS POINT'         THEN 1 -- rooftop
                          WHEN 'BUILDING CENTROID'             THEN 1 -- rooftop
                          WHEN 'BUILDING CENTROID MANUAL'      THEN 1 -- rooftop
                          WHEN 'CENTRE-LINE DROPPED FRONTAGE'  THEN 2 -- interpolation
                          WHEN 'DRIVEWAY FRONTAGE'             THEN 3 -- driveway
                          WHEN 'FRONT DOOR ACCESS'             THEN 1 -- rooftop
                          WHEN 'FRONTAGE CENTRE'               THEN 2 -- interpolation
                          WHEN 'FRONTAGE CENTRE SETBACK'       THEN 2 -- interpolation
                          WHEN 'PROPERTY ACCESS POINT'         THEN 3 -- driveway
                          WHEN 'PROPERTY ACCESS POINT SETBACK' THEN 3 -- driveway
                          WHEN 'PROPERTY CENTROID'             THEN 2 -- parcel
                          WHEN 'PROPERTY CENTROID MANUAL'      THEN 2 -- parcel
                          WHEN 'UNIT CENTROID'                 THEN 2 -- parcel
                          WHEN 'UNIT CENTROID MANUAL'          THEN 2 -- parcel
                          WHEN 'GAP GEOCODE'                   THEN 4 -- interpolation
                          WHEN 'STREET LOCALITY'               THEN 4 -- interpolation
                          WHEN 'LOCALITY'                      THEN 4 -- interpolation
                                                               ELSE 5 -- unknown
```

https://github.com/openaddresses/openaddresses/blob/master/scripts/au/gnaf.sh#L86-L103

the distribution is

| type | % |
| -- | -- |
| 1 rooftop | 5% |
| 2 parcel | 90% |
| 3 driveway | 2% |
| 4 interpolation | 3% |
| 5 unknown | 0.4% |

iandees:
In theory you could use the newly-added `map` function in OpenAddresses to do that, but since you already have the mapping and we already have to do the external processing it doesn't matter.

andrewharvey:
> In theory you could use the newly-added map function in OpenAddresses to do that, but since you already have the mapping and we already have to do the external processing it doesn't matter.

I agree, doing the map within the confirm would be better, so that the source data OA consumes is closer to as-published by the source, and `conform` does all the mapping into OA's format.

PR #7934, opened 2025-12-02
Update Mexico countrywide

jeffdefacto:
Newly updated file from Address4all adds postcode data and fills in region and city gaps.


jeffdefacto:
Oops. This is already done.

PR #7939, opened 2025-12-03
document the map attribute function

andrewharvey:
documents the map attribute function added in https://github.com/openaddresses/batch-machine/pull/82

cc @iandees


iandees:
Thanks!

PR #7943, opened 2025-12-04
fix au/qld/townsville

andrewharvey:


andrewharvey:
Thanks for looking at this @iandees I have no idea why it's giving a schema error, other sources like https://github.com/openaddresses/openaddresses/blob/d885ab599c07fb8379cafde2725a580608d03c5f/sources/us/pa/elk.json#L42 do it as a literal number....

iandees:
I think the Elk county situation works because the source is a CSV, and you can get the 5th column of a CSV. When building a request for an ESRI service we build a list of columns we want to fetch from the server using the conform dict, and that step assumes all values of the conform dict are string column names or functions that result in column names.

This means the schema and ESRI code are at odds and need to be updated:
1. Schema should have number removed as an option for the value of accuracy, and functions added
2. batch-machine's esri downloader should check to make sure all values are string before sorting to get column names for the request

andrewharvey:
> Schema should have number removed as an option for the value of accuracy, and functions added

It would be nice if we could still accept the literal number as equivalent to the constant function, that would make it simpler for all those sources where the whole dataset is of the same accuracy and retains backwards compatibility for all the existing sources.

iandees:
Yea I agree. I ended up doing that in https://github.com/openaddresses/batch-machine/pull/84.

PR #7949, opened 2025-12-04
🇳🇨 fix nc/countrywide + add parcels

andrewharvey:


andrewharvey:
I just realised the addresses are included in [fr/nc/statewide](https://github.com/openaddresses/openaddresses/blob/master/sources/fr/nc/statewide.json) by @KuSh with quite different fields.

Anyone have any thoughts on if these should be fr/nc/statewide or nc/countrywide?

I'm happy to update this PR either way, but I don't have any local knowledge

andrewharvey:
cc @jdesboeufs

andrewharvey:
I just saw #7915 which mentions also including parcels for French Overseas Territories.

jdesboeufs:
Hi,

#7915 use a data source which doesn't cover nc parcels.
"DGFiP" only covers mainland and some overseas (971, 972, 973, 974, 976, 977, 978).
977/978 are inside 971 dataset (historical reason).

KuSh:
I've added NC in the FR part mainly because it was the same source for all data (DGFiP). I wasn't aware of another official source.

If data.gouv.nc provides all or even some of the necessary data I think we can move it to its own source, that would ease maintenance. Same for other overseas if available!

I'll adapt my PR accordingly if everyone agree.

Also thanks @jdesboeufs for the tip about 977/978 being inside 971 dataset, it wasn't obvious in the doc. I'll see what I can do to make use of that.

jdesboeufs:
For addresses, the French Addresses Database covers mainland + overseas.

For parcels, DGFiP (cadastre.data.gouv.fr) only covers mainland + DROM (971, 972, 973, 974, 976) + Saint-Martin and Saint-Barthelemy (977/978) that were historically attached to 971.
FP and NC have their own data for their territory => official data.

andrewharvey:
Thanks for your comments @KuSh @jdesboeufs .

In that case, shall we proceed with this PR?

KuSh:
> In that case, shall we proceed with this PR?

That seems to be the right choice yup. I'll follow up with a split of mine after that!!

PR #7955, opened 2025-12-04
support building levels in building schema

andrewharvey:


schema/layers/building_conform.json
```diff
@@ -18,6 +18,10 @@
     "height": {
       "description": "height of the building (from the ground) in meters",
       "$ref": "../util/functions/base.json"
+    },
+    "levels": {
```
KuSh: Shouldn't this also be documented in CONTRIBUTING.md?

andrewharvey: yes, at first I was just to explore this concept, to see if it would be supported and accepted, and if it made sense.

If it becomes in use and supported we should include it in CONTRIBUTING.

PR #7962, opened 2025-12-07
Add missing map function in chain

KuSh:


iandees:
What're you using to format the JSON schema files?

KuSh:
> What're you using to format the JSON schema files?

It's autoformatting from zed editor. I think it uses prettier under the hood for JSON.

PR #7971, opened 2025-12-14
Make accuracy schema match existing fields

iandees:
#7910 did this in a way that is inconsistent with other fields


iandees:
Too many changes, so I'm going to fix the Esri downloader instead.

PR #7986, opened 2025-12-17
Fix housenumber for GG source

thismakessand:
New field name


iandees:
The sample doesn't seem to have any housenumbers populated. Is it really sparse or should every row have housenumber?

thismakessand:
It's sparse - mostly building/house names which doesn't appear to be supported.  Looks like previously ~1/3 of records have numbers (latest run was 0 due to column name change which this PR fixes).  I do see one housenumber in the sample.

image

PR #7988, opened 2025-12-22
au LGA sources to include the short LGA name for spatial matching

andrewharvey:
Makes all the AU LGA level sources have consistent use of the `county` field for the LGA shortname.


andrewharvey:
not sure why the CI is trying to check the template source inside `scripts`, since we have https://github.com/openaddresses/openaddresses/blob/d059ca5c939d62a41221a0a81bd5a6ee4bea16fe/ci/run_changed_sources.py#L218 which is meant to be only allowing those files in `sources` to be processed...

iandees:
Root-caused this: the CI failure isn't from a `sources/`-prefix filter missing in this repo (`ci/run_changed_sources.py` already has it correctly). It's a bug in the `openaddresses/batch` service that actually drives the `data-pls` check — `CI.internaldiff()` in `api/lib/ci.js` turns *any* changed/added JSON file with a `layers` key into a job, with no check that it lives under `sources/`. `scripts/au/tas/LIST_template.json` (a template with unresolved `{PLACEHOLDER}` tokens) gets picked up as if it were a real source, its job fails on the unresolvable data URL, and that single failing job flips the whole run/check to `Fail` — masking the ~60 legitimate `sources/au/**` changes in this PR.

Opened a fix: https://github.com/openaddresses/batch/pull/666 (adds a `sources/*.json`-only guard to `internaldiff()`, mirrors the filter already in `ci/run_changed_sources.py`). Not merged/deployed yet.

This PR is blocked on that batch PR landing and being deployed — the `data-pls` check will keep failing here until then, even though the source changes themselves look fine.

Issue #485, opened 2014-10-15
us-in-monroe


jalbertbowden:
Monroe County
not county site, but there's some data here
https://spatialdata.iu.edu/DOQQS/city_county/monroe_co/2005/

PR #8029, opened 2026-01-14
feat: add Norwegian parcel data

Johannes-Andersen:
I am not that familiar with how coverage should be filled out correctly to get the website data to show the coverage correctly.
So I would love some help with getting those correctly filles out


andrewharvey:
> I am not that familiar with how coverage should be filled out correctly to get the website data to show the coverage correctly.

https://github.com/openaddresses/batch/blob/0bd42d14b3f67eeb796f09e51dd25e0e93c82efd/api/lib/types/map.js#L272C17-L272C21

says that if you provide the ISO 3166 alpha2 it will be used for the coverage code, this code then get's matched against the map layers at

https://github.com/openaddresses/batch/blob/63fb25e16969ed6df90116d85fe76399e6f3ee08/api/lib/types/map.js#L16-L19

It looks like the codes you provided match the data files that are being used for the coverage, so that's probably fine. I just don't know if the coverage represents address only.

Johannes-Andersen:
> > I am not that familiar with how coverage should be filled out correctly to get the website data to show the coverage correctly.
> 
> [openaddresses/batch@`0bd42d1`/api/lib/types/map.js#L272C17-L272C21](https://github.com/openaddresses/batch/blob/0bd42d14b3f67eeb796f09e51dd25e0e93c82efd/api/lib/types/map.js#L272C17-L272C21)
> 
> says that if you provide the ISO 3166 alpha2 it will be used for the coverage code, this code then get's matched against the map layers at
> 
> [openaddresses/batch@`63fb25e`/api/lib/types/map.js#L16-L19](https://github.com/openaddresses/batch/blob/63fb25e16969ed6df90116d85fe76399e6f3ee08/api/lib/types/map.js#L16-L19)
> 
> It looks like the codes you provided match the data files that are being used for the coverage, so that's probably fine. I just don't know if the coverage represents address only.

Hummm... do you by any chance see what Denmark does that Norway dosn't?
Denmark is all Green and good with this config: https://github.com/openaddresses/openaddresses/blob/master/sources/dk/countrywide.json
Screenshot 2026-01-16 at 14 48 36

While Norway is all white with this config: https://github.com/openaddresses/openaddresses/blob/master/sources/no/countrywide.json
Screenshot 2026-01-16 at 14 50 15

andrewharvey:
Yeah the country coverage file being used at https://v2.openaddresses.io/country.geojson countains the incorrect country code, it has "-99" rather than "no".

I asked for details on how these files were generated at #485

image

and while it might make it work if you change the country code to -99 let's not do that.

PR #8045, opened 2026-01-23
Updated Seminole Florida

Mbryan1541:
Changed address and parcel URLs


jeffdefacto:
Parcels has a shapely issue `shapely.errors.GEOSException: ParseException: Unknown type: 'MULTISURFACE'` but the url is more correct at least and there is not an alternative.

Merging as is

PR #8053, opened 2026-01-30
Update NL and documentation

jeffdefacto:
Update NL to official BAG geopackage which has a clear CC0 license and is regularly updated.

Also added references to gpkg support in the documentation. 

gpkg was too big to run successfully, so added a script to download and convert to csv.


jeffdefacto:
Keeping as draft until batch-machine changes are pushed

PR #8066, opened 2026-02-06
Amador County CA update url and variables

asomphanith:
Removed instances of regexp, and update urls.


jeffdefacto:
This is pointing to the owner address, not the parcel address.
try `https://services9.arcgis.com/Z65sx307Hunxj6eq/ArcGIS/rest/services/Parcels_AmadorCo_20210531/FeatureServer/0`
instead

PR #8074, opened 2026-02-16
Fix typo in layers section of CONTRIBUTING.md

Cj-Malone:
Tests expect `addresses` not `addressess`.


Cj-Malone:
I actually haven't got `npm test` to pass on master, it seems like `non-integer accuracy value should fail` is failing on null.

```
  2 failed  

  Failed Tests:   There were 2 failures

    x non-integer accuracy value should fail
    x null
```

iandees:
When we moved the "accuracy" field into the conform object in #7910, it had to allow strings for column names to be consistent with other fields in the conform.

iandees:
... so the tests are probably wrong.

PR #8147, opened 2026-04-10
Fix attribution spelling mistake

Cj-Malone:


iandees:
Huh. Interesting that pre-commit didn't catch this as part of jsonschema validation.

PR #8151, opened 2026-04-11
Update waushara.json

starblazr:
address formatting correction to [N/W][Number] Suffix


starblazr:
checked results over, correct formatting.

PR #8153, opened 2026-04-11
Update statewide.json

starblazr:
removed spaces


starblazr:
checked.  fixed the [n/w/e/s][number] for most counties and then fixed milwaukee/waukesha's [n]xxx[w]xxxxxx numbering system also.

PR #8183, opened 2026-05-14
Update door.json

starblazr:
changed field names to match new ones, didnt seem to pull from the webservice


starblazr:
checked data, checked with a resident of door county that has family in emergency management, address structure matches (see other imports for fixing N/S/E/W addresses)

PR #8205, opened 2026-06-06
Process Address Lookup Service (ALS) source for HK

tijptjik:
## Source schema changes

Most consequential changes -- please consider whether these match the project goals and standards:

- added `city` based on `LocationName` in the [ALS schema](https://www.als.gov.hk/docs/Data_Dictionary_for_ALS_EN.pdf)
- fixed the `zh-hant` street to be based on `ChiStreet` instead of `EngStreet` 
- fixed the number to include the range, e.g. `{BuildingNoFrom}-{BuildingNoTo}`
   - In Hong Kong, ranges are more indicative of which building you're actually looking for, so it's IMHO important to include.
- referenced the dedicated [als.gov.hk/](https://www.als.gov.hk/) as the source site

## Codebase changes

- added `README.md` for the extraction script
- used `uv` to track dependencies in `pyproject.toml`
- added `lib/download.py` which automates the archive fetching
- refactored filesystem operations and data processing into `lib/fs.py` and `lib/process.py`
- refactored `hk.py` to only run as a script and print out feedback to user 
- clean the produced key names

## Projectwide changes

- added `__pycache__` to global `.gitignore`
- fixed the openaddresses.io links in docs

## Changes Required

The data-pls test is failing as the source is still pointing to the old .zip -- I don't have upload rights on OA, so figured I'd just leave it as such so someone can upload to an authoritative URL, and update the source accordingly.

### Automatic download

image

### Manual download

image

## Output stats

image

## Sample output

[hk-sample.csv](https://github.com/user-attachments/files/28660895/hk-sample.csv)


tijptjik:
The data-pls test is failing as the source is still pointing to the old .zip -- I don't have upload rights on OA, so figured I'd just leave it as such so someone can upload to an authoritative URL, and update the source accordingly.

PR #8245, opened 2026-07-06
au/nz centerlines

andrewharvey:
## nz

https://docs.topo.linz.govt.nz/data-dictionary/tdd-class-road_cl.html


andrewharvey:
I might split up this PR to refine it further without triggering every source to rebuild each time.

PR #8276, opened 2026-08-07
Add us/il/christian parcels

### LINKED ISSUE #8275: Add us/il/christian parcels
Esri endpoint here:

https://services.arcgis.com/Xn3XOQd1zDlYr9z7/arcgis/rest/services/Parcels/FeatureServer/0

rmx90210:
Closes #8275

### DIFF
--- a/sources/us/il/christian.json
+++ b/sources/us/il/christian.json
@@ -33,6 +33,17 @@
                 "note": "bhamaps.com servers are dead as of 2026-04",
                 "skip": true
             }
+        ],
+        "parcels": [
+            {
+                "name": "county",
+                "data": "https://services.arcgis.com/Xn3XOQd1zDlYr9z7/arcgis/rest/services/Parcels/FeatureServer/0",
+                "protocol": "ESRI",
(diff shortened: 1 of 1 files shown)


rmx90210:
@iandees can you think of any reason why the [View Map](https://batch.openaddresses.io/job/887578) from Data Please says "No Preview Image Found"? Another recent PR (#8279) seems to have a working map, so just curious if this is mergeable or needs fixing  👀

iandees:
I see it now, so I wonder if it takes a little bit of time to show up?

rmx90210:
@iandees Ahh yes, it looks like this was cached by my browser:

`GET https://batch.openaddresses.io/job/887579`

The cached response had `output.preview: false`, so I had to clear the cache to get rid of the "No Preview Image Found" message. I think it might help to lower the `Cache-Control max-age` for `/job` (currently set to `432000`) or just turn off caching, since I would assume the responses change quite a bit as the job progresses. I can open an issue over at the batch repo!

Issue #8280, opened 2026-08-07
Find new parcel sources for 13 IL counties previously using bhamaps.com


rmx90210:
Commit 8bcae7d set `skip: true` for 13 counties in Illinois that previously sourced data from bhamaps.com, which hosted Esri endpoints that appeared to stop working all at once. AFAIK these counties still have public parcel viewers, so there may be new endpoints we can use. Note that I've only listed parcel sources, but some of the servers may have address points as well.

- Christian
  - Notes: already an issue at #8275 and a PR at #8276 
- Clark
  - Map viewer: https://beacon.schneidercorp.com/Application.aspx?App=ClarkCountyIL&PageType=Map
  - Notes: it's a Beacon map, so a no-go for now (or until #6375 is revisited)
- Edgar
  - Map viewer: https://edgar-il.bhamaps.com/
  - Endpoint: https://services5.arcgis.com/6VdFeKk2GI4rFk2D/arcgis/rest/services/Reach_Parcels/FeatureServer/0
- Jackson
  - Map viewer: https://jackson-il.bhamaps.com/
  - Endpoint: https://services1.arcgis.com/23PZWuLADSqYBEO7/arcgis/rest/services/Reach_Parcels/FeatureServer/2
- Jersey
  - Endpoint: https://services.arcgis.com/4YineAQdtmx0tv46/ArcGIS/rest/services/JerseyILFeatures/FeatureServer/5
  - Notes: endpoint was found incidentally via the landing page of a public ArcGIS server
- Lawrence
  - Map viewer: https://lawrence-il.bhamaps.com/
  - Endpoint: https://services2.arcgis.com/AtoBX9sHdZu4i3gx/arcgis/rest/services/Parcels/FeatureServer/0
- Macoupin
  - Map viewer: https://macoupin-il.bhamaps.com/
  - Endpoint: https://services.arcgis.com/LYDdxoqTV7fw9teL/arcgis/rest/services/Reach_Parcels/FeatureServer/0
- Massac
  - Notes: already an issue at #8277 and PR at #8278
- Menard
  - Notes: Beacon
- Moultrie
  - Map viewer: https://experience.arcgis.com/experience/08a3417caac9468ab90b77ce1e7f85af
  - Endpoint: https://www.colesco.illinois.gov/arcgis/rest/services/Moultrie/Moultrie_Parcels/MapServer/0
- Putnam
  - Map viewer: https://putnamil.wthgis.com/
[...]

PR #8283, opened 2026-08-12
Update us/il/edgar parcels

rmx90210:
Relates to #8280


rmx90210:
[Logs](https://batch.openaddresses.io/job/892410/log) have a 500 error.

rmx90210:
@iandees this one seems to have glitched. Would it be possible to trigger a rerun, unless anything jumps out at you that I can fix?

Issue #8285, opened 2026-08-12
Add/update IL parcel sources


rmx90210:
In addition to the parcel sources that were added/updated in #8280 (which was scoped to sources previously using bhamaps.com), a number of other counties have parcels that can be added or updated:

## Add new source (county doesn't exist in OA yet)

- [x] Clay
- [x] Clinton
- [x] Fayette
- [x] Ford
- [x] Gallatin [depends on #8320]
- [x] Hamilton
- [x] Hancock
- [x] Henderson
- [x] Iroquois
- [x] Jasper
- [x] Jefferson
- [x] Marshall
- [x] Mason [depends on #8320]
- [x] Monroe
- [x] Pike [depends on #8320]
- [x] Richland [depends on #8320]
- [x] Wayne

## Add parcels (we already have them as an address source)

- [x] Douglas [addresses: 🟢 ok]
- [x] DuPage [addresses: 🟢 ok]
- [x] Fulton [addresses: 🟢 ok]
- [x] Henry [addresses: 🟢 ok]
- [x] Logan [addresses: 🟢 ok]
- [x] Rock Island [addresses: 🟢 ok] 
- [x] Effingham [addresses: 🟡 stale]
- [x] Kane [addresses: 🟡 stale]
- [x] Kankakee [addresses: 🟡 stale]
- [x] Knox [addresses: 🟡 stale]
- [x] Lee [addresses: 🟡 stale]
- [x] Piatt [addresses: 🟡 stale]

## Stale / not working

- [ ] Carroll [token now required]
- [x] Cook
- [x] McHenry
- [x] Sangamon
- [x] Stark
- [ ] Will [works sometimes; seems to be intermittently timing out?]
- [x] Woodford

PR #8305, opened 2026-08-13
us/il/effingham: add parcels and update addresses

rmx90210:
Relates to #8285

Also skips address source for now; URL is broken


rmx90210:
Looking at the [jobs](https://batch.openaddresses.io/run/15185) it looks like parcels succeeded but the addresses failing, even though I set `"skip": true`. Is that error state expected when skipping a source @iandees?

Same results on #8307 AFAICT.

iandees:
Yea, I think that's expected. When we built the skip functionality I think it was meant to be a temporary provision and sources that are skipped in the long term should be removed from the dataset. Maybe the "removed" categorization should be a "successfully downloaded nothing" case?

rmx90210:
That helps to know @iandees. Maybe there's room for something like a "skipped" or "paused" status? I think I would intuitively know what that meant if I saw it on the health dashboard, for instance.

For the time being I fixed the addresses for this county so everything should be good to go :)

rmx90210:
This should be good to merge now, whenever you have a chance @iandees 🙏

Issue #8306, opened 2026-08-13
Add a `headers` field to the source schema for HTTP sources


rmx90210:
Implements #1745. Unblocks #4550, #6226, and likely other sources that expect HTTP headers.

Depends on: openaddresses/batch-machine#102 — that must be released and deployed first, or sources will pass CI validation while their headers are silently dropped.

## Proposed schema addition

New `schema/util/headers.json`, following the `util/timeout.json` pattern:

```json
{
    "type": "object",
    "description": "Additional HTTP request headers to send when downloading this source.",
    "additionalProperties": { "type": "string" }
}
```

Referenced as `"headers": { "$ref": "../util/headers.json" }` from all four layer schemas (`address`, `building`, `centerline`, `parcel`) — all four are `additionalProperties: false`, so each needs it. Usage:

```json
{
  "name": "champaign",
  "protocol": "http",
  "data": "https://example.gov/addresses.geojson",
  "headers": { "Referer": "https://example.gov/gis/" }
}
```

## Heads-up on naming

`conform.headers` already exists and means something unrelated (which CSV row holds the column names). The new field sits one level up as a sibling of `protocol`/`data`, so they're structurally unambiguous, but the docs for each should cross-reference the other. Happy to use `request_headers` instead if folks prefer to avoid the naming overload.

## Also included

- Validation tests in `test/schema_validation_v2.test.js`.
- A `CONTRIBUTING.md` row documenting the field — including an explicit warning not to put credentials or API keys in it, since this repo is public.
- Headers apply to `http` and `ESRI` protocols only; they're ignored for `ftp`.

I'm happy to open the PRs for both this and `batch-machine`.

iandees:
I think we should wrap `headers` in something like a `http_request_settings` or something so that we can also support setting a `token` query param. Otherwise this looks good.

rmx90210:
Good call to wrap `headers` @iandees 👍

Any objection to calling the wrapper `request` (so for example the headers would land in `request.headers`)? Just thinking it's a bit shorter and matches what some HTTP clients call it. But happy to stick with `http_request_settings` too.

iandees:
`request` sounds good.

iandees:
I think we should be all merged and deployed here.

rmx90210:
@iandees thanks so much for taking a look and merging — I will try adding a source with headers next!

rmx90210:
#8315 seems to have run OK (and uses `request.headers`), so I think this is working! 🚀

Issue #6226, opened 2023-01-29
[add data] Schuylkill County, PA


rbrtmrtn:
Schuylkill County has a map viewer:

https://services.co.schuylkill.pa.us/mapviewer/

which makes calls an Esri REST service like this:

https://services.co.schuylkill.pa.us/Proxy/proxy.ashx?https://services.co.schuylkill.pa.us/arcgis/rest/services/ParcelLocator/Base_Layers/MapServer/identify?f=json&tolerance=10&returnGeometry=true&imageDisplay=1330%2C370%2C96&geometry=%7B%22x%22%3A-8482639.223952241%2C%22y%22%3A4965700.163761963%7D&geometryType=esriGeometryPoint&sr=3857&mapExtent=-8482881.672651313%2C4965579.536576711%2C-8482087.444154356%2C4965800.487361578&layers=all%3A0%2C1%2C2%2C3

You can curl it if you set `Referer: https://services.co.schuylkill.pa.us/mapviewer/`.

I'm newish to OA — is this too scrapy? 😄

Issue #4550, opened 2019-06-05
Fix Champaign, IL


justinelliotmeyers:
broken:
http://results.openaddresses.io/sources/us/il/champaign

new source:
https://services.ccgisc.org/proxy/proxy.ashx?https://server.ccgisc.org/arcgisserver/rest/services/Webmap_Addressing/AddressPointLabels/MapServer/0

rmx90210:
Champaign has a map viewer here:

https://maps.ccgisc.org/public/Default.aspx

Sample parcel query:

https://services.ccgisc.org/portal/sharing/servers/9ed62f9e670e4d8e9d797ed62b89b859/rest/services/General/ParcelSearch/MapServer/0/query?f=json&where=&returnGeometry=false&spatialRel=esriSpatialRelIntersects&geometry=%7B%22x%22%3A1024438.9743186121%2C%22y%22%3A1302233.9386791703%2C%22spatialReference%22%3A%7B%22wkid%22%3A3435%7D%7D&geometryType=esriGeometryPoint&inSR=3435&outFields=PIN&outSR=3435

It doesn't work unless you pass `referer = https://maps.ccgisc.org/`

Issue #1745, opened 2016-06-21
Allow `referer` request header to be passed to HTTP sources


trescube:
I found a source at `http://cmap.clackamas.us/dataportal/index.jsp` that POSTs (but GETs work fine, too) to a utility script that requires a specific `referer` header value of `http://cmap.clackamas.us/dataportal/index.jsp`.

rmx90210:
Hello @iandees ! 👋 

I was trying to get `us/il/champaign` addresses working (relates to #4550) but the endpoint requires a `referer` header. Do you happen to know if anyone else has tried adding `referer` support over the years, and if so, how it went? I could only find this issue which seems to still be ready for work. I could take a crack at it.

This would potentially resolve #6226 as well.

iandees:
I don't think anyone has tackled this. I believe there's support in batch-machine and machine to send arbitrary headers, it's just not wired into the json schema.

PR #8310, opened 2026-08-13
Add `request.headers` field to the source schema

rmx90210:
Implements #1745 and #8306. Unblocks #4550 (Champaign, IL) and likely #6226 (Schuylkill County, PA).

### What

Adds a `request` object to the four layer schemas (`address`, `building`, `centerline`, `parcel`), with a `headers` property that lets a source specify additional HTTP request headers to send when downloading — e.g. a `Referer` header for hosts that gate downloads on it.

```json
{
  "name": "champaign",
  "protocol": "http",
  "data": "https://example.gov/addresses.geojson",
  "request": {
    "headers": { "Referer": "https://example.gov/gis/" }
  }
}
```

Headers are nested under `request` rather than flat, per @iandees's request in #8306, so a future setting (e.g. a query-param token) can be added alongside it without a second schema migration. `request` has `additionalProperties: false`, so it only accepts `headers` for now.

Not to be confused with `conform.headers`, an existing, unrelated field that selects which CSV row holds column names. The schema description and CONTRIBUTING.md both cross-reference the other so this doesn't read as a naming mistake in review.

### Changes

- `schema/util/headers.json` — new, describes the headers map itself
- `schema/util/request.json` — new, the container (`additionalProperties: false`, holds `headers`)
- `schema/layers/{address,building,centerline,parcel}.json` — each now references `request`
- `test/schema_validation_v2.test.js` — new cases: valid `request.headers`, non-object `request`, non-object `request.headers`, non-string header values, acceptance on all four layer types, and that an unrecognized key inside `request` is still rejected
- `CONTRIBUTING.md` — documents `{layer}.request.headers`, cross-referenced with `conform.headers`

### ⚠️ Deployment dependency

The corresponding read of this field lives in `openaddresses/batch-machine`, on branch `support-custom-http-headers` — **not yet merged or released**. That branch reads `source_config.data_source.get('request', {}).get('headers', {})` and threads it through the download path (including the file-extension pre-flight request, and ESRI via `extra_headers`).

**This PR should not be merged before `batch-machine` ships that change.** If it merges first, a source using `request.headers` would pass schema validation here but have the field silently ignored by the machine — headers wouldn't actually be sent, and the contributor would have no indication anything was wrong.

Recommended order:
1. Merge and release `batch-machine`'s `support-custom-http-headers` work
2. Merge this PR
3. Only then add `sources/us/il/champaign.json` (or other sources) using `request.headers`

### Testing

```bash
npm test   # 3391 passing; 2 pre-existing failures unrelated to this change (non-integer accuracy test)
npm run lint
```

Manually confirmed:
- a fixture with `request.headers` validates
- a fixture with the old flat `headers` key (no `request` wrapper) now correctly fails `additionalProperties`
- a fixture with an unrecognized key inside `request` (e.g. `request.bogus`) fails


rmx90210:
Hello @iandees! Any chance you might have a moment to take a look at this PR, and the corresponding one in `batch-machine`? I'd love to add a few more sources that need headers, so this would be a huge help.

PR #8327, opened 2026-08-20
us/il/wayne: add parcels and addresses

rmx90210:
Relates to #8285


rmx90210:
Hey @iandees, it says this has been running for close to 24 hours. I can't see the logs [here](https://batch.openaddresses.io/run/15309). Do you have any more visibility into what might be happening?

iandees:
There was a problem with the API deploy getting stuck in a loop. Should be good now.

PR #8334, opened 2026-08-23
Close unclosed polygon rings in WTH parcel scraper

iandees:
The WTH ThinkGIS parcel viewer scraper (`scripts/us/il/wth/lib.js`, added in #8321) decodes each parcel's delta-encoded `<poly>` ring but never checked whether the ring actually closed. The site draws these as polylines for its own hover UI, so some don't loop back to the start point - GDAL then rejects them as invalid geometry (`Points of LinearRing do not form a closed linestring`), which is what broke CI on #8331 (richland, F=1138) and #8330 (pike, similar).

Fix: `decodeRing()` now appends the first point to close the ring if the last decoded point doesn't match it. Verified against richland's live service - re-scraping F=1..1140 (which includes the previously-failing F=1138) now produces 0 unclosed rings across 1140 parcels.

This only fixes the scraper going forward. #8331 and #8330 point at already-uploaded snapshots from before this fix, so they'll still need to be re-scraped and re-uploaded to pick it up.

### DIFF
--- a/scripts/us/il/wth/lib.js
+++ b/scripts/us/il/wth/lib.js
@@ -31,6 +31,14 @@ function decodeRing(polyText) {
     y += parseInt(coords[i * 2 + 1], 10);
     ring.push(z192LL(x, y));
   }
+  // the site draws these as polylines, not necessarily closed rings - GDAL
+  // rejects an unclosed LinearRing, so close it if the last point didn't
+  // land back on the first
+  if (ring.length >= 2) {
+    const [firstLon, firstLat] = ring[0];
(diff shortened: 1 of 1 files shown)


iandees:
FYI @rmx90210 I'll merge this but can you run it again?

rmx90210:
Good catch @iandees, I'll pull your changes down and run it again 👍

rmx90210:
@iandees #8330 and #8331 should be good to merge momentarily (one succeeded and the other is still running — expecting they're both fixed at this point)

PR #8353, opened 2026-09-06
fix broken AU-NSW dataset

k-yle:
the [dataset for `au/nsw/statewide`](https://batch.openaddresses.io/job/907114) is currently unusable, because the `number` field is missing from over 2 million addresses. 

The regex currently assumes that every housenumber contains a `/`, which is wrong. only appartment buildings have this character.

test cases for the regex here: https://regexr.com/8ocsq

cc @andrewharvey

### DIFF
--- a/sources/au/nsw/statewide.json
+++ b/sources/au/nsw/statewide.json
@@ -35,7 +35,7 @@
                     "number": {
                         "function": "regexp",
                         "field": "housenumber",
-                        "pattern": "^(?:.*/)(.*)",
+                        "pattern": "^(?:.*/)?(.*)",
                         "replace": "$1"
                     },
                     "street": {


andrewharvey (approved):
Glad you caught this, thank you!

andrewharvey:
Looks like this only became an issue recently per https://batch.openaddresses.io/data/489/history#map=-0.77/0/-113.9

image

It must have been a change in the source data, since our configuration hasn't changed.

k-yle:
yeah i don't really understand what changed, since that regex hasn't been touched in 5 years. 

maybe the source data used values like `/12`  until recently? 🤷

andrewharvey:
> maybe the source data used values like /12 until recently? 🤷

I think so.

PR #8354, opened 2026-09-09
Clean up validation issues

jeffdefacto:
Removes empty strings, redundant lat, lons, and a couple smaller inconsistencies in the datas.


jeffdefacto:
The remaining failures are source issues. This is good enough for now.

PR #8379, opened 2026-09-12
Fix broken Tulsa County addresses source

iandees:
The addresses/county layer for Tulsa County, OK was failing every job with:

```
esridump.errors.EsriDownloadError: Could not retrieve layer metadata: Service Address_Points/MapServer not started
```

The old `Address_Points/FeatureServer/2` service on INCOG's map server (map11.incog.org) has been replaced by `Address_Points_V3/FeatureServer/2` on the same server. It's the same INCOG regional address point dataset (same 452,971 feature count, same extent covering the Tulsa metro area, same field set) just republished under a new service name, with the house number field renamed from `Address` to `AddNumber`.

Verified before committing:
- `?f=json` returns a valid `fields` array, no auth error
- Feature count query returns 452,971 features
- Extent covers the expected Tulsa County / INCOG metro geography
- Sample records confirm `AddNumber`, `PreDir`, `Street`, `StreetType`, `SufDir`, `City`, `Zipcode` field names

Only the `addresses`/`county` layer was touched; `parcels` and `centerlines` were not reported broken and were left as-is.

### DIFF
--- a/sources/us/ok/tulsa.json
+++ b/sources/us/ok/tulsa.json
@@ -14,11 +14,11 @@
         "addresses": [
             {
                 "name": "county",
-                "data": "https://map11.incog.org/arcgis11wa/rest/services/Address_Points/FeatureServer/2",
+                "data": "https://map11.incog.org/arcgis11wa/rest/services/Address_Points_V3/FeatureServer/2",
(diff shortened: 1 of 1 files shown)


iandees:

PR #8380, opened 2026-09-12
Fix broken Osage County, OK addresses source

iandees:
The INCOG ArcGIS service backing this source was renamed/reorganized; the old `Address_Points/FeatureServer/2` layer now returns "Service Address_Points/MapServer not started" (confirmed via jobs 908446, 903496, 898604, all Fail).

- Found the replacement layer on the same server: `Address_Points_V3/FeatureServer/2` (identical duplicate also exists as `Address_Points2_V3`), still published by INCOG.
- Verified via `?f=json` and sample queries: fields present, 452,971 total features, 6,266 with `County = 'OSAGE COUNTY'`, no auth errors. Coverage for Osage County addresses is limited to the southern portion of the county near the Tulsa metro area, matching the same limited coverage the prior (now-dead) service provided for this source.
- Field `Address` was renamed to `AddNumber` in the new service; `PreDir`, `Street`, `StreetType`, `SufDir`, `City`, and `Zipcode` are unchanged.
- Only the `addresses`/`county` layer was touched; `parcels`/`county` was left as-is since it was not reported broken.
- Validated against the schema (inline `test/lib.js` compile+validate: VALID).

### DIFF
--- a/sources/us/ok/osage.json
+++ b/sources/us/ok/osage.json
@@ -14,11 +14,11 @@
         "addresses": [
             {
                 "name": "county",
-                "data": "https://map11.incog.org/arcgis11wa/rest/services/Address_Points/FeatureServer/2",
+                "data": "https://map11.incog.org/arcgis11wa/rest/services/Address_Points_V3/FeatureServer/2",
                 "protocol": "ESRI",
                 "conform": {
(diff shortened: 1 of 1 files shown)


iandees:

PR #8384, opened 2026-09-12
Fix broken Rogers County, OK addresses source

iandees:
The Rogers County addresses source (`sources/us/ok/rogers.json`) was failing with `Service Address_Points/MapServer not started` — INCOG's `Address_Points` service was retired and replaced by `Address_Points_V3`.

- Confirmed the last 3 jobs (908444, 903494, 898602) all failed with this error.
- Verified the new service (`Address_Points_V3/FeatureServer/2`) is public, returns 452,971 features region-wide including 47,145 for Rogers County, and covers the correct extent (near Claremore, OK).
- Field names changed slightly: the house number field is now `AddNumber` (was `Address`); `PreDir`, `Street`, `StreetType`, `SufDir`, `City`, and `Zipcode` are unchanged.
- Updated the `data` URL and `number` conform field accordingly. No other fields needed changes.
- This is the same shared INCOG regional address layer already used identically by the other INCOG-area OK county sources (creek, osage, tulsa, wagoner), so no coverage widening — just pointing at the renamed service.
- Parcels layer for this source uses a separate, unrelated service and was left untouched.

Validated against the schema with `npm test`.

### DIFF
--- a/sources/us/ok/rogers.json
+++ b/sources/us/ok/rogers.json
@@ -14,11 +14,11 @@
         "addresses": [
             {
                 "name": "county",
-                "data": "https://map11.incog.org/arcgis11wa/rest/services/Address_Points/FeatureServer/2",
+                "data": "https://map11.incog.org/arcgis11wa/rest/services/Address_Points_V3/FeatureServer/2",
                 "protocol": "ESRI",
                 "conform": {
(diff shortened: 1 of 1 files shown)


iandees:

PR #8385, opened 2026-09-12
Fix broken Wagoner County, OK addresses source

iandees:
The addresses/county layer was failing with `Service Address_Points/MapServer not started` — the INCOG regional service backing this source (`Address_Points/FeatureServer/2`) has been retired.

Checked the server's service listing (`map11.incog.org/arcgis11wa/rest/services`) and found the replacement regional address point layer, `Address_Points_V3/FeatureServer/2`, which still covers the full INCOG region including Wagoner County. Confirmed via query that it returns 18,707 features where `County = 'WAGONER COUNTY'`, and sampled records to verify field names.

Changes:
- Updated `data` URL to the new service (`Address_Points_V3` instead of the retired `Address_Points`).
- Updated `number` conform field from `Address` (no longer present) to `AddNumber`, the current house-number field.
- `street`, `city`, and `postcode` conform fields (`PreDir`/`Street`/`StreetType`/`SufDir`, `City`, `Zipcode`) are unchanged and confirmed present in the new schema.

The parcels layer for this source was not touched.

Note: several other OK county sources (Creek, Osage, Tulsa, Rogers) pointed at the same retired `Address_Points/FeatureServer/2` URL and are being fixed independently in parallel.

### DIFF
--- a/sources/us/ok/wagoner.json
+++ b/sources/us/ok/wagoner.json
@@ -14,11 +14,11 @@
         "addresses": [
             {
                 "name": "county",
-                "data": "https://map11.incog.org/arcgis11wa/rest/services/Address_Points/FeatureServer/2",
+                "data": "https://map11.incog.org/arcgis11wa/rest/services/Address_Points_V3/FeatureServer/2",
                 "protocol": "ESRI",
                 "conform": {
                     "format": "geojson",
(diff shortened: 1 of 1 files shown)


iandees:

PR #8410, opened 2026-09-12
Fix broken Coles County, IL addresses source

iandees:
The addresses layer had been manually set to `"skip": true` with the note "colesco.illinois.gov has SSL certificate failure", causing every job for this layer to fail with `Error: out.geojson not found` (jobs 909631, 904681, 899785).

The underlying SSL issue is resolved: the server now presents a valid certificate (issued Mar 2026, expiring Sep 2026), and the `parcels` layer — which reads the exact same MapServer layer (`ColesCounty/WebTaxMap2023/MapServer/98`) — has succeeded on every recent weekly run. I independently confirmed the layer:

- Returns valid metadata with a `fields` array, no auth errors
- Reports 27,608 features (reasonable for a single county)
- Serves paginated queries quickly (2,000 records in ~1.6s, well within job time limits)
- Still exposes the same field names used in the existing conform (`gis_data.site_address`, `gis_data.site_csz`)

This removes the stale `skip`/`note` keys so the addresses layer resumes normal processing. No conform changes were needed since the field names are unchanged.

### DIFF
--- a/sources/us/il/coles.json
+++ b/sources/us/il/coles.json
@@ -35,13 +35,13 @@
                     "region": {
                         "function": "regexp",
                         "field": "gis_data.site_csz",
-                        "pattern": "(IL)",
+                        "pattern": "^.*\\s(IL)\\s\\d{5}$",
                         "replace": "$1"
                     },
                     "postcode": {
                         "function": "regexp",
                         "field": "gis_data.site_csz",
-                        "pattern": "(\\d{5})",
+                        "pattern": "^.*\\sIL\\s(\\d{5})$",
                         "replace": "$1"
                     }
                 }
--- a/sources/us/il/coles.json
+++ b/sources/us/il/coles.json
@@ -44,9 +44,7 @@
                         "pattern": "(\\d{5})",
                         "replace": "$1"
                     }
-                },
-                "skip": true,
-                "note": "colesco.illinois.gov has SSL certificate failure"
+                }
             }
         ],
         "parcels": [


iandees:
Closing — this isn't actually fixed. Verified directly against the service: `supportsPagination: false` and any `resultOffset` query returns `{"code":400,"message":"Pagination is not supported."}`, even near the end of the 27,608-record dataset. `maxRecordCount` is only 2000, so removing the `skip` flag would make the batch job silently retrieve only the first ~7% of the county's parcels with no way to page through the rest (or error, depending on how esridump handles it) — not a real fix. The original `skip: true`/note should stay, just updated to reflect the real current blocker (pagination disabled server-wide) rather than the stale SSL-certificate reason. Re-verify if the county's ArcGIS Server ever gets a config update enabling pagination.

iandees:
Reopening — I was wrong to close this. I tested `resultOffset`-based and `objectIds`-based queries manually and both failed, but that's not proof the source is unfixable: the `parcels` layer on this exact same MapServer/98 service has been completing successfully in production with the full 27,608-record count (confirmed via job #909632's actual output), not a truncated 2,000. Whatever fallback pagination strategy esridump uses for this server, it clearly works — my manual curl tests just didn't replicate it. Sorry for the noise; this fix is good.

iandees:
Pushed a follow-up fix: `region` and `postcode` were both returning the full `gis_data.site_csz` string (e.g. `"TRILLA IL 62469"`) instead of the isolated `IL`/`62469` values, because their regex patterns were unanchored — only `city`'s pattern happened to be anchored across the whole string, which is why it alone came out correct. Verified against the actual CI job output (job 912154) before and after: normal records now correctly extract just `IL` and the 5-digit zip, and the ~75 malformed `site_csz` records (missing a proper zip) correctly fall through to empty rather than garbage, matching `city`'s existing behavior on those same rows.

PR #8420, opened 2026-09-12
Fix broken Harford County, MD addresses source

iandees:
The addresses/county source pointed at https://geodata.md.gov/appdata/rest/services/Addressing/MD_Addressing/MapServer/13, which has been returning HTTP 503 "Site Maintenance" for the whole geodata.md.gov host since at least 2026-09-04 (jobs 904306 and 909256 both failed on this, and the host is still down as of today).

The same layer (MapServer layer 13, named "Harford", same 115,238-record Harford-County-only extent) is live under an alternate hostname for Maryland's statewide addressing service: https://mdgeodata.md.gov/appdata/rest/services/Addressing/MD_Addressing/MapServer/13. Verified:

- `?f=json` returns valid metadata, no auth error
- Feature count: 115,238, all `County` = "Harford County"
- Spatial extent matches Harford County (~-76.57 to -76.08 lon, 39.39 to 39.72 lat)

The field schema on this layer has also changed since the source was last updated (from `ADDID`/`ADDNUMCPLT`/`NAMECPLT`/etc. to NENA-style `Add_Id`/`Add_Number`/`St_Name`/`St_PreDir`/`St_PosTyp`/`St_PosDir`/`Unit`/`Post_Comm`/`County`/`State`/`Post_Code`), so `conform` was rewritten to match the current field names, verified against sample records.

### DIFF
--- a/sources/us/md/harford.json
+++ b/sources/us/md/harford.json
@@ -14,18 +14,18 @@
         "addresses": [
             {
                 "name": "county",
-                "data": "https://geodata.md.gov/appdata/rest/services/Addressing/MD_Addressing/MapServer/13",
+                "data": "https://mdgeodata.md.gov/appdata/rest/services/Addressing/MD_Addressing/MapServer/13",
                 "protocol": "ESRI",
                 "conform": {
                     "format": "geojson",
-                    "id": "ADDID",
-                    "number": "ADDNUMCPLT",
-                    "street": "NAMECPLT",
-                    "unit": "SUBBADDCPLT",
-                    "city": "CITY",
-                    "district": "COUNTY",
-                    "region": "STATE",
-                    "postcode": "ZIPCODE"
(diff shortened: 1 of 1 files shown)


iandees:
Sanity check on job 912167 (run 15714): number/street/city/postcode/unit and coordinates all look correct and well-formatted (bounds within Harford County, MD). However, `region` comes out as "Ma" for all 115,238 records, not "MD". Confirmed this is an upstream data issue: querying the ESRI service directly (mdgeodata.md.gov .../MapServer/13, field `State`) shows the raw attribute value is literally "Ma" (field length is capped at 2 chars) for every record, not a mapping mistake in this PR. Holding off on merge since region is wrong for 100% of output rows. Suggest changing the `region` conform key to `{"function": "constant", "value": "MD"}` instead of passing through the raw `State` field, then re-running CI.

PR #8421, opened 2026-09-12
Fix broken Okanogan County, WA addresses source

iandees:
The addresses layer was failing (FTP 550 "Failed to open file") because the source pointed at a stale dated filename (`DISPATCH/2024-01-12_address_points.zip`) that no longer exists on the county's FTP server.

The server's `DISPATCH` directory was also renamed to `Dispatch` (case-sensitive FTP). Updated the `data` URL to the current file, `ftp://okgis.ddns.net/Dispatch/2026-07-20_address_points.zip`, and bumped `year` to 2026.

Verified the new file downloads as a valid shapefile ZIP and contains the same field names already used in the conform (`e911_add`, `e911_unit`, `dis_road`, `city`, `zip`), so no conform changes were needed.

Note: the `parcels` layer on this same source is also currently failing for a similar reason, but the replacement file's internal shapefile structure has changed (renamed to `Landuse`), so it needs separate investigation and is left out of this PR.

### DIFF
--- a/sources/us/wa/okanogan.json
+++ b/sources/us/wa/okanogan.json
@@ -16,8 +16,8 @@
                 "name": "county",
                 "protocol": "ftp",
                 "compression": "zip",
-                "year": "2024",
-                "data": "ftp://okgis.ddns.net/DISPATCH/2024-01-12_address_points.zip",
+                "year": "2026",
+                "data": "ftp://okgis.ddns.net/Dispatch/2026-07-20_address_points.zip",
                 "website": "https://okanogancounty.org/departments/planning/gis_mapping/available_digital_data.php",
                 "conform": {
                     "format": "shapefile",


iandees:
Investigated the repeated failure here — this isn't a flaky/infra issue, it's a genuine bad record in Okanogan County's own shapefile.

One address point (PIN `8848500040`, e911_add 268 LOST RIVER RD, MAZAMA) has a null/no-data sentinel geometry — both X and Y are `-1.7976931348623157e+308` (the classic ESRI-shapefile "no value" double). Everything downstream chokes on it: shapely emits an overflow warning converting it, and the batch pipeline's tippecanoe/pmtiles step then fails to parse the resulting degenerate coordinate, which fails the whole job (`BrokenPipeError`/EPIPE in `render_pmtiles`, then a JSON parse error in `stats.js`).

There's no conform-level way in the OA schema to drop a single bad record from a shapefile. The real fix needs either:
1. The county re-exporting their shapefile without the null-geometry record, or
2. Someone preprocessing this shapefile (drop the one bad feature) and re-uploading the cleaned copy via https://batch.openaddresses.io/upload, then pointing this PR's `data` URL at that cache URL instead.

I don't have upload credentials for (2), so I can't finish this one myself. Leaving it open rather than merging broken data or guessing at a workaround.

iandees:
Re-verified after this fix: the newer file (`2026-07-20_address_points.zip`) still has the identical defect as the old one — at least one record has a degenerate/malformed `coordinates` array (`[0]` instead of a proper [lon, lat] pair), which crashes the whole pipeline's JSON/geometry parsing rather than just producing bad data for that one record (confirmed via job 912296's log: `Found misspelling of null` followed by a full job failure, zero features processed). So this PR's URL fix is correct and worth keeping (it points at the current, real file instead of a 404), but the source will keep failing entirely until the county fixes their export or someone re-hosts a cleaned copy via batch.openaddresses.io/upload. Not merging yet — leaving open per the earlier assessment.

PR #8460, opened 2026-09-12
Fix broken Franklin County, AL addresses, parcels, and buildings sources

iandees:
The KCS GIS server hosting Franklin County's public map service was reorganized: `web5.kcsgis.com/.../Franklin/Public/MapServer` was replaced with a `Public_monthly` service that no longer publishes the parcel/address layer (105), causing `Service Franklin/Public/MapServer not started` failures for all three layers (addresses, parcels, buildings).

Found the county's current parcel viewer (isv.kcsgis.com/al.franklin_revenue) still references a live backend at `web6.kcsgis.com/kcsgis/rest/services/Franklin/Franklin_Public_ISV/MapServer`, which republishes the same layer structure under the same layer IDs (105 = Parcels, 102 = Buildings) with the same field names (SitusAddNumber, SitusAddName, SitusAddCity, PARCELID).

Verified before committing:
- Layer 105 (Parcels): 23,519 features, fields match the existing conform exactly, sample records show real Franklin County parcels (Russellville, Spruce Pine, etc.)
- Layer 102 (Buildings): 33,315 features
- Service full extent (~-88.20 to -87.48 lon, ~34.22 to 34.49 lat) matches Franklin County, AL
- No auth/token errors, no special User-Agent required
- Schema validated with `test/lib.js`

Only the `data` URLs changed; conform blocks are unchanged since field names carried over identically.

### DIFF
--- a/sources/us/al/franklin.json
+++ b/sources/us/al/franklin.json
@@ -19,8 +19,7 @@
                 "conform": {
                     "format": "geojson",
                     "number": "SitusAddNumber",
-                    "street": "SitusAddName",
-                    "city": "SitusAddCity"
+                    "street": "SitusAddName"
                 }
             }
         ],
--- a/sources/us/al/franklin.json
+++ b/sources/us/al/franklin.json
@@ -14,7 +14,7 @@
         "addresses": [
             {
                 "name": "county",
-                "data": "https://web5.kcsgis.com/kcsgis/rest/services/Franklin/Public/MapServer/105",
+                "data": "https://web6.kcsgis.com/kcsgis/rest/services/Franklin/Franklin_Public_ISV/MapServer/105",
                 "protocol": "ESRI",
                 "conform": {
                     "format": "geojson",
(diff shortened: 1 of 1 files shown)


iandees:
Held back from auto-merge during sanity check despite CI SUCCESS. Job 912218 (addresses, run 15749->15753/job 912218): count 23519, but validity shows 9031/23519 (~38%) number failures, and of the 14287 'valid' records, 10128 (~71%) have number literally '0'. City field (SitusAddCity) is empty for 100% of sampled records. The field mapping itself is unchanged from the previous (dead) source — this may just reflect genuine data quality in Franklin County's GIS layer (rural county, many parcels lacking assigned site addresses) — but the proportion is high enough that I'd like a human to confirm the new web6.kcsgis.com/Franklin_Public_ISV service is actually equivalent to the old one before merging, rather than a lesser-quality mirror.

iandees:
Fixed the `city` mapping bug: `SitusAddCity` is an ESRI `SmallInteger` field with no domain/lookup exposed by the service — it was a numeric code, not a city name, so it produced garbage. Removed it from the conform.

The other issue (many records with `number: 0`) is not a mapping bug — this layer (`MapServer/105`) is a **parcel polygon layer**, not an address-point layer, and `SitusAddNumber` is genuinely 0 for parcels with no assigned situs address (vacant land, road ROW, etc). This is the same 'sparse parcels-used-as-addresses' tradeoff flagged on a few other PRs from this batch (e.g. Mills County IA, Cass County IA) — real but incomplete data, not something I can improve without a different upstream source. Left for your judgment on whether partial coverage here is acceptable.

PR #8463, opened 2026-09-12
Fix broken City of Amarillo, TX addresses source

iandees:
The addresses/city source for Amarillo, TX has been failing all recent weekly jobs with:

esridump.errors.EsriDownloadError: Could not retrieve layer metadata: Invalid URL Invalid URL

The old ArcGIS Online-hosted service (services6.arcgis.com/.../AddressPointsArcGISOnline/FeatureServer/1) no longer exists.

Investigation:
- The City of Amarillo's ArcGIS Online org (same org id, Vdk8uHgdgYx8ZqS6) has a similarly-named public service (AddressPointsDroneSense/FeatureServer/0), but it is a regional joint-jurisdiction dataset (Potter/Randall 911 addressing, covering Amarillo plus Canyon, Bushland, Lake Tanglewood, Timbercreek Canyon, and unincorporated county) with no way for this repo's ESRI source schema to filter it down to just Amarillo, so it was not used.
- Found the City's own on-premise ArcGIS Server at gismaps.amarillo.gov, which hosts an "AddressPoints" FeatureServer with the address points layer at index 1 — the same layer index, and identical field names/schema (ADDRESS, APP_UNIT_ID, ST_DIR, ST_NAME, ST_TYPE, ST_POSTDIR, POST_CITY, ZIPCODE) as the previously-working source. This is the authoritative, currently-maintained replacement for the dead AGOL copy.
- Verified the new endpoint returns valid metadata, no auth is required, and it returns 125,608 features with the expected fields.

Updated the `data` URL to the new FeatureServer endpoint and added a `website` link to the City's GIS portal. The `conform` block is unchanged since all field names carried over exactly.

### DIFF
--- a/sources/us/tx/city_of_amarillo.json
+++ b/sources/us/tx/city_of_amarillo.json
@@ -27,7 +27,12 @@
                 "protocol": "ESRI",
                 "conform": {
                     "format": "geojson",
-                    "number": "ADDRESS",
+                    "number": {
+                        "function": "regexp",
+                        "field": "ADDRESS",
+                        "pattern": "^([0-9]+)(?:\\.0)?$",
+                        "replace": "$1"
+                    },
                     "street": [
                         "ST_DIR",
                         "ST_NAME",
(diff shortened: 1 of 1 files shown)


iandees:
Held back from auto-merge during sanity check despite CI SUCCESS. Job 912224 (addresses): count 125608, validity good (99.99%), number/street/city all look correct (e.g. AMARILLO/CANYON, real street names). However postcode is systematically malformed: every sampled record has a trailing '.0' (e.g. '79119.0' instead of '79119'), which points to the new gismaps.amarillo.gov ZIPCODE field being typed as a double instead of a string/text field. Recommend wrapping postcode in a regexp (e.g. strip trailing '.0') before merging.

PR #8518, opened 2026-09-13
Fix unanchored regexp bugs in Cumberland County, IL conform

iandees:
The \`region\` and \`postcode\` conform fields for \`sources/us/il/cumberland.json\` used the \`regexp\` function with a \`replace\` pattern that was completely unanchored (no \`^\`/\`$\`). Since \`replace\` performs a find/replace on the whole string rather than a clean extraction, an unanchored pattern with \`replace: "$1"\` leaves the entire original string in place except for the small matched substring being swapped for itself — the field ends up containing the full raw \`TSC_site_csz\` value (e.g. \`"NEOGA IL 62447"\`) instead of just \`"IL"\` or \`"62447"\`.

Confirmed via live samples from the ESRI service that \`TSC_site_csz\` is consistently shaped as \`"<CITY> IL <ZIP5>"\` (e.g. \`CHARLESTON IL 61920\`, \`NEOGA IL 62447\`, \`LERNA IL 62440\`, \`MATTOON IL 61938\`, \`TRILLA IL 62469\`).

Fix: anchored both patterns with \`^\`/\`$\` so the match consumes the entire string:
- \`region\`: \`(IL)\` -> \`^.*(IL).*$\`
- \`postcode\`: \`(\\d{5})\` -> \`^.*(\\d{5})$\`

Verified against real sample values with a standalone \`re.sub\` simulation and confirmed schema validity with the project's test harness.

No other files touched.

### DIFF
--- a/sources/us/il/cumberland.json
+++ b/sources/us/il/cumberland.json
@@ -35,13 +35,13 @@
                     "region": {
                         "function": "regexp",
                         "field": "TSC_site_csz",
-                        "pattern": "(IL)",
+                        "pattern": "^.*(IL).*$",
                         "replace": "$1"
                     },
                     "postcode": {
                         "function": "regexp",
                         "field": "TSC_site_csz",
-                        "pattern": "(\\d{5})",
+                        "pattern": "^.*(\\d{5})$",
                         "replace": "$1"
                     }
                 }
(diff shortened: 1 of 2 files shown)


iandees:
Closing — this branch accidentally picked up an unrelated Kankakee County change from a concurrent parallel fix run sharing the same working directory. Replaced by #8533, which contains only the Cumberland fix.

PR #8522, opened 2026-09-13
Fix unanchored regexp bug in Geelong addresses number conform

iandees:
The addresses \`number\` conform in \`sources/au/vic/city_of_greater_geelong.json\` is a \`chain\` of \`remove_postfix\`/\`regexp\` steps that strips the suburb, state, and postcode text off the source's combined \`FullAddres\` field (e.g. \`"335 Clarkes Road, ANAKIE VIC 3213"\`) down to just the house number.

Two of the \`regexp\` steps used \`replace\` with a pattern anchored only on the left (\`^\`) and not on the right (\`$\`):

- \`"pattern": "^(.*)[ ]+VIC", "replace": "$1"\`
- \`"pattern": "^(.*),", "replace": "$1"\`

\`regexp\`+\`replace\` only substitutes the matched portion of the string — any trailing text outside the match is left in place untouched. Because these patterns aren't anchored at the end, a matched-and-replaced result can retain trailing text that was supposed to be considered "consumed," which then breaks the following \`remove_postfix\` step (which expects the working value to end exactly with the suburb/street name) and cascades into every step after it.

I downloaded the actual shapefile behind this source (\`coggcadasterll.zip\`, 141k parcels) and replayed the full chain against real \`FullAddres\` values. With the bug, a value like \`"335 Clarkes Road, ANAKIE VIC 3213"\` comes out of the chain as \`"335 Clarkes Road ANAKIE "\` instead of \`"335"\` — the suburb and street name leak straight into the \`number\` field because the unanchored regexp steps leave stray characters that prevent the subsequent \`remove_postfix\` calls from matching.

Fix: anchor both patterns at the end (\`.*$\`) so each step consumes the entire matched prefix through end-of-string, matching the pattern already used correctly elsewhere in this same file for \`region\` (\`"^.*(VIC).*$"\`). Re-running the chain against real sample values after the fix produces the correct house numbers (\`"335"\`, \`"410"\`, \`"1435-1475"\`, \`"150"\`, \`"154"\`, etc.) for both plain and unit-prefixed addresses.

No other files or conform keys were touched.

### DIFF
--- a/sources/au/vic/city_of_greater_geelong.json
+++ b/sources/au/vic/city_of_greater_geelong.json
@@ -40,7 +40,7 @@
                             {
                                 "function": "regexp",
                                 "field": "oa:number_wip",
-                                "pattern": "^(.*)[ ]+VIC",
+                                "pattern": "^(.*)[ ]+VIC.*$",
                                 "replace": "$1"
                             },
                             {
@@ -51,7 +51,7 @@
                             {
                                 "function": "regexp",
                                 "field": "oa:number_wip",
-                                "pattern": "^(.*),",
+                                "pattern": "^(.*),.*$",
                                 "replace": "$1"
                             },
                             {
--- a/sources/us/ca/alpine.json
+++ b/sources/us/ca/alpine.json
@@ -32,7 +32,7 @@
                     "unit": {
                         "function": "regexp",
                         "field": "Situs",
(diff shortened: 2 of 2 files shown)


iandees:
Closing — this branch accidentally picked up an unrelated Alpine County, CA change from a concurrent parallel fix run sharing the same working directory. Replaced by #8534, which contains only the Geelong fix.

PR #8530, opened 2026-09-13
Fix disconnected chain steps in Fond du Lac County, WI conform

iandees:
## Bug

Per `ATTRIBUTE_FUNCTIONS.md`, any `chain` step after the first must reference the accumulated result via `oa:<variable>`, not the bare `variable` name or the original raw field — otherwise that step (and everything after it, unless something downstream happens to reference it correctly) is disconnected from the pipeline.

Both chains in `layers.addresses[0].conform` for this source had this bug:

- `street` (`variable: "street"`, 4 steps): steps 2-4 referenced the bare field `"street"` instead of `"oa:street"`, so the `USH`/`CTH` highway-prefix regexps and the final `DIRECTION`/street/`TYPE` join never saw each other's output.
- `city` (`variable: "c"`, 28 steps): steps 2-28 all referenced the bare field `"c"` instead of `"oa:c"`. This is a long sequence of `regexp` steps mapping the county's 2-letter `JURISDICTI` township/city codes (e.g. `OA` -> Oakfield, `FO`/`FD` -> Fond du Lac, `RP`/`RI` -> Ripon) to full names. With the bug, only step 1 (`OA` -> `Oakfield`) could ever fire; every other code passed through unmapped.

## Fix

Changed every step after the first in both chains to read `"oa:street"` / `"oa:c"` respectively, so each step now operates on the previous step's output as intended, turning both into coherent multi-step pipelines.

I verified this is safe by downloading and inspecting the county's live `Addresses.gdb.zip` with `ogrinfo`/`ogr2ogr`:

- `street`: real `PRI_NAME`/`DIRECTION`/`TYPE` samples confirm normal streets (e.g. `DIRECTION=N, PRI_NAME=BARTON, TYPE=RD`) pass through steps 1-3 unchanged and are correctly joined into `"N BARTON RD"` in step 4, while county highway records (e.g. `PRI_NAME=CTH W`, `DIRECTION`/`TYPE` blank) get correctly rewritten to `"County Highway W"` by step 3 before the final join, matching the source's own pre-built `SITE` field values. There are no `STH`/`USH`-prefixed names in the live data currently, but those regexps are harmless pass-throughs when they don't match, so leaving them chained is correct and future-proof.
- `city`: none of the 27 code patterns overlap with each other's *output* strings (e.g. `"Oakfield"` can never accidentally match a later `^XX$` 2-letter pattern), so chaining them all sequentially through `oa:c` is safe regardless of step order.

## Note for reviewers (not fixed, out of scope for this PR)

Two separate, pre-existing issues I found but did not touch, since they're outside the scope of the chain-wiring bug:

1. **The `JURISDICTI` field is currently empty for all 48,007 records** in the live `Addresses.gdb.zip` download (confirmed via `ogrinfo`/`ogr2ogr` — `MUNICIPALI` and `City` are also entirely blank). So even with the chain fixed, the `city` conform currently produces no value on any row until the county repopulates that field. This is a data-source issue, not a conform bug, and didn't originate with this fix.
2. **Duplicate pattern**: two different steps both match `^CA$` — one mapping to `"Calumet"` (a town) and a later one to `"Campbellsport"` (a village). Since chaining is sequential, only the first (`Calumet`) could ever actually fire for a raw `CA` code; the second is unreachable. I could not determine from the county's public data (the field is blank, and there's no embedded domain/lookup table in the GDB) which 2-letter code is actually correct for Campbellsport, so I left both patterns as originally written rather than guess. A maintainer with knowledge of the county's actual jurisdiction code scheme should resolve this — Campbellsport's code is very likely something other than `CA` (e.g. `CP`).

### DIFF
--- a/sources/us/wi/fond_du_lac.json
+++ b/sources/us/wi/fond_du_lac.json
@@ -36,19 +36,19 @@
                             },
                             {
                                 "function": "regexp",
-                                "field": "street",
+                                "field": "oa:street",
                                 "pattern": "^USH (.+)",
                                 "replace": "US Highway $1"
                             },
                             {
                                 "function": "regexp",
-                                "field": "street",
+                                "field": "oa:street",
                                 "pattern": "^CTH (.+)",
                                 "replace": "County Highway $1"
                             },
                             {
                                 "function": "join",
-                                "fields": ["DIRECTION", "street", "TYPE"],
+                                "fields": ["DIRECTION", "oa:street", "TYPE"],
                                 "separator": " "
                             }
                         ]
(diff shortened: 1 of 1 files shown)


iandees:
## The previous fix regressed `street` for essentially all records

I pulled the real production output from job 912348 (`https://v2.openaddresses.io/batch-prod/job/912348/validated.geojson.gz`, 43,246 records) and found `street` was wrong for effectively 100% of rows — every value was a bare direction+type abbreviation (`'ST'`, `'E ST'`, `'S AV'`, etc., only 39 distinct values total in the whole file). The actual street name (e.g. `BARTON`, `MAIN`) was missing from every single record.

### Root cause

I traced the real `row_fxn_chain`/`row_fxn_regexp`/`row_fxn_join` implementation in `batch-machine`'s `openaddr/conform.py`. Because `variable` is set to `"street"` here (same name as the output field), the chain accumulates directly in `oa:street`/`street` for every step. The previous fix had steps 2–4 read `"field": "oa:street"` — the *always-current* accumulator, which a regexp step resets to `''` the instant its own pattern doesn't match. So:

1. Step 1 (`PRI_NAME` → `^STH (.+)`) doesn't match for a normal street like `BARTON` → `oa:street` becomes `''`.
2. Step 2 reads `oa:street` (now `''`), tries `^USH (.+)` → still `''`.
3. Step 3 reads `oa:street` (`''`), tries `^CTH (.+)` → still `''`.
4. Step 4 joins `DIRECTION` + `oa:street` (`''`) + `TYPE` → the real street name is gone, leaving only the direction/type, e.g. `'E ST'`.

Only the ~6.8% of records whose `PRI_NAME` actually starts with `CTH ` (no `STH `/`USH ` values exist in the current source data) would have partially worked — everything else lost the street name entirely. This matches the observed production output exactly.

### The fix

I restructured the chain so a non-matching highway-prefix step can't wipe out the accumulated name, using the "sticky bare field" mechanic in `row_fxn_chain` (`row[key]` is only overwritten when the previous step's `oa:`-prefixed result is non-empty, so it retains the last good value):

```json
"street": {
    "function": "chain",
    "variable": "street",
    "functions": [
        { "function": "regexp", "field": "PRI_NAME", "pattern": "^(.*)$", "replace": "$1" },
        { "function": "regexp", "field": "street", "pattern": "^STH (.+)$", "replace": "State Highway $1" },
        { "function": "regexp", "field": "street", "pattern": "^USH (.+)$", "replace": "US Highway $1" },
        { "function": "regexp", "field": "street", "pattern": "^CTH (.+)$", "replace": "County Highway $1" },
        { "function": "join", "fields": ["DIRECTION", "street", "TYPE"], "separator": " " }
    ]
}
```

Step 1 establishes `PRI_NAME` as the sticky baseline (an always-matching identity regexp, so `street` bare always holds the raw name). Steps 2–4 now read the bare `street` (not `oa:street`) as input, so a non-matching step's fresh-empty `oa:street` doesn't get read as input by the next step — the bare value stays intact until a highway prefix actually matches.

I verified this against `batch-machine`'s actual `openaddr/conform.py` source (not just `ATTRIBUTE_FUNCTIONS.md` prose) and its own test suite (`test_row_fxn_chain_nested` uses this exact bare-field-carries-forward pattern), then downloaded the live source GDB (`https://gisweb.fdlco.wi.gov/download/Addresses.gdb.zip`, 48,007 records) and simulated the new chain in Python against every real row:

- 0 empty outputs across all 48,007 rows
- 1,913 distinct `street` values (reasonable diversity for a county this size)
- Normal streets now produce full names, e.g. `PRI_NAME=BARTON, DIRECTION=N, TYPE=RD` → `"N BARTON RD"` (previously `"RD"`)
- `PRI_NAME=CTH BB` (a county highway, direction/type blank) → `"County Highway BB"`, still correctly expanded
- Spot-checked 10 additional random real rows (e.g. `E BANK ST`, `N MAIN ST`, `N PIONEER RD`, `CENTURY LA`, `W2966 ELMORE DR`) — all produced full, correct street names

No `STH `/`USH `-prefixed values currently exist in the live data, so I couldn't hand-verify those two branches against real rows, but they follow the identical structure as the now-confirmed-working `CTH` branch.

### `city` chain — left unchanged, here's why

The `city` chain (`variable: "c"`, keyed off `JURISDICTI`) has the exact same structural issue — every step after the first reads `oa:c` instead of a sticky value, so only the very first condition (`^OA$` → Oakfield) could ever actually take effect; every other town code would be swallowed by the same wipe-out bug.

However, I confirmed against the live GDB that `JURISDICTI`, `MUNICIPALI`, and `City` are **empty on all 48,007 records** — so the `city` output is `''` in production today regardless of this bug, whether fixed or not. Since there's no live impact to verify against and fixing it blind carries risk, I left it as-is rather than blindly mirroring the `street` fix.

One more thing worth flagging for whoever revisits this: the chain has a duplicate `^CA$` rule (line mapping to `Calumet`, and a second, later one mapping to `Campbellsport`) — a pre-existing data ambiguity unrelated to the chain-structure bug. If this chain is ever reworked (a single `map` function would be a much more robust replacement for this style of mutually-exclusive lookup than 28 chained conditional regexps), that duplicate needs to be resolved first since a JSON `mapping` object can't hold both keys.

### Validation

- `npm test` passes with the same 2 pre-existing unrelated failures present on `master` (`non-integer accuracy value should fail`, `null` — both in `schema_validation_v2.test.js`, unaffected by this change)
- Schema validation of the updated file passes (`VALID`)

PR #8552, opened 2026-09-16
Fix street conform case mismatch for 7 Kansas counties

ilissablech:
anderson_county, atchison_county, bourbon_county, crawford_county, doniphan_county, linn_county, and nemaha_county all map conform.street to "street", but each source's cache CSV (data.openaddresses.io/cache/us-ks-*.csv) has the column named STREET (uppercase). The case mismatch means every row's street field ends up empty, which fails OA's own build-validity check (job_details.stats.validity.valid = 0, requiring both number and street to be populated) even though the underlying data is fine.

Confirmed against anderson_county's exported data: 3,912/3,912 rows had a number but zero had a street, and the upstream CSV header is literally "X,Y,NUM,STREET". All 6 other counties share the same source template and the same bug.


ilissablech:
@iandees are you able to review 🙇

iandees:
Looks good, thank you!

